PR #85, opened 2015-04-24
Implement new HTML styles

hawkowl:
This implements new HTML styles for pydoctor, using bootstrap.

As a result it makes a bunch of changes to the code.

  - removed the `html-use-sorttable`, `html-use-splitlinks`, and `html-shorten-lists` command-line options,
  - moved from `t:invisible` to `t:transparent` nodes in the template
  - renamed `maybeShortenList` to `assembleList`
  - removed abbreviation in `commonPage.mediumName`
  - got rid of `has_lineno_col` attribute of `ChildTable` and `TableRow`

The net effect is that this greatly simplifies the number of types of output that can be produced, but produces a much better and more aesthetically pleasant default.

You can preview the changes with a run on the Twisted codebase here: https://dl.dropboxusercontent.com/u/14290114/pydoctortest/twisted/index.html


pydoctor/templatewriter/writer.py
```diff
@@ -74,7 +76,7 @@ def writeDocsFor(self, ob, functionpages):
     def writeDocsForOne(self, ob, fobj):
         if not ob.isVisible:
             return
-        # brrrrrrrr!
+        # brrrrrrr!
```
mwhudson: lolz

pydoctor/templates/common.html
```diff
@@ -5,51 +5,75 @@
...
-        <p>
-          Inherited from <t:slot name="baseName" />:
-        </p>
-        <t:slot name="baseTable" />
-      </t:invisible>
-    </div>
-    <div id="bigTable" t:render="bigTable">
-    </div>
-    <t:slot name="packageInitTable" />
+    <nav class="navbar navbar-default">
```
mwhudson: <nav> is IE9+ I think, is that OK?

pydoctor/templates/common.html
```diff
@@ -5,51 +5,75 @@
   <head>
     <title><t:slot name="title">Something</t:slot> : API documentation</title>
     <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
-    <link rel="stylesheet" type="text/css" href="apidocs.css" />
-    <script t:render="ifusesorttable" type="text/javascript"
-            src="sorttable.js"></script>
-    <script t:render="pydoctorjs" type="text/javascript"
-            src="pydoctor.js"></script>
```
mwhudson: I guess you can remove the .js files (and supporting code) unless there are other references to them.

pydoctor/templates/common.html
mwhudson: Also it's HTML5 and not xhtml1-strict so I guess we should update the dtd?

mwhudson: And finally, all these navbar blocks are the same aren't they? Is there really no way to combine them?

mwhudson:
Well so I barely feel competent to review this any more :) I made a few comments but I'm tempted to just say yes once you answer them, unless you can persuade anyone else to review it...

pydoctor/templates/common.html
hawkowl: I think that we can safely stop supporting IE9. And I'll update the dtd, and look at combining the navbar blocks. I'm not completely sure how to do this nicely in t.w.t.

pydoctor/templates/common.html
hawkowl: I've got a few in-progress changes that rip all the supporting code out. I'll get everything working and push them up too.

pydoctor/templates/common.html
hawkowl: Updated (all?) the dtds.

pydoctor/templates/common.html
hawkowl: Ripped out everything to do with them. :)

pydoctor/templates/apidocs.css
```diff
@@ -1,287 +1,210 @@
...
+.navbar {
+    margin-bottom: 0px;
+}
+
+.page-header {
+    margin-top: 22px;
+}
+
+
+ul {
```
mwhudson: isn't this going to affect <ul>s in docstrings too?  feels like it should be a _bit_ more qualified.

mwhudson:
I think this is basically good to land.  It's a bit of a jumble of things, but I don't think it's really worth the effort of separating them out.  Can you try to be a bit more focused in future changes though?

Squash merge it and write a proper commit message pls.

glyph:
I wrote the proper commit message.  Why would you want a squash merge though, vandalizing your history is bad :(

mwhudson:
If I was being a purist I'd have made @hawkowl chop this whole PR up into about 4 different ones!

Issue #87, opened 2015-05-13
new stylesheet doesn't seem to hide private stuff


glyph:
https://twisted.github.io/tubes/docs/classIndex.html shows all private names looking equivalent to public ones.

mwhudson:
Did this change with @hawkowl's recent stuff?

glyph:
I think so yes.

hawkowl:
Yeah, this is actually a different thing that what I implemented, mine was on the class itself, not the class index.

mthuurne:
This got fixed in #311.

PR #88, opened 2015-05-14
Hide private stuff for real

hawkowl:
For #87 -- this uses a little bit of JS to hide private things using CSS. The no-JS fallback is that private things are still shown.

I had some reservations about performance, but this approach only modifies the DOM once, and the rest is done in CSS, making it very fast.

Test pages are available at https://dl.dropboxusercontent.com/u/14290114/pydoctortest/twisted-hidden/index.html .

### DIFF
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -182,13 +182,13 @@ code > .code {
 }
 
 
-div.function {
+#childList > div {
     border-left-color: #03a9f4;
     border-left-width: 1px;
     border-left-style: solid;
 }
 
-div.function .functionHeader {
+#childList > div .functionHeader {
     font-family: monospace;
 }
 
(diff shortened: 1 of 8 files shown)


pydoctor/templates/apidocs.css
```diff
@@ -182,13 +182,13 @@ code > .code {
 }
 
 
-div.function {
+#childList > div {
```
glyph: I'm not a super fan of "div" in selectors. It's not exactly self-documenting. `#childList` at least I sort of know what that means.  Not a blocker though, this is all CSS, YOLO

hawkowl:
I'm assuming this #YOLO from glyph is a review :p 

It's in production, should be fine

glyph:
yes sorry I should have applied the "yolo" label

PR #90, opened 2015-05-22
add a data-type for doc2dash

hawkowl:
This adds a data-* attribute in the nameIndex.html for doc2dash to extract.


mwhudson:
Maybe add a comment?  But looks fine.

hawkowl:
Added a comment. :)

Issue #94, opened 2015-07-01
installs on python3


rbtcollins:
pip install pydoctor on python3 doesn't error. But its not useful today.

I suspect this might be because there are no trove classifiers on pydoctor -at all- (but haven't cross-checked that). Anyhow - a simple version check in setup.py + raising on version 3, or try to import compiler and raise if its missing, would be a good way to signal failure early.

wsanchez:
This seem like a dup of #96, especially given that #97 is done, no?
I guess this is older, but #96 has more info.

mthuurne:
The current master branch runs fine on Python 3. I think this can be closed.

wsanchez:
```console
wsanchez$ pipx install --python=python3.8 --spec=git+https://github.com/twisted/pydoctor.git@master pydoctor 
  installed package pydoctor 19.11.0, Python 3.8.0
  These apps are now globally available
    - pydoctor
done! ✨ 🌟 ✨
```

PR #97, opened 2015-08-09
Disable installation on Python 3

hawkowl:
Would fix the "installs but doesn't work" problem in #94.


mwhudson:
Seem fine. would be nice if there was a way to get pypi to tell pip to not even try on py3 but if there isn't this'll do.

PR #108, opened 2016-07-07
Remove server code

hawkowl:
It's unused, the tests are currently broken, so I think we should remove it for now.


mwhudson:
+1

It's a shame but it's not like I'm going to spend time working on this so focus++

PR #109, opened 2016-07-07
Release 16.0

hawkowl:
cc @mwhudson

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ from setuptools import setup
 
 setup(
     name='pydoctor',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -10,7 +10,7 @@
 
 setup(
     name='pydoctor',
-    version='15.1.0',
+    version='16.0.0',
     author='Michael Hudson-Doyle',
     author_email='micahel@gmail.com',
```
mwhudson: Eh I guess we should update these bits :)

PR #110, opened 2016-07-07
Make it monkeypatchable

hawkowl:
This allows some horrible hacks on Twisted's side to make it load local templates.

cc @mwhudson


mwhudson:
errrr ok

Issue #92, opened 2015-05-29
Can't use a custom --system-class (AttributeError)


ikalev:
Using a custom subclass of ``pydoctor.twistedmodel.TwistedSystem``:

```
Traceback (most recent call last):
  File "C:\Anaconda\lib\site-packages\pydoctor\driver.py", line 390, in main
    writer.writeIndividualFiles(subjects, options.htmlfunctionpages)
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\writer.py", line 43, in writeIndividualFiles
    self.writeDocsFor(ob, functionpages=functionpages)
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\writer.py", line 68, in writeDocsFor
    self.writeDocsFor(o, functionpages)
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\writer.py", line 68, in writeDocsFor
    self.writeDocsFor(o, functionpages)
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\writer.py", line 68, in writeDocsFor
    self.writeDocsFor(o, functionpages)
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\writer.py", line 65, in writeDocsFor
    self.writeDocsForOne(ob, f)
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\writer.py", line 86, in writeDocsForOne
    flattenToFile(fobj, page)
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\writer.py", line 19, in flattenToFile
    raise err[0]
twisted.web.error.FlattenerError: Exception while flattening:

  File "C:\Anaconda\lib\site-packages\twisted\web\_flatten.py", line 317, in _flattenTree
    element = stack[-1].next()
  File "C:\Anaconda\lib\site-packages\twisted\web\_flatten.py", line 240, in _flattenElement
    result = renderMethod(request, rootClone)
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\pages\__init__.py", line 159, in all
    heading=self.heading(),
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\pages\__init__.py", line 71, in heading
    tags.code(self.mediumName(self.ob)), " ",
  File "C:\Anaconda\lib\site-packages\pydoctor\templatewriter\pages\__init__.py", line 277, in mediumName
    r.append(taglink(o, n))
[...]

PR #111, opened 2016-07-09
fix: System has no method "warning"

ikalev:
See #92 for details.

### DIFF
--- a/pydoctor/templatewriter/util.py
+++ b/pydoctor/templatewriter/util.py
@@ -8,7 +8,7 @@ import os, urllib
 
 def link(o):
     if not o.isVisible:
-        o.system.warning("html", "don't link to %s"%o.fullName())
+        o.system.msg("html", "don't link to %s"%o.fullName())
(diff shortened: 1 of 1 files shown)


adiroiban:
Thanks for the patch

Changes look good, but the warning method is also used in pydoctor/astbuilder.py and implemented as

```
    def warning(self, type, detail):
        self.system._warning(self.current, type, detail)
```

so maybe we should also use self.system._warning here.


Now, I have no idea why this method is private.


I would argue that instead of using msg, we should make pydoctor.model.System._warning public by renaming it to pydoctor.model.System.warning

and also add a docstring :)

What do you think?

ikalev:
This was my intuition too, but ``_warning()`` has a different signature. Looking at the calling code, the author clearly intended to call a method with a signature matching ``msg()``, not ``_warning()``.

adiroiban:
True. Please update merge master in this branch and I will merge.

Sorry for the trouble... protected branched in GitHub are a PITA for non-team members.

ikalev:
Done

PR #118, opened 2016-07-19
[Fix #117] Report intersphinx error as level -1

### LINKED ISSUE #117: Add support for reporting only errors
all information and error messages are logged with the same level.

adiroiban:
See #117 

intershinx errors are reported with level 0 just like normal info messages.

It was my fault for not understand how pydoctor.model.System.msg function works and not paying attention that I use the same method for both info and error messages.

Now errors are reported with -1

I have updated the release number, so that after this is merge we can do a release as it will help Twisted :)

I have updated the travis-ci job to also run against the latest twisted trunk... in this way I hope that we can prevent some regressions

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,12 @@ matrix:
       env: TOX_ENV=test-pypy
     - python: 2.7
       env: TOX_ENV=test-py27-codecov-travis
+    - python: 2.7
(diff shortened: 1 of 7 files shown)


adiroiban:
ready for review. thanks!

Issue #125, opened 2016-08-03
Document Documentable.expandName


adiroiban:
The expandName() docstring say it returns a fully qualified name, but I can also see that at least for interface references, it is not returning FQN.

For example in the context of twisted.web.client.Response._construct  `expandName('IResponse.request') -> 'request'`

We need a set of automated tests to help understand what this code should do and to review the docstring.

tristanlatr:
I feel like this issue has been fixed. 

At least for the docstring part, it's quite clear now. 
I've added a few details in #414. 

<img width="750" alt="Screen Shot 2021-06-26 at 8 31 24 PM" src="https://user-images.githubusercontent.com/19967168/123529371-7d0b7d00-d6bd-11eb-8d39-daf4e087ea2e.png">

Maybe there is still an issue with interfaces, I've not added special tests for that.

Issue #124, opened 2016-08-01
Fix intersphinx name resolution


adiroiban:
Somehow the name resolution find the wrong name 

Here is an error
```
twisted.web._http2.H2Connection:65 invalid ref to 'priority.PriorityTree' resolved as 'PriorityTree'
twisted.web._http2.H2Connection.priority:0 invalid ref to 'priority.PriorityTree' resolved as 'PriorityTree'
```

adiroiban:
Most probably this regression was introduced here https://github.com/twisted/pydoctor/pull/113  it looks like in some cased `fullerID` is failing to get the right id.

adiroiban:
So,

we have this 

```
class H2Connection(Protocol, TimeoutMixin):
    """
    @ivar priority: A HTTP/2 priority tree used to ensure that responses are
        prioritised appropriately.
    @type priority: L{priority.PriorityTree}
    """
```

pydoctor.model.Documentable.expandName will think that the referred ` L{priority.PriorityTree}` is `H2Connection.priority.PriorityTree` since the parent of `PriorityTree` is `priority` which is the same name as the `ivar`

mwhudson:
Ugh, this stuff is all so fiddly. Can we override expandName on Attribute to just forward directly to self.parent?

adiroiban:
Well, we can shortcut the attribute, but the issue is still there as the class is also failing to resolve the name.

```
twisted.web._http2.H2Connection:65 invalid ref to 'priority.PriorityTree' resolved as 'PriorityTree'
```

------------

The ugly part for me is that this code has no unit tests so I am using the twisted codebase to check changes...and this is slow 

I have created a separate ticket for expandName https://github.com/twisted/pydoctor/issues/125

I have submitted an PR with an ugly hack to fix this https://github.com/twisted/pydoctor/pull/126 as I have no other idea and I don't have more time to spend debugging this code :(

tristanlatr:
We should check if this kind of errors also happens with latest pydoctor.

PR #126, opened 2016-08-03
Fix name resolution for intersphinx links

adiroiban:
See #124 

The root problem is Documentable.expandName but I was not able to fix its implementation. 
See #125 for more details.

I went with a dirty hack.

I have also increased the release number, so once this is merged we can create a tag and release a new version.

### DIFF
--- a/pydoctor/__init__.py
+++ b/pydoctor/__init__.py
@@ -1,3 +1,3 @@
 """PyDoctor, an API documentation generator for Python libraries."""
 
-version_info = (16, 2, 0, '', 0)
+version_info = (16, 2, 1, '', 0)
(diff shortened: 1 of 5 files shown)


glyph:
@adiroiban - I want to say "add a test case for this" but somehow this just _increases_ coverage already? :)

adiroiban:
I start writing unit tests for Documentable.expandName ... but I found out that I had no idea what is going on there and how to write a proper fix and proper tests for it.

with a previous branch I have added functional end-to-end tests by running pydoctor against Twisted trunk...so we do have tests for this change... just that they are not unit tests and not easy to follow.

so, I can write a unit tests for translate_identifier_xref ... but the problem is that I don't really know what is wrong here and how to write the docstring for the test

glyph:
@mwhudson - any help?

hawkowl:
Since this has 100% coverage, and it fixes an issue, and we don't really know any better way than yoloing it... sure, let's go with this.

PR #127, opened 2016-08-03
Improve UX by making the button text indicate the target state.

dreid:


adiroiban:
Looks good!

Please also update the version number, so that after this is merged, we can create a tag and have the new version auto-published.

Thanks!

glyph:
Hooray!

glyph:
Since there are two positive comments already and the feedback was dealt with, landing.

adiroiban:
don't forget that after landing, to create a tag so that it will be auto-released ... happy to get feedback about how we can make the release process easier :)

glyph:
@adiroiban - pydoctor has fully automated releases?

adiroiban:
@glyph I don't know what do you expect from a fully automated release... but if you push a new tag, travis-ci will run the tests and publish if on pypy . This was done in #115  and is briefly documented here https://github.com/twisted/pydoctor#releasing-a-new-package :)

glyph:
@adiroiban That is more or less _exactly_ what I would expect from a fully automated release :).

glyph:
Thanks very much for doing that!  I can't wait until we have a similar thing for Twisted...

adiroiban:
I have pushed a tag for 16.2.1 ... and is now in pypi

Travis-CI is a bit crazy as there is no dedicated job for the release/deployment so without webhooks it is just a big pile of hacks :(

PR #130, opened 2017-01-18
Switch to exception syntax specified in PEP 3110.

rodrigc:
See:
https://www.python.org/dev/peps/pep-3110/

This is for compatibility with Python 3


mwhudson:
Seems OK to me, what does @hawkowl think?

rodrigc:
@hawkowl @mwhudson Is this OK?

PR #131, opened 2017-01-18
Use print function as specified in PEP 3105

rodrigc:
See:
https://www.python.org/dev/peps/pep-3105

This is for compatibility with Python 3


mwhudson:
test failures look real (maybe silly tests though)

rodrigc:
@mwhudson I fixed the issue, now the tests are working.

PR #132, opened 2017-01-18
Use exec as a function, for compatibility with Python 3

rodrigc:
See:
https://docs.python.org/3/whatsnew/3.0.html#removed-syntax


mwhudson:
lgtm

glyph:
same.

PR #134, opened 2017-01-20
import builtins on Python 3, import __builtins__ on Python 2

rodrigc:


pydoctor/epydoc2stan.py
```diff
@@ -2,7 +2,14 @@
 Convert epydoc markup into renderable content.
 """
 
-import __builtin__
+try:
+    # Python 3
+    import builtins
+except ImportError:
+    # Python 2
+    import __builtin__
```
glyph: It might be better to add `six` to `install_requires` and then do `from six.moves import builtins`.

If not, `import __builtin__ as builtin` will avoid binding two names and having some broken code rely on `__builtin__` by accident.

glyph:
I'd land it, except: conflicts :(

rodrigc:
@glyph I have made the suggested change.  I cannot merge this because I don't have write permissions to this repository.  Please merge this for me.

PR #140, opened 2017-09-02
Switch to pytest.

### LINKED ISSUE #139: Switch to pytest
[nosetests appears to be abandoned](https://nose.readthedocs.io/en/latest/#note-to-users).  The easiest path forward would be to replace it with [pytest](https://docs.pytest.org/en/latest/)

markrwilliams:
Fixes #139

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,8 +7,8 @@ requires = epydoc
 doc_files = README.txt
 
 
-[nosetests]
-match=^test_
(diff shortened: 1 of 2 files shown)


mahmoud:
Per CR request, LGTM. I like the usage of `-m pytest` instead of relying on the `{envdir}` and the executable. Also, good riddance to nose!

PR #142, opened 2017-09-03
twisted-apidoc tox env builds against post src-Twisted.

### LINKED ISSUE #141: twisted-apidoc tox environment broken
The `twisted-apidocs` tox environment [fails](https://travis-ci.org/twisted/pydoctor/jobs/271280937#L480).  
It's an allowed failure, so I guess nobody noticed.

We should fix it or remove it.

markrwilliams:
The move to src/ broke this.

Closes #141.

Note that this environment will still fail because Twisted has errors in its documentation.


markrwilliams:
@moshez Putting absolute paths in quiets tox.  I prefer doing this to adding commands to `whitelist_commands`.

Issue #129, opened 2016-10-29
allow for local intersphinx caching


glyph:
Right now the only way I can see that pydoctor allows for intersphinx inventories to be used is to fetch them every time.

There should be a way to cache these so they can be checked into a project repo and not fetched every time, to allow for running tests like Twisted's, which check that documentation generation doesn't fail, offline.

mwhudson:
Yeah, I think @adiroiban and I talked about this when I reviewed the initial intersphinx work and it should just be done.

adiroiban:
There is also this ticket https://github.com/twisted/pydoctor/issues/120

twm:
#143 seems to have resolved this.

PR #143, opened 2017-09-03
Cache Interpshinx objects.

### LINKED ISSUE #120: Add cache for intersphinx inventories
Sometimes the intersphinx inventories are down 

In this case, it would help to use a cached version, rather than to generate a documentation with invalid links...

The cache can have an expire age (ex 1 week), to prevent the case when the documentation was permanently moved or is permanently down

The cache can be placed in `~/.cache/pydoctor` so that we can persist the cache in Travis-CI 

We can start with Linux/Unix only cache support ... but maybe is easy to put it in APPDATA on Windows

There can also be a flat to disable cache `--cache-age=0` so that at a release we are forced to check the state of external api docs.

---

We use pydoctor as part of the Twisted's test suite and I hope that with the cache we can prevent false failures due to temporary unavailability of a API doc site.

markrwilliams:
Closes #120, #129, and #137 

This introduces four new command line options:

1. **--enable-intersphinx-cache** Enable the Intersphinx object cache, which is disabled by default.
2. **--intersphinx-cache-path** The location of the Intersphinx object cache.  By default, this is a user cache dir selected by appdirs.
3. **--clear-intersphinx-cache** Remove the Intersphinx cache directory.  This does not require the cache be enabled.
4. **--intersphinx-cache-max-age** The maximum age for *new* items in the Intersphinx cache directory. 
 This defaults to 1 week.  It's specified by a number followed by a single letter representing the units, e.g., `20h` is 20 hours.

The implementation CacheControl with an ExpiresAfter heuristic.  The heuristic was chosen because most of Twisted's objects.inv URLs do not set the necessary cache headers.

CacheControl requires requests, so SphinxInventory now uses requests.  This closes #137.

### DIFF
--- a/pydoctor/driver.py
+++ b/pydoctor/driver.py
@@ -3,7 +3,13 @@
 from __future__ import print_function
 
 from pydoctor import model, zopeinterface
-from pydoctor.sphinx import SphinxInventory
+from pydoctor.sphinx import (
+    SphinxInventory,
+    prepareCache,
+    MAX_AGE_HELP,
+    USER_INTERSPHINX_CACHE
+)
+
 import sys, os, datetime
 
 BUILDTIME_FORMAT = '%Y-%m-%d %H:%M:%S'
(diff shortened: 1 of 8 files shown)


hawkowl:
what could possibly go wrong

PR #145, opened 2017-09-04
Release 17.9 & Use Incremental

hawkowl:


pydoctor/__init__.py
```diff
@@ -1,5 +1,5 @@
 """PyDoctor, an API documentation generator for Python libraries."""
 
-from __future__ import print_function
+from ._version import __version__
 
-version_info = (16, 2, 1, '', 0)
```
markrwilliams: This is a backwards-incompatible change.  The README, however, says that pydoctor exists for Twisted, so as long as we don't need this it's OK.

hawkowl: pydoctor doesn't really have a stable public API, either :P

hawkowl: ah twisted relies on this but thats from the days we didn't use setuptools properly

PR #146, opened 2017-12-12
Make the output reproducible

lamby:
Whilst working on the Reproducible Builds effort [0], we noticed
that pydoctor generates output that is not reproducible.

By iterating over the filesystem (which returns values in a non-
determinstic order) we end up with different id="" attributes in
the HTML.

This affects other packages such as "subvertpy" amongst others.

 [0] https://reproducible-builds.org/

### DIFF
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -623,7 +623,7 @@ class System(object):
         package = self.ensurePackage(package_full_name)
         package.filepath = dirpath
         self.setSourceHref(package)
-        for fname in os.listdir(dirpath):
+        for fname in sorted(os.listdir(dirpath)):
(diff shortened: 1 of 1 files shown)


glyph:
Ideally this would come with a change, but it looks like a very straightforward change, and pydoctor doesn't have the greatest coverage to begin with :).

Thanks for this contribution!

lamby:
@glyph 

> Ideally this would come with a change,

Hm? :)

ldanielburr:
@lamby I think @glyph meant to say "Ideally this would come with a **test,..."**

lamby:
Ah! Thanks :)

glyph:
Indeed that is what I meant.

PR #150, opened 2019-03-31
Fix XMLFile sans FilePath deprecation

twm:
Fix this warning:

    pydoctor/templatewriter/pages/table.py:36
      /home/twm/pydoctor/pydoctor/templatewriter/pages/table.py:36: DeprecationWarning: Passing filenames or file objects to XMLFile is deprecated since Twisted 12.1.  Pass a FilePath instead.
        loader = XMLFile(util.templatefile('table.html'))


twm:
(The test suite will fail due to #149.)

twm:
:tada:

PR #151, opened 2019-03-31
Switch from the compiler module to the ast module

hawkowl:


pydoctor/astbuilder.py
```diff
@@ -17,10 +20,10 @@ class str_with_orig(str):
     """
     pass
 
-class MyTransformer(transformer.Transformer):
```
mwhudson: FWIW, since the server code got deleted ages ago, MyTransformer and str_with_orig can go too.

pydoctor/astbuilder.py
```diff
@@ -64,45 +67,51 @@ def parseFile(path):
 
 def parse(buf):
     """Duplicate of L{compiler.parse} that uses L{MyTransformer}."""
-    return MyTransformer().parsesuite(buf)
+    p = ast.parse(buf)
+    return MyTransformer().visit(p)
 
 
 def node2dottedname(node):
+    print(node)
```
mwhudson: Probably don't want this?

pydoctor/zopeinterface.py
```diff
@@ -75,9 +77,10 @@ def docsources(self):
                         yield io2.contents[self.name]
 
 def addInterfaceInfoToModule(module, interfaceargs):
+    print("!!!!!!!!!!!!!!")
```
mwhudson: ??????

pydoctor/zopeinterface.py
```diff
@@ -267,6 +270,7 @@ def visitClass(self, node):
         for n, o in zip(cls.bases, cls.baseobjects):
             if schema_prog.match(n) or (o and o.isschemafield):
                 cls.isschemafield = True
+        print(cls.decorators)
```
mwhudson: more debugging?

mwhudson:
Oh yeah, and you should add 'astor' to setup.py I guess.

GaretJax:
@hawkowl let me know if I can help in any way with this.

glyph:
So this, plus the epydoc stuff, means we have a hope of actually moving to py3, right?

GaretJax:
I'm using it on py3.6 here with this branch https://github.com/twisted/pydoctor/pull/152

glyph:
@GaretJax Did the epydoc import just... work by accident, without importing it?

GaretJax:
I had to run 2to3 on it plus apply a manual fix (see the comment at https://github.com/twisted/pydoctor/pull/152#issue-287959487).

glyph:
It looks like this just needs some conflicts resolved, then it should be landed?

hawkowl:
@glyph it fails on Twisted as the unit tests are incomplete, so, not quite

glyph:
@hawkowl should we just add twisted doc generation to CI here?

hawkowl:
It is, it's just allowed failure.

glyph:
Maybe it should be required? I can't imagine we'd ever want to do a pydoctor release *not* gated on twisted doc gen

mthuurne:
Please have a look at #160. It would be nice if the ast module port could land in master soonish, since other big changes such as Python 3 support and type annotations support would have to be built on top of it.

glyph:
@mthuurne @hawkowl @mwhudson 

I'm inclined to say: let's merge #160 into this, let's merge this, and then keep going from there.

Any objections?

mthuurne:
@glyph I fully agree with that approach.

I'm already working on rebasing #152; merging that could be the next step after merging this PR.

mwhudson:
I'm definitely in favour of rolling forward and fixing breakage rather than hanging around for too long.

Issue #96, opened 2015-08-09
Support Python 3


hawkowl:
This is an organizing for Python 3 support, to keep everything under one roof.

TODO: add checklist?

mwhudson:
The major item of work is moving away from the compiler package. astbuilder is the core hard part of pydoctor!

twm:
https://github.com/twisted/pydoctor/issues/147 (epydoc only works on Python 2) seems relevant.

mthuurne:
As far as I know, the master branch of pydoctor runs fine on Python 3 now. Please try it on your code and report any issues.

The next step would be dropping Python 2.7 support. Should that be tracked in this issue or should I create a new issue for that?

mikepurvis:
```
$ pip3 install git+https://github.com/twisted/pydoctor.git --upgrade
...
Successfully built pydoctor CacheControl
ERROR: pydoctor 19.11.0 requires enum34, which is not installed.
Installing collected packages: incremental, appdirs, msgpack, lockfile, CacheControl, pydoctor
```

The `enum34` dep should be conditional on `sys.version_info`. Otherwise it seems to work great for my use-case. Thanks for your work on this and for pinging to let me know.

mthuurne:
> ```
> $ pip3 install git+https://github.com/twisted/pydoctor.git --upgrade
> ...
> Successfully built pydoctor CacheControl
> ERROR: pydoctor 19.11.0 requires enum34, which is not installed.
> Installing collected packages: incremental, appdirs, msgpack, lockfile, CacheControl, pydoctor
> ```
> 
> The `enum34` dep should be conditional on `sys.version_info`. Otherwise it seems to work great for my use-case. Thanks for your work on this and for pinging to let me know.

The entry in `setup.py` reads `"enum34;python_version<'3.4'"`, which is supposed to restrict the requirement to Python versions that actually need it.

I can't reproduce the problem: if I run the same command line in a fresh virtualenv, it ignores `enum34`, as it should. I'm using Python 3.6.9 and pip 19.3.1, on Linux.

mikepurvis:
Ah, okay— this is probably just my computer's python environment being a superfund site. Thanks for looking!

wsanchez:
I created a [PR in Klein](https://github.com/twisted/klein/pull/315) to generate API docs.  We can test Python 3 support there.

sandrotosi:
[...]

PR #152, opened 2019-06-13
Python 3.6 compatibility

GaretJax:
This is an initial effort to tackle #96. This is not ready to be merged but should serve as a ground for discussion.

The changes in this PR currently reflect the bare minimum to make the test suite pass on both python 2.7 and python 3.6. This is probably not sufficient as the test coverage is currently ~64%.

Requires `epytext` to be run through the 2to3 tool (`2to3 --no-diffs -w $(python3 -c 'import epydoc, os; print(epydoc.__path__[0])')` and to manually replace one occurrence of `string.expandtabs` with `str.expandtabs` (in `epydoc/markup/epytext.py:256`).


GaretJax:
Well... I should probably have checked the open PRs before working on this... https://github.com/twisted/pydoctor/pull/151 seems to do a big chunk and then some more of the work done here.

hawkowl:
Closing this since mthuurne combined both our approaches :)

Issue #147, opened 2018-09-05
Drop epydoc support


jelmer:
Please consider moving away from the use of Epydoc in pydoctor.  Epydoc is basically unmaintained upstream.  Also, it is only supported for Python 2, so it will reach its end of life along with Python 2 sometime in 2020.

glyph:
Probably the best way forward here is to vendor in epydoc and start porting it to py3.

mwhudson:
We probably only use a fairly small part for epydoc.

mthuurne:
Please have a look at #155, it imports those parts of epydoc that pydoctor uses into the pydoctor source tree.

hawkowl:
This is done. Thank you @mthuurne !

glyph:
Presumably “this” not being dropping epytext support? :)

hawkowl:
yes :)

PR #155, opened 2019-07-30
Epydoc import

mthuurne:
This is a potential fix for #147.

I tried to cut away all the parts of epydoc that aren't used by pydoctor. Likely more can be cut away with in-depth analysis, but I ran out of obvious candidates. The full epydoc is over 27k lines of code; less than 9k remain after cutting. Please let me know if anyone sees other parts that can be removed.

I tested by running pydoctor's unit tests and running pydoctor to generate Twisted's documentation. Both pass with warnings, but the warnings don't seem to have anything to do with my changes.


mthuurne:
The Travis build of the Twisted docs seems to be running a mixture of old and new code, which leads to `get_parser()` in `epydoc2stan` looking up the old name `epydoc.markup.epytext` and then not finding it because the module is now named `pydoctor.epydoc.markup.epytext`.

mthuurne:
Is the graphviz support used by pydoctor? That is a fair amount of code with relatively low test coverage.

glyph:
> Is the graphviz support used by pydoctor? That is a fair amount of code with relatively low test coverage.

I don't believe it is.

glyph:
cc @mwhudson since he is the one who suggested this, more or less :)

mwhudson:
Yeah happy with the idea, will try to look at it more on Monday. pydoctor
certainly doesn't deliberately use the graphvoz stuff...

mthuurne:
I removed the graph generation code. There are now 6.2k lines of epydoc left, excluding tests.

hawkowl:
I removed javadoc, since we don't use that. Epydoc and rst, we do kind of want, though.

mwhudson:
yay!

glyph:
*ZOMGGGGG*

Thank you @mthuurne for this amazing contribution!  And thank you @hawkowl for integrating it!

glyph:
(Very excited for pydoctor to get some much-needed maintenance in the coming weeks!)

PR #156, opened 2019-08-06
Remove more unused epydoc code

mthuurne:
I found more parts of epydoc that can be cut. In particular the removal of most of the `apidoc` module reduces the line count a lot.


hawkowl:
@mthuurne this does cause the twisted builder to break(which is set as allow fail): https://travis-ci.org/twisted/pydoctor/jobs/568348428#L258

mthuurne:
`ImportError: No module named epydoc.markup` makes me think there is something in the test setup causing this. I can't reproduce it on my development machine and I can't find any occurrence of `epydoc.markup` (rather than `pydoctor.epydoc.markup`) in the source tree either.

PR #157, opened 2019-08-06
Remove live object checker

mthuurne:
It seems this is a feature that was never finished:
- the `--livecheck` command line option doesn't do anything except logging an error
- the unit test is disabled
- the module itself is documented as "Extremely experimental code"

This code looks like it is particularly cumbersome to port to Python 3, since it contains a lot of assumptions on Python internals. I'd rather save some time by removing the feature for now. Should it be needed later, it can be ported and tested then.

@jelmer: On IRC, @mwhudson mentioned that you might have a use case for this feature. Is that still the case? And if so, do you object to the feature being removed for now?


jelmer:
Thanks for working on the Python3 port! I don't remember using livecheck (and am not sure I understand what it does), but I did rely on introspect-c-modules.

mthuurne:
The implementation of introspect-c-modules is in `model.py`, so it is unaffected by this removal.

mwhudson:
Ah OK I think I'd muddled up the live object checker and the introspect c module thing. Killing the former seems like a good idea then!

PR #161, opened 2019-11-04
Make unit tests pass on Python 3.6 (updated patch)

mthuurne:
This is an updated version of #152. It should be applied after #151/#160. For code that overlapped between @hawkowl's and @GaretJax's `ast` module ports, I picked hawkowl's changes. This did lead to a unit test regression I had to fix in a later commit, but overall the code is smaller, possibly due to using the external "astor" library.

With this patch, Twisted API doc generation still works under Python 2.7, but does not yet work under Python 3.6. It breaks on `src/twisted/python/_pydoctor.py` inside Twisted's source tree, which isn't Python 3 compatible yet. It uses the `compiler` and `urllib2` modules, for example.


hawkowl:
tests pass -- the coverage isn't complete, but I think I'm just going to accept there might be a few broken eggs in the name of progress. I tagged a stable version of pydoctor, so we don't have to stress too much.

let's go for it.

PR #166, opened 2019-11-17
Make Python2 and Python3 generate as-similar-as-possible outputs & add a script to see the differences

hawkowl:


glyph:
@hawkowl is this dead, in favor of @mthuurne's work?

mthuurne:
I think having the ability to easily diff between two runs of pydoctor is still useful.

Some of the changes are hard for me to evaluate though, since there are commit descriptions or unit tests that explain what issues they solve.

moshez:
Is this still relevant given that we do not support 2.7 anymore?

adiroiban:
I think that this can be closed as PY2 is no longer supported.

PR #167, opened 2019-11-18
Clean up various code with no coverage

mthuurne:
I went through the coverage reports and found some dead code.


wsanchez:
I suspect that un-removing the text passed to `NotImplementedError` may make this pass CI.

mthuurne:
> I suspect that un-removing the text passed to `NotImplementedError` may make this pass CI.

As far as I'm concerned, it has already passed CI:
- it's impossible to get patch coverage when you're removing code
- the codecov results are unreliable anyway, because sometimes the upload of results fails and then partial results will be used (for example, only Python 3 results) when comparing to a baseline that has all results or vice versa
- building Twisted's API docs fail because of problems inside Twisted; if pydoctor runs to completion I consider it a success

Anyway, the goal shouldn't be to figure out how to get code past CI, but CI should be a tool to get us to produce better code. In my opinion the text passed to `NotImplementedError` is redundant and as such prone to getting out of sync (not a theoretical problem either; it already happened), but if there is consensus among the maintainers that it should stay then I'll drop that commit from patch.

wsanchez:
@mthuurne: I've had this argument with @glyph many times.

That said, we need to pass CI.  The current failure in CI has nothing to do with any of your three bullet points and the `NotImplementedError` code change isn't related to the cleanup.

You'd have an approving review right now if this passed CI, is all I'm saying.

glyph:
I'm amenable to killing the `codecov/patch` check on this repo if folks think it's not useful so we don't get the red "X".

mthuurne:
Well, either CI should not produce false positives, or it should be allowed to override it. Either of those options is fine with me. But I'm not going to spend time working around a CI that is flagging errors where there are no actual problems in the code.

glyph:
Would you mind fixing up this PR to include the config change to codecov too, then? https://docs.codecov.io/docs/commit-status  As I said I'd be happy to land.

mthuurne:
> Would you mind fixing up this PR to include the config change to codecov too, then? https://docs.codecov.io/docs/commit-status As I said I'd be happy to land.

I can't find any current configuration file for `codecov` in pydoctor's Git, only the line in `tox.ini` that runs the `codecov` tool. Does that mean it runs in the default configuration and I should just add a new configuration file or does it mean it is configured on the web site instead of in the repository? Both are possible according to [the docs](https://docs.codecov.io/docs/codecov-yaml).

glyph:
There's no configuration on the web site, only a "team yaml" which is a yaml file that overrides a repo's configuration. I double checked and we don't use one of those.

mthuurne:
I can't get it to work.

I have added a `.codecov.yml` file. The contents of this file are accepted by their validation service. The file seems to be picked up by the `codecov` command line tool, since it added `yaml=.codecov.yml` to the URL in the Travis logs in this PR, while that part wasn't there in older PRs.

However, the `codecov/patch` status is still shown as failed, while informational mode is described as "If `true` is specified the resulting status will pass no matter what the coverage is or what other settings are specified."

So either informational mode doesn't do what I think it does, or the configuration somehow doesn't apply to the right context. That was the case in my first attempt, which only set `codecov/project` to informational mode, but the second attempt does configure `codecov/patch` as well.

wsanchez:
The requirement that `codecov/patch` pass before merges is in the repo settings in GitHub.  Looks like it's turned off now, so I can merge, but it also looks like `codecov/project` got turned off, which I don't think is intended…?

wsanchez:
Note also that `codecov/patch` is not a false positive.  The point of requiring that is that you shouldn't change code that isn't tested without first adding a test that covers the code.

I agree that some exceptions are reasonable, but what I don't get is why you think that code change is a necessary part of this PR.

mthuurne:
> Note also that `codecov/patch` is not a false positive. The point of requiring that is that you shouldn't change code that isn't tested without first adding a test that covers the code.

I don't consider `raise NotImplementedError` as actual code that is ever supposed to run, but as the legacy Python syntax for declaring abstract methods. Having a unit test verify what happens when you call an abstract method doesn't seem very useful to me. When using Python's abstract base classes, an exception would be thrown when you attempt to instantiate an abstract class, so getting line coverage on the abstract methods is not even possible in that case.

What is more useful to test is that abstract classes are not instantiated. The most effective tool for that is a static code checker. PyLint considers every class with a method body raising `NotImplementedError` as abstract and will warn if you try to instantiate it. For mypy I think you have to inherit from `abc.ABC` and decorate the methods with `@abstractmethod`. Since I assume we'll be using mypy for type checking anyway, we could adopt the syntax that mypy understands.

> I agree that some exceptions are reasonable, but what I don't get is why you think that code change is a necessary part of this PR.

The entire PR is the result of me going through areas of pydoctor that had zero coverage to check whether the code could be removed. The abstract methods containing redundant/outdated strings is just something I happened to spot in the process. I could put it into a separate PR if you prefer, but that seemed like unnecessary overhead to me.

It is also not this particular change that I'm attached to, but a general worry that if we (everyone, not just the Twisted ecosystem) don't question the outcome of algorithms we'll end up having no power over anything.

wsanchez:
@mthuurne I would expect that the code in Python that implements ABCs should have a test that ensures that an exception is raised.  If we have code that raises an exception, we should have such a test.  It doesn't matter if that code isn't supposed to run if the subclasses do it right; the point of units tests is to make sure all of the code is correct.

I'm very much in support of the idea that blindly relying on the toolchain to tell you how to write code is a bad move, but I don't think this is a good example of that.

mthuurne:
Maybe it's because I programmed in Java and C++ before Python, but in my mind an abstract method is a method that doesn't exist. Ideally, it should be impossible to call it, so its runtime behavior would be irrelevant. So if we can get mypy or another tool to report instantiation of abstract classes, that is sufficient in my eyes. An added benefit is that we wouldn't need to maintain any test code for it.

Note that when using ABCs, you cannot call an abstract method, since the instantiation of the class is blocked:
```py
from abc import ABC, abstractmethod

class C(ABC):
    @abstractmethod
    def f(self) -> None: ...

C()
```
When executed, this produces:
```
  File "testcase.py", line 7, in <module>
    C()
TypeError: Can't instantiate abstract class C with abstract methods f
```
And when checked with mypy:
```
$ mypy testcase.py 
testcase.py:7: error: Cannot instantiate abstract class 'C' with abstract attribute 'f'
```
By the way, when using the `@abstractmethod` decorator, we don't actually need a `raise NotImplementedError` in our code anymore. Does that help to make it acceptable to exclude this from unit testing?

PR #169, opened 2019-11-21
Do not disable marker before unordered list items

mthuurne:
Lists become hard to read without a marker before each item.


glyph:
@mthuurne could you share a quick before/after screenshot just so I can understand what this is doing at a glance? I'm inclined to approve & merge

mthuurne:
This is how it looks right now:

list-without-markers

And this is how it looks after the change:

list-with-markers

wsanchez:
Oh yeah, the new way is better. Or anyway, what I'd expect.

PR #176, opened 2019-12-18
Remove HTML to Stan parsing step when formatting epytext

mthuurne:
When using the epytext documentation format, the documentation goes through a lot of conversion steps:

docstring → epytext DOM → HTML → Stan → HTML

This PR replaces the "epytext DOM → HTML → Stan" chain by a single "epytext DOM → Stan" step.

My original intention was to do the same for ReStructuredText, but the amount of effort required to implement a Stan writer is rather significant due to the large number of node types that exists in Docutils. The HTML writer of Docutils is about 2000 lines of code and while outputting Stan would require slightly less code than outputting HTML strings, it would not be an order of magnitude less. And we don't have a lot of test coverage for this either, so that would have to be written as well. Since I don't actually use ReStructuredText in any of my projects, I'm not motivated to put in that amount of effort.

This PR does make Stan the format used for HTML fragments throughout pydoctor and moves the dealing with raw HTML strings to the edges: when interacting with Docutils and when writing the final output to disk. So even though the original goal wasn't met, it's still an improvement.

The generation of Twisted's API docs is about 5% faster on my PC with this PR. Not a huge difference, but still welcome, especially since that also reduces the time we have to wait for CI results.

I also added two unit tests, one for the colorization of Python code and one for epytext to HTML conversion. This should lead to a few percent higher overall coverage.

As a result of those unit tests, I found and fixed two bugs in Epydoc:
- the keyword list for colorizing Python code inside docstrings had some keywords stuck together: for example `return` and `break` were not colorized because they were in the list as a single `returnbreak` keyword
- literal blocks were indented one space too few

I don't know yet if it's feasible and desirable to also get rid of the epytext DOM intermediate representation and produce Stan directly from the epytext parser.

*Edit:* I also removed support for the `X{term}` syntax in epytext. This isn't used by Twisted and wasn't properly supported anyway (you could define terms but no index page existed for them).


mthuurne:
I realized that Stan is the predecessor and not the current name, but couldn't think of anything better. They're not really tags either, since technically a tag is only the name that identifies what kind of node we're dealing with and doesn't include attributes and children. Calling them trees is correct, but not very specific, since for example the Epytext DOM is also a tree.

I agree a consistent terminology would be better, but indeed I would like that to be done outside the scope of this PR. The module was already called `epydoc2stan`, so this is not a regression ;)

PR #178, opened 2019-12-22
Fix Sphinx inventory loading

mthuurne:
This PR fixes the following errors from the Twisted API docs build:
```
Failed to uncompress inventory from https://docs.python.org/2
Failed to uncompress inventory from https://docs.python.org/3
Failed to uncompress inventory from https://pyopenssl.readthedocs.io/en/stable
Failed to uncompress inventory from https://zopeinterface.readthedocs.io/en/latest
```
The first commit improves the error reporting, the second commit fixes the actual bug.

### DIFF
--- a/pydoctor/sphinx.py
+++ b/pydoctor/sphinx.py
@@ -157,7 +157,7 @@ class SphinxInventory(object):
                 'Failed to uncompress inventory from %s' % (base_url,))
             return ''
         try:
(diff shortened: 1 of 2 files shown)


mthuurne:
The third and fourth commit fix other issues I discovered while looking at the Sphinx inventory module and unit test.

PR #185, opened 2020-02-10
Small setup.py fixes to prepare for Python 3.

mikepurvis:
Per the discussion on #96, this package is ready for use with Python 3. This change makes a few small updates to the `setup.py` file:

- Drop the executable bit and shebang; these are unnecessary.
- Drop the external bin script in favour of using an entry_point.
- Update the classifiers to indicate Python 3.6 support (the version in the tox file, though I am currently testing on Ubuntu Xenial's Python 3.5 and all seems well).

With these changes merged, it would be great to have a new pydoctor released!


mikepurvis:
Can this get merged please so that it is at least easier to install a Python 3 supporting `pydoctor` from source?

mthuurne:
Sorry it took a while; I didn't have commit privs until today.

PR #188, opened 2020-02-29
Use OrderedDict to keep track of Documentable contents

mthuurne:
This was done with two variables until now: 'contents' as an unordered
dictionary and 'orderedcontents' as a list of the contents in order of
adding.


glyph:
@mthuurne You can use your newfound powers to do the minor merge-master update and merge yourself :)

PR #189, opened 2020-03-05
Provide source links for attributes

mthuurne:
For classes and functions pydoctor could already link to their location in a source viewer, but for attributes/variables this was not implemented yet.


pydoctor/test/test_astbuilder.py
```diff
@@ -582,6 +604,8 @@ class C:
         @ivar e: fifth
 
         @type f: C{str}
+
+        @type g: C{str}
```
wsanchez: In #190 you replaced `C{str}` with `string` FWIW.

mthuurne: Well spotted. This has to be changed to `string` as well; I'm doing so in the rebase.

PR #191, opened 2020-03-06
Update docstrings when __doc__ is assigned to

### LINKED ISSUE #102: Detect and use docstrings set after-the-fact.
Twisted has some code that does something like

``` python
if _PY3:
   def func(args):
       pass
else:
    def func(args):
        pass

func.__doc__ = """
func does some stuff.
```

to avoid duplicating a docstring. Unfortunately, pydoctor doesn't pick this up, so it appears that `func` is undocumented.

mthuurne:
Only simple assignments are supported, but for example `twisted.python.compat.reraise()` has proper documentation now, while it was listed as undocumented before.

Fixes #102

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -307,6 +307,40 @@ class ModuleVistor(ast.NodeVisitor):
         if not self._handleAliasing(target, expr):
             self._handleClassVar(target, annotation, lineno)
 
+    def _handleDocstringUpdate(self, targetNode, expr, lineno):
+        def warn(msg):
+            self.system.msg('ast', "%s:%d: %s" % (
+                    getattr(self.builder.currentMod, 'filepath', '<unknown>'),
+                    lineno, msg))
+
+        # Figure out target object.
+        full_name = node2fullname(targetNode, self.builder.current)
+        if full_name is None:
+            warn("Unable to figure out target for __doc__ assignment")
+            # Don't return yet: we might have to warn about the value too.
+            obj = None
+        else:
+            obj = self.system.objForFullName(full_name)
+            if obj is None:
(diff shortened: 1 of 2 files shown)


mthuurne:
In this case the coverage bot was right to complain ;)

pydoctor/astbuilder.py
```diff
@@ -307,6 +307,41 @@ def _handleAssignmentInClass(self, target, annotation, expr, lineno):
...
+            self.system.msg('ast', "%s:%d: %s" % (
+                    getattr(self.builder.currentMod, 'filepath', '<unknown>'),
+                    lineno, msg))
+
+        # Figure out target object.
+        full_name = node2fullname(targetNode, self.builder.current)
+        if full_name is None:
+            warn("Unable to figure out target for __doc__ assignment")
+            # Don't return yet: we might have to warn about the value too.
+            obj = None
```
wsanchez: Won't this cause the code to emit a "not a string" argument below? Since that's not caused by the source code, it may be a confusing/misleading warning.

One option is to put the "not a string" warning in the "if obj is not None" condition, or to set this to a sentinel value that isn't `None` (eg. `UNSET=object()` at the module scope) and skip the warning if that's the value.

mthuurne: `obj` is the object assigned to (left hand side of assignment), while "not a string" is reported if the value assigned (right hand side of assignment) is not a string. It is possible for both warnings to be logged for the same assignment and that is by design: if there are two problems in the assignment, both should be reported. But if the target cannot be figured out while the docstring is a simple string, there will only be a warning about the target.

mthuurne: I'll merge this now. I'm working on a new PR that improves error messages, so if there is an issue I can fix it there.

PR #193, opened 2020-03-25
Fix calls to zope.interface decorators that use *() that are actually None

hawkowl:


twm:
Shouldn't there be a test of this?

PR #196, opened 2020-05-23
Ignore documentables inside functions

### LINKED ISSUE #195: Inner functions documented like methods
The function `readBody` defines two inner functions. These are private helpers and didn't used to appear in pydoctor's output. Now they do, which is a bit confusing:

image

Compare:

- [19.10.0](https://twistedmatrix.com/documents/19.10.0/api/twisted.web.client.readBody.html)
- [20.3.0 rendering](https://twistedmatrix.com/documents/20.3.0/api/twisted.web.client.readBody.html)

mthuurne:
These cannot be called directly, so they should not be considered to
be part of the API.

In theory they could be returned from the function and then exposed,
but I don't think our aliasing support is powerful enough to support
that anyway.

Note that documented local variables were never included in the
output, but I added them to the test case to make sure that won't
happen in the future either.

Closes #195

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -75,6 +75,11 @@ class ModuleVistor(ast.NodeVisitor):
         self.builder.pop(self.module)
 
     def visit_ClassDef(self, node):
+        # Ignore classes within functions.
+        parent = self.builder.current
+        if isinstance(parent, model.Function):
+            return
+
         rawbases = []
         bases = []
         baseobjects = []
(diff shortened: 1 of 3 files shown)


mthuurne:
The `KeyError: '_DeprecatedProperty'` reported in the CI seems serious, but is triggered by a change in Twisted rather than something that changed in this PR, so it will have to be investigated and solved separately.

mthuurne:
I fixed the issue and updated the unit tests so they can catch this and similar issues in the future.

The problem was that the unit tests for basic functionality (`test_astbuilder.py`) only tested the base `System` implementation; `ZopeInterfaceSystem` was only tested for the functionality it added, but we also need to test whether the basic functionality still works on the extended system.

glyph:
Oof, looks like your other change conflicted :( sorry!

mthuurne:
No problem, it was an easy fix.

Issue #105, opened 2016-06-27
Remove TwistedSystem


adiroiban:
Since this is a Twisted specific thing it should be hosted by the Twisted project itself and not in pydoctor.

I am moving this to Twisted as part of https://twistedmatrix.com/trac/ticket/8515

Twisted PR https://github.com/twisted/twisted/pull/238

PR #198, opened 2020-06-22
Tell Travis to run twisted-apidocs check under Python 3.8

mthuurne:
Twisted trunk no longer supports Python 2.7, so we need a Python 3.x
to be able to execute its release script and parse its code.

I picked 3.8 since the ast module has optional support for parsing
type comments since Python 3.8, see #202.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
       env: TOX_ENV=test-py37
     - python: 3.8
       env: TOX_ENV=test-py38
-    - python: 2.7
+    - python: 3.8
       env: TOX_ENV=twisted-apidoc
 
   allow_failures:
--- a/tox.ini
+++ b/tox.ini
@@ -47,7 +47,7 @@ commands =
     ; Run current version against twisted trunk
     twisted-apidoc: rm -rf {toxworkdir}/twisted-trunk
     twisted-apidoc: git clone --depth 1 --branch trunk https://github.com/twisted/twisted.git {toxworkdir}/twisted-trunk
-    twisted-apidoc: /bin/sh -c "{toxworkdir}/twisted-trunk/bin/admin/build-apidocs {toxworkdir}/twisted-trunk/src {toxworkdir}/twisted-apidocs-build > {toxworkdir}/twisted-apidocs.log"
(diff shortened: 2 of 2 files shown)


mthuurne:
Don't be fooled by the green status; the checks are **not OK**.

The problem is that two versions of Twisted get mixed up:
- the most recent release on PyPI, as installed by `tox` into `site-packages`
- the Git `trunk` version of which the documentation is being extracted

The rule in `tox.ini` executes `twisted-trunk/bin/admin/build-apidocs`, which in turn imports `twisted.python._release.BuildAPIDocsScript` from `site-packages`, which then calls pydoctor. However, this combination is not guaranteed to work and is in fact broken at the moment, since `BuildAPIDocsScript` from the last release only works on Python 2.7 while the code in `trunk` can only be parsed under Python 3.

I see the following options to fix this:

1. Change the `tox.ini` to run pydoctor on and using Twisted's last release
2. Change the `tox.ini` to run pydoctor on and using Twisted from `trunk`
3. Move `twisted.python._release.BuildAPIDocsScript` out of the `twisted` namespace, for example into the `bin/admin` directory and make sure it doesn't use new Twisted features until they are present in a release.

Option 1 has the lowest chance of breaking, but at the risk of not discovering pydoctor issues that are triggered by changes in the Twisted code.

Option 2 is the inverse of 1: it will hit integration problems earlier, but at the cost of possibly breaking on issues unrelated to the PR it is testing.

Option 3 sits in between: pydoctor will extract documentation from Twisted `trunk` but will run the code from the last Twisted release.

My preference would be option 2, since the point of CI is to catch integration issues as early as possible. An alternative would be to run the `master` version of pydoctor as part of Twisted's CI.

Since this affects Twisted's release process, I'd like opinions from @hawkowl  and @glyph.

mthuurne:
I implemented option 2. Please review whether this is an acceptable solution.

mthuurne:
I changed the implementation of option 2: instead of setting `PYTHONPATH`, I now tell `tox` to install Twisted from Git.

The advantage of this new approach is that we can also run the unit tests against the trunk version of Twisted now, which will become relevant when we remove the outdated copy of `TwistedSystem` that pydoctor currently uses (see #105).

mthuurne:
I restored the `PYTHONPATH` solution in this PR. I'll create a separate PR that includes the `git+` URL solution and the removal of `TwistedSystem`.

mthuurne:
I did a bit of creative dependency management in #205, which uses the last Twisted release for Python 2.7 and trunk for 3.x. Since I think that solution is superior, I'm closing this PR. I'll leave the branch around for now, just in case.

PR #199, opened 2020-06-22
Improve handling of relative imports

mthuurne:
The code for handling relative imports was two generations behind:

- it supported implicit relative imports, which were dropped in Python 3
- it didn't properly support explicit relative imports: it matched on name instead of level, which usually works in practice but can find the wrong package


mthuurne:
> Dropping support for implicit relative imports should be noted in the changelog if this makes the 20.7.0 release, since that release still supports Python 2.7.

That's a good point. I think it shouldn't be a big issue to not have support for implicit relative imports, since that was considered a bit of a language misfeature anyway, but it should be mentioned.

PR #202, opened 2020-06-25
Use annotations from type comments

mthuurne:
In Python 3.8, the `ast` module got an option to return the contents of type comments. Since type comments are currently being added to Twisted, I think this is useful.

The middle commit adds the actual processing of type comments; the other two are cleanups in related code.


pydoctor/astbuilder.py
```diff
@@ -467,34 +476,60 @@ def visit_FunctionDef(self, node):
...
+        try:
+            return _AnnotationStringParser().visit(node)
+        except SyntaxError as ex:
+            builder = self.builder
+            builder.system.msg(
+                'parsing',
+                '%s:%s: %s' % (
+                    builder.currentMod.description,
+                    node.lineno,
+                    'syntax error in annotation: %s' % ex),
```
twm, line 510: Why the nested `%` interpolation?

pydoctor/test/test_astbuilder.py
```diff
@@ -844,6 +844,26 @@ def __init__(self):
     assert m.docstring == """module-level"""
     assert to_html(get_parsed_type(m)) == '<code>bytes</code>'
 
+@typecomment
+@systemcls_param
+def test_type_comment(systemcls, capsys):
+    mod = fromText('''
+    d = {} # type: Dict[str, int]
+    i = [] # type: ignore[misc]
```
twm, line 852: Here we are testing that type inference is done when the type comment is an ignore? If that's correct, perhaps mention that in the test docstring?

mthuurne: Actually it's only testing that `type: ignore` doesn't lead to an invalid type being reported. I was worried that `ignore` would be reported as the type, but Python 3.8's `ast` module will store `type: ignore` information in a separate per-module list. However, since the interface of `ast` has received incompatible changes before (like how line numbers are handled), I thought it was a good precaution to add a check that verifies that behavior for the future.

I'll add a comment explaining that.

pydoctor/astbuilder.py
mthuurne: It was copy-pasted code that I forgot to clean up once it worked. I'll replace it with something more readable.

PR #203, opened 2020-06-26
Fix docstring line numbers in Python 3.8

mthuurne:
In Python 3.8, `end_lineno` was added to contain the last line of a string, while `lineno` now contains the start line, unlike in older Python versions where it contained the end line. Therefore on 3.8 we have to skip the workaround that computes the start line from the end line, or we overcompensate and end up with incorrect line numbers for docstrings.


mthuurne:
The new unit test pointed out that PyPy was already putting the start line in `lineno`, so we now skip the workaround for PyPy as well.

Test coverage is reported as down because the two Python versions for which coverage data is collected are both ones where the workaround is active. We actually test more though.

PR #204, opened 2020-06-26
Improve reporting of missing docstring fields

mthuurne:
The exact line number is now reported, instead of the name of the object the docstring belongs to.

The priority of these messages was also increased, so they are now no longer silenced by a single `--quiet` option, as is used when extracting Twisted's API docs.


twm:
Shouldn't there be test coverage of this?

mthuurne:
> Shouldn't there be test coverage of this?

Yes; I'll add it.

PR #206, opened 2020-07-01
Prepare for release

mthuurne:
These are  documentation and metadata updates to prepare pydoctor for a release.


README.rst
```diff
@@ -2,57 +2,93 @@ pydoctor
...
 
-    python2 -m tox -e pyflakes
+You can see the full list of options using ``pydoctor --help``.
 
+Markup
+------
+
+pydoctor currently supports the following markup languages in docstrings:
+
+    `epytext`__ (default)
```
twm, line 41: The level of indentation here is putting this in a blockquote which doesn't seem right.

setup.py
```diff
@@ -1,12 +1,30 @@
...
     author='Michael Hudson-Doyle',
     author_email='micahel@gmail.com',
-    url='http://github.com/twisted/pydoctor',
+    maintainer='Maarten ter Huurne',
+    maintainer_email='maarten@boxingbeetle.com',
     description='API doc generator.',
+    long_description=long_description,
+    long_description_content_type='text/x-rst',
     license='MIT/X11',
+    url='http://github.com/twisted/pydoctor',
```
twm, line 23: http**s**://

tox.ini
```diff
@@ -19,7 +19,7 @@ passenv = *
 
 
 deps =
-    test-{py36,py37,py38},twisted-apidoc: git+https://github.com/twisted/twisted.git
+    test-{py35,py36,py37,py38},twisted-apidoc: git+https://github.com/twisted/twisted.git
```
twm, line 22: It would probably be faster to use the archive URL ``https://github.com/twisted/twisted/archive/trunk.zip`` instead of a full Git clone.

LICENSE.txt
```diff
@@ -1,4 +1,4 @@
-  Copyright 2006-2008 Michael Hudson mwh@python.net
+  Copyright 2006-2008 Michael Hudson <mwh@python.net>
```
twm, line 1: While adding copyright notices, perhaps both sections of this license file should have an additional notice that covers the recent changes? Say,

> Copyright 2006-2020 The pydoctor Authors

Then add an AUTHORS file with the full list?

Or list everyone right here, like in https://github.com/twisted/twisted/blob/trunk/LICENSE

LICENSE.txt
```diff
@@ -16,3 +16,30 @@ FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT,
 NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION
 WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
+
+
+Support for epytext and reStructuredText was imported from epydoc.
+This code can be found under pydoctor/epydoc/ and is licensed as follows:
```
twm, line 22: Should this say "originally licensed as follows" rather than "is licensed as follows"? The above license is what applies to the combined work and the modifications to the original made in this repository, no?

README.rst
mthuurne: You're right. It looked fine when I converted it to HTML using `rst2html5.py`, so I didn't notice the superfluous blockquote. I'll dedent it.

setup.py
mthuurne: Agreed.

tox.ini
mthuurne: That's probably a good idea, but I'm always worried about stale cached data. Since this doesn't impact end users, I think it can be changed after the release.

twm: I don't think it changes anything from a caching perspective, actually. I believe that Pip does the clone in a temporary directory that gets thrown away after building the distribution. FWIW I gave up running the full suite locally because it took so long on my slow internet connection.

I agree this doesn't need to be done before the release.

LICENSE.txt
mthuurne: I'd prefer to have a low-maintenance solution. Git already keeps track of author information, in much more detail, and is always up-to-date. Would it be sufficient to list the main author(s) in LICENSE and add a line like this?

> Copyright 2006-2020 Various contributors (see Git history)

LICENSE.txt
mthuurne: "Originally licensed" might suggest the code was re-licensed, which is not the case. The original code is still under the original license, modifications are under pydoctor's license. But most of the changes I made to the epydoc code are removing features that pydoctor doesn't need and replacing intermediate steps, plus a few bug fixes. So there is not any new feature or cluster of code that stands by itself, it's small modifications in many places.

tox.ini
mthuurne: I once worked at a company that had a poorly implemented corporate proxy which was happy to serve you stale cached data unless you used every possible HTTP header to tell it not to. It made me very paranoid about caches.

LICENSE.txt
twm: From a copyright perspective the contributors (human beings that did the work) and authors (copyright holders) may be different, though I've no real reason to think that is the case here. The specific wording I suggested was derived from the Go license, which seems a fairly common pattern:

https://github.com/golang/go/blob/a4ba411b19fa0111a3c8fe059fcf5489d3bd6bbf/LICENSE#L1

That project maintains separate [CONTRIBUTORS](https://github.com/golang/go/blob/master/AUTHORS) and [AUTHORS](https://github.com/golang/go/blob/master/AUTHORS) files.

hyper-h2 is another MIT-licensed project:

https://github.com/python-hyper/hyper-h2/blob/06aa5c9e63c4a8e0146c33f0abf14e8f7caf3945/LICENSE#L3

And Pip is similar:

https://github.com/pypa/pip/blob/e774eccf46f8465db707de108ea41a9cd9a41b66/LICENSE.txt#L1

In the absence of any clear Twisted policy on this, I think your suggestion is a fine one. What seems important to me is to avoid incorrect information, which a copyright notice that doesn't mention the additional contributors would be.

mthuurne:
Thanks for the review!

PR #211, opened 2020-07-10
Fix cross-references to builtin types in standard library

mthuurne:
Builtin types such as `str` were linked to the documentation page for builtin exceptions instead of the page for builtin types.

If we're going to do a 20.7.1 release, this fix should be included.

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -17,11 +17,6 @@ from twisted.web.template import Tag, tags
 from pydoctor.epydoc.markup import DocstringLinker, ParsedDocstring
 import pydoctor.epydoc.markup.plaintext
 
-try:
-    import exceptions
-except ImportError:
-    exceptions = builtins
-
 
 def _find_stdlib_dir():
     """Find the standard library location for the currently running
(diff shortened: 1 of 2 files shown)


pydoctor/epydoc2stan.py
```diff
@@ -20,11 +20,6 @@
 from pydoctor.epydoc.markup import DocstringLinker, ParsedDocstring
 import pydoctor.epydoc.markup.plaintext
 
-try:
-    import exceptions
-except ImportError:
-    exceptions = builtins
```
moshez, line 26: We should probably remove this cross-version compat shims, and just use `builtins` explicitly below.

mthuurne: The problem with using `builtins` is that since Python 3 it is no longer possible to tell which names represent exceptions and which represent non-exception types based on the module they're defined in; that's exactly what was causing the bug. The fix instead uses `issubclass()` to check whether a name belongs to an exception class.

mthuurne:
> Please remove the 2/3 compatibility shims and merge.

This was the only place where `ImportError` was caught when importing from the standard library and that code is removed in this PR. I don't know of any other 2/3 compatibility shims that could be removed.

PR #214, opened 2020-07-13
Add type annotations to pydoctor.sphinx module

mthuurne:
I picked `pydoctor.sphinx` as a first module to annotate since it's relatively self-contained, so I could go through the process without complications.

### DIFF
--- a/pydoctor/sphinx.py
+++ b/pydoctor/sphinx.py
@@ -283,7 +283,16 @@ class InvalidMaxAge(Exception):
     """
 
 
-def parseMaxAge(maxAge):
+def parseMaxAge(maxAge: str) -> Dict[str, int]:
+    """
+    Parse a string into a maximum age dictionary.
+
+    @param maxAge: A string consisting of an integer number
+        followed by a single character unit.
+    @return: A dictionary whose keys match L{datetime.timedelta}'s
+        arguments.
+    @raises: L{InvalidMaxAge} when a string cannot be parsed.
+    """
     try:
         amount = int(maxAge[:-1])
     except (ValueError, TypeError):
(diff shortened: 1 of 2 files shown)


pydoctor/sphinx.py
```diff
@@ -15,6 +18,12 @@
 from cachecontrol.caches import FileCache
 from cachecontrol.heuristics import ExpiresAfter
 
+if TYPE_CHECKING:
+    from pydoctor.model import Documentable
+else:
```
moshez, line 23: If you do the `__future__` import you won't need the `else` branch.

pydoctor/sphinx.py
```diff
@@ -23,15 +32,22 @@ class SphinxInventory:
...
         @param project_name: Dummy argument to stay compatible with
                              L{twisted.python._pydoctor}.
         """
-        self._links = {}
-        self.error = lambda where, message: logger(where, message, thresh=-1)
+        self._links: Dict[str, Tuple[str, str]] = {}
+        self._logger = logger
 
-    def update(self, cache, url):
+    def error(self, where: str, message: str) -> None:
```
moshez, line 47: Why is this method added here?

(Ditto for the `info` method below)

pydoctor/sphinx.py
```diff
@@ -7,6 +7,9 @@
 import shutil
 import textwrap
 import zlib
+from typing import (
```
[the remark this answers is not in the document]
mthuurne: That's a 3.7 feature, so we'd have to drop 3.6 support. Probably not worth it for that feature alone.

pydoctor/sphinx.py
mthuurne: Because adding annotations to a lambda expression is not possible. The alternatives would be an inner function or a relatively complex `Callable` annotation, both of which are harder to read than a method, in my opinion.

pydoctor/sphinx.py
mthuurne: This construct is here to break the import cycle, since `pydoctor.model` imports `SphinxInventory` from `pydoctor.sphinx`. mypy doesn't care about circular imports, but the Python VM and pytest do.

mthuurne:
Thanks for the review!

pydoctor/sphinx.py
mthuurne: Actually, you are right that the `else` branch could be dropped (while keeping the `if`) when using the `__future__` import. But I'd rather not do that right now to preserve Python 3.6 support.

PR #218, opened 2020-10-04
Update use of 3rd party type annotations

mthuurne:
Both Twisted and pytest ship with type annotations now, so we should use those in our mypy runs.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -7,6 +7,7 @@ doc_files = README.txt
 
 
 [tool:pytest]
+testpaths=pydoctor/test
 python_files=test_*.py
 addopts=--doctest-glob='*.doctest'
 doctest_optionflags=ELLIPSIS IGNORE_EXCEPTION_DETAIL
--- a/tox.ini
+++ b/tox.ini
@@ -62,6 +62,7 @@ deps =
     mypy-zope
     ; Libraries which include type annotations:
     hypothesis
+    pytest>=6.0.0
 
 commands =
     mypy                                       \
--- a/tox.ini
+++ b/tox.ini
@@ -19,7 +19,7 @@ passenv = *
 
 
 deps =
-    test-{py36,py37,py38,pypy3},twisted-apidoc: git+https://github.com/twisted/twisted.git
+    test,test-{py36,py37,py38,pypy3},twisted-apidoc: git+https://github.com/twisted/twisted.git
 
     test: coverage
     test: pytest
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -62,6 +60,8 @@ deps =
     mypy-zope
     ; Libraries which include type annotations:
     hypothesis
+    pytest>=6.0.0
```
graingert, line 63:
```suggestion
    pytest>=6.1.1
```

mthuurne: Why 6.1.1? If I read the ChangeLog correctly, the 6.0.0 release should also have the type annotations.

tox.ini
```diff
@@ -19,7 +19,7 @@ passenv = *
 
 
 deps =
-    test-{py36,py37,py38,pypy3},twisted-apidoc: git+https://github.com/twisted/twisted.git
+    test,test-{py36,py37,py38,pypy3},twisted-apidoc: git+https://github.com/twisted/twisted.git
 
     test: coverage
     test: pytest
```
graingert, line 25:
```suggestion
    test: pytest>=6.1.1
```

tox.ini
graingert: @mthuurne unlike production deps, there's no benefit in supporting a wide range of test deps. Might as well upgrade to the latest

setup.cfg
```diff
@@ -7,6 +7,7 @@ doc_files = README.txt
 
 
 [tool:pytest]
+testpaths=pydoctor/test
```
graingert, line 10: won't this skip your doctests?

graingert: ah you're not running inline doctests - they're all in `pydoctor/test/epydoc/*.doctest`

graingert (approved):

setup.cfg
mthuurne: Yeah, they are doctests in separate files because we inherited them from epydoc and the author thought doctests were a more convenient format than unit tests. That may have been true at the time, but now that we have pytest with less boilerplate and more detailed reporting of failures, I'm not sure doctests are still superior.

graingert: sorry I don't have permission to resolve comments

tox.ini
mthuurne: 6.0.0 represents the feature we depend upon, while 6.1.1 represents the time at which this dependency was added. The former feels more important to me. In practice, pip installs the latest version when you set up a virtual env, so I'm not sure the difference really matters.

PR #219, opened 2020-10-05
Add type annotations for unit tests

mthuurne:
My hope is that the calls made by the test cases to the code under test will act as a sanity check when adding type annotations to that code.

In any case, the process of adding the annotations to the unit tests led to several code and test cleanups already. I put these in separate commits, for easier reviewing.

### DIFF
--- a/pydoctor/driver.py
+++ b/pydoctor/driver.py
@@ -330,7 +330,6 @@ def main(args=sys.argv[1:]):
                 writerclass.__name__))
 
             writer = writerclass(options.htmloutput)
-            writer.system = system
             writer.prepOutputDirectory()
 
             if options.htmlsubjects:
--- a/tox.ini
+++ b/tox.ini
@@ -57,6 +57,7 @@ description = run mypy (static type checker)
 deps =
     mypy>=0.770
     mypy-zope
+    typing-extensions
     ; Libraries which include type annotations:
     hypothesis
 
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -15,7 +15,7 @@ import platform
 import sys
 import types
 from enum import Enum
-from typing import TYPE_CHECKING, Type
+from typing import TYPE_CHECKING, Optional, Type
 
 from pydoctor.sphinx import SphinxInventory
 
@@ -408,7 +408,7 @@ class System:
     # Not assigned here for circularity reasons:
     #defaultBuilder = astbuilder.ASTBuilder
     defaultBuilder: Type[ASTBuilder]
-    sourcebase = None
+    sourcebase: Optional[str] = None
 
     def __init__(self, options=None):
         self.allobjects = {}
--- a/pydoctor/sphinx.py
+++ b/pydoctor/sphinx.py
@@ -363,26 +363,6 @@ class IntersphinxCache:
             return None
 
 
-@attr.s(auto_attribs=True)
-class StubCache:
-    """
-    A stub cache.
-    """
-
-    _cache: Dict[str, bytes]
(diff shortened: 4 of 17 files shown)


pydoctor/epydoc/markup/__init__.py
```diff
@@ -108,16 +110,15 @@ def html2stan(html):
     stan.tagName = ''
     return stan
 
-def flatten(stan):
+def flatten(stan: Any) -> str:
```
graingert, line 113:
```suggestion
def flatten(stan: object) -> str:
```

graingert: Any is much wider than object

mthuurne: The problem with `object` is that as soon as `twisted.web.template.flattenString()` receives annotations, mypy will flag an error when passing an `object` as the second argument. While `Any` is not very descriptive, it is less fragile.

graingert: That's the advantage of using object here, you get notified of the type change straight away

graingert: You might want to set https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-any-explicit too

mthuurne: But it won't happen straight away: it will happen on some random PR, which then either has to include a fix for an unrelated issue or be put on hold until a fix has been merged.

mthuurne: pydoctor is almost 10k lines of code and very little of it has type annotations yet. The first priority is to get mypy coverage up to a level where it can detect inconsistencies. Making the ruleset  stricter is a good idea for the future, but doing that right would be counter-productive, in my opinion.

mthuurne: I think I found a workable compromise: accept only `Tag` objects. That gives mypy a specific type to work with, while we can always wrap strings, sequences etc. in a transparent tag to convert them to a `Tag` object.

pydoctor/driver.py
```diff
@@ -330,7 +330,6 @@ def main(args=sys.argv[1:]):
                 writerclass.__name__))
 
             writer = writerclass(options.htmloutput)
-            writer.system = system
```
graingert, line 333: ?

graingert: ah it's redundant

tox.ini
```diff
@@ -58,6 +58,7 @@ description = run mypy (static type checker)
 deps =
     mypy>=0.770
     mypy-zope
+    typing-extensions
```
graingert, line 61:
```suggestion
```

I don't think you need this - as it's only imported in `if TYPE_CHECKING:`

graingert (approved):
double check you need the typing-extensions dep, but otherwise good to go

pydoctor/driver.py
mthuurne: Unfortunately GitHub's review system doesn't do a great job presenting details from commit comments: you either have to review individual commits or you'll get the total change set without any context. It would be useful if they had some kind of `git blame` integration to show which commit is responsible for each changed line.

mthuurne:
> double check you need the typing-extensions dep, but otherwise good to go

I couldn't find anything in mypy's docs about providing typing-extensions for us, so I'm going to assume that if they are available, it's by accident rather than by design, in which case we should still be declaring the dependency. Note that the dependency is in the `mypy` tox env only, so it doesn't affect unit/system tests or the wheel.

PR #220, opened 2020-10-06
Add return type information only for functions

mthuurne:
I noticed a stray "Returns" at the end of module descriptions when using pydoctor's `master` branch, but not with the 20.7.1 release. So this is a regression, probably introduced as part of #209.

The problem is that an empty `FieldDesc` is assigned to `return_desc`, which ends up in the HTML output as a return value because it evaluates to `True` even when empty. Rather than trying to define exactly how empty a `FieldDesc` must be to be considered `False`, I decided it was clearer to leave `return_desc` on its default value of `None` for `Documentables` other than functions.

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -319,12 +319,12 @@ class FieldHandler:
             for name, value in annotations.items()
             }
         ret_type = formatted_annotations.pop('return', None)
-        ret_value = cls(obj, formatted_annotations)
+        handler = cls(obj, formatted_annotations)
         if ret_type is not None:
             return_type = FieldDesc()
             return_type.body = ret_type
-            ret_value.handle_returntype(return_type)
-        return ret_value
+            handler.handle_returntype(return_type)
+        return handler
 
     def redef(self, field):
         self.obj.system.msg(
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -314,15 +314,16 @@ class FieldHandler:
     @classmethod
     def from_ast_annotations(cls, obj: model.Documentable, annotations: Mapping[str, ast.expr]) -> "FieldHandler":
         linker = _EpydocLinker(obj)
-        annotations = {name: AnnotationDocstring(value).to_stan(linker)
-                       for name, value in annotations.items()}
-        return_type = FieldDesc()
-        try:
-            return_type.body = annotations.pop("return")
-        except KeyError:
-            pass
-        ret_value = cls(obj, annotations)
-        ret_value.handle_returntype(return_type)
+        formatted_annotations = {
+            name: AnnotationDocstring(value).to_stan(linker)
+            for name, value in annotations.items()
+            }
+        ret_type = formatted_annotations.pop('return', None)
+        ret_value = cls(obj, formatted_annotations)
+        if ret_type is not None:
+            return_type = FieldDesc()
+            return_type.body = ret_type
+            ret_value.handle_returntype(return_type)
         return ret_value
 
     def redef(self, field):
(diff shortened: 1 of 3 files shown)


pydoctor/epydoc2stan.py
```diff
@@ -316,13 +316,14 @@ def from_ast_annotations(cls, obj: model.Documentable, annotations: Mapping[str,
         linker = _EpydocLinker(obj)
         annotations = {name: AnnotationDocstring(value).to_stan(linker)
                        for name, value in annotations.items()}
-        return_type = FieldDesc()
-        try:
-            return_type.body = annotations.pop("return")
-        except KeyError:
-            pass
         ret_value = cls(obj, annotations)
```
mthuurne, line 319: Moving the `pop()` call below the constructor call means that the annotations can now include a `return` key while that wasn't possible in the original code. I don't think this will cause problems, but it is confusing and should be fixed.

I'm wondering though: if having a pseudo-argument named `return` is useful for annotations, is it also useful for epytext fields?

pydoctor/test/__init__.py
```diff
@@ -1,6 +1,22 @@
...
+
+class CapLog:
+    records: Sequence[LogRecord]
+
+class CaptureResult(NamedTuple):
+    out: str
+    err: str
+
+class CapSys:
+    def readouterr(self) -> CaptureResult: ...
```
graingert, line 22: might be better to use Protocols here, rather than concrete types - as that's what the pytest fixture types in `pytest.` will be.

otherwise you can use the exported types in `_pytest.logging.LogCaptureFixture`

```suggestion
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing_extensions import Protocol

    class CapLog(Protocol):
        records: Sequence[LogRecord]

    class CaptureResult(Protocol):
        out: str
        err: str

    class CapSys(Protocol):
        def readouterr(self) -> CaptureResult: ...
else:
    CapLog = CaptureResult = CapSys = None
```

mthuurne: That's probably a good idea, for #219 though.

pydoctor/epydoc2stan.py
```diff
@@ -314,15 +314,16 @@ def __init__(self, obj, annotations):
...
-            pass
-        ret_value = cls(obj, annotations)
-        ret_value.handle_returntype(return_type)
+        formatted_annotations = {
+            name: AnnotationDocstring(value).to_stan(linker)
+            for name, value in annotations.items()
+            }
+        ret_type = formatted_annotations.pop('return', None)
+        ret_value = cls(obj, formatted_annotations)
+        if ret_type is not None:
```
moshez, line 323: Instead of doing a `None` sentinel, couldn't we do it with exceptions?

```
try:
    ret_type = formatted_annotations.pop("return")
except KeyError:
    pass
else:
        return_type = FieldDesc()
        return_type.body = ret_type
        ret_value.handle_returntype(return_type)
```

moshez (approved):
One stylistic comment: either modify or not as you think right, and merge.

pydoctor/epydoc2stan.py
mthuurne: Since `ret_value` is needed whether the `return` annotation exists or not, its creation must happen on both flows. But if it is moved before the `pop()`, then the `return` annotation will be mixed into the parameter annotations, which isn't how `FieldHandler` is designed. So we'd either have to duplicate the `ret_value` initialization or put `ret_type = None` in the `except:` clause and keep the `if ret_type is not None:`, neither of which is appealing.

Actually `ret_value` is a poorly chosen name here: it's the return value of the method, but has nothing to do with the `return` statement, unlike `ret_type`. I'll rename it to `handler`.

PR #221, opened 2020-10-06
Include pydoctor version in 'generated by' info

mthuurne:
This is useful for pydoctor developers when diagnosing bad output. It is also useful for end users to evaluate whether it is worth re-generating API docs for the same release of their package with a newer release of pydoctor.

Twisted uses custom templates, so those will need to be updated to take advantage of this feature. But in the mean time, nothing will break: providing renderers and slots that are not used in the templates is harmless.


Julian (approved):
lgtm

PR #222, opened 2020-10-07
Fix issues in HTML rendering of function fields

mthuurne:
This fixes two noticeable rendering issues for functions:
- parameter descriptions should not be formatted as code (regression from #209)
- string literal annotations should be parsed ("unstringed") instead of output as strings

It also improves the way errors in annotations are handled:
- avoid a crash when a string literal annotation contains a statement rather than an expression
- if a string literal annotation fails to parse, render it as a string instead of discarding it

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -3,7 +3,7 @@
 import ast
 import sys
 from itertools import chain
-from typing import Iterator, Mapping, Optional, Tuple
+from typing import Iterator, Mapping, Tuple
 
 import astor
 from pydoctor import epydoc2stan, model
@@ -507,15 +507,15 @@ class ModuleVistor(ast.NodeVisitor):
             kwargs = base_args.kwarg
             if kwargs:
                 yield kwargs
-        def _get_all_ast_annotations() -> Iterator[Tuple[str, Optional[ast.expr]]]:
+        def _get_all_ast_annotations() -> Iterator[Tuple[str, ast.expr]]:
             for arg in _get_all_args():
-                yield arg.arg, arg.annotation
+                if arg.annotation:
+                    yield arg.arg, arg.annotation
             returns = func.returns
             if returns:
                 yield 'return', returns
-        return {name: value
-                for name, value in _get_all_ast_annotations()
-                if value is not None}
+        return {name: self._unstring_annotation(value)
+                for name, value in _get_all_ast_annotations()}
 
     def _unstring_annotation(self, node: ast.expr) -> ast.expr:
         """Replace all strings in the given expression by parsed versions.
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -517,12 +517,14 @@ class ModuleVistor(ast.NodeVisitor):
                 for name, value in _get_all_ast_annotations()
                 if value is not None}
 
-    def _unstring_annotation(self, node):
+    def _unstring_annotation(self, node: ast.expr) -> Optional[ast.expr]:
         """Replace all strings in the given expression by parsed versions.
-        Returns the resulting node, or None if parsing failed.
+        Returns the resulting node, or L{None} if parsing failed.
         """
         try:
-            return _AnnotationStringParser().visit(node)
+            expr = _AnnotationStringParser().visit(node)
+            assert isinstance(expr, ast.expr), expr
+            return expr
         except SyntaxError as ex:
             self.builder.currentMod.report(
                     f'syntax error in annotation: {ex}',
@@ -531,22 +533,32 @@ class ModuleVistor(ast.NodeVisitor):
 
 
 class _AnnotationStringParser(ast.NodeTransformer):
+    """Implementation of L{ModuleVistor._unstring_annotation()}."""
+
+    def _parse_string(self, value: str) -> ast.expr:
+        stmt, = ast.parse(value).body
+        if isinstance(stmt, ast.Expr):
+            expr = self.visit(stmt.value)
+            assert isinstance(expr, ast.expr), expr
+            return expr
+        else:
+            raise SyntaxError("expected expression, found statement")
 
     # For Python >= 3.8:
 
-    def visit_Constant(self, node):
+    def visit_Constant(self, node: ast.Constant) -> ast.expr:
         value = node.value
         if isinstance(value, str):
-            expr, = ast.parse(value).body
-            return self.visit(expr.value)
+            return self._parse_string(value)
         else:
-            return self.generic_visit(node)
+            const = self.generic_visit(node)
+            assert isinstance(const, ast.Constant), const
+            return const
 
     # For Python < 3.8:
 
-    def visit_Str(self, node):
-        expr, = ast.parse(node.s).body
-        return self.visit(expr.value)
+    def visit_Str(self, node: ast.Str) -> ast.expr:
+        return self._parse_string(node.s)
 
 def _infer_type(expr):
     """Infer an expression's type.
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -517,9 +517,10 @@ class ModuleVistor(ast.NodeVisitor):
                 for name, value in _get_all_ast_annotations()
                 if value is not None}
 
-    def _unstring_annotation(self, node: ast.expr) -> Optional[ast.expr]:
+    def _unstring_annotation(self, node: ast.expr) -> ast.expr:
         """Replace all strings in the given expression by parsed versions.
-        Returns the resulting node, or L{None} if parsing failed.
+        @return: The unstringed node. If parsing fails, an error is logged
+            and the original node is returned.
         """
         try:
             expr = _AnnotationStringParser().visit(node)
@@ -529,7 +530,7 @@ class ModuleVistor(ast.NodeVisitor):
             self.builder.currentMod.report(
                     f'syntax error in annotation: {ex}',
                     lineno_offset=node.lineno)
-            return None
+            return node
 
 
 class _AnnotationStringParser(ast.NodeTransformer):
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -556,6 +556,19 @@ class _AnnotationStringParser(ast.NodeTransformer):
         else:
             raise SyntaxError("expected expression, found statement")
 
+    def visit_Subscript(self, node: ast.Subscript) -> ast.Subscript:
+        value = self.visit(node.value)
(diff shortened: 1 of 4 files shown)


pydoctor/astbuilder.py
```diff
@@ -52,33 +52,6 @@ def node2fullname(expr, ctx):
         return None
 
 
-def _get_function_signature_annotations(func_ast: ast.FunctionDef) -> Mapping[str, ast.expr]:
```
moshez, line 55: This code seems to have been moved whole-sale into the method. This makes it harder to figure out if any of the logic changed. I'm not sure why it moved? Why couldn't the other function just call this as is?

pydoctor/astbuilder.py
```diff
@@ -512,36 +485,81 @@ def _annotation_from_attrib(self, expr, ctx):
...
+            if kwargs:
+                yield kwargs
+        def _get_all_ast_annotations() -> Iterator[Tuple[str, ast.expr]]:
+            for arg in _get_all_args():
+                if arg.annotation:
+                    yield arg.arg, arg.annotation
+            returns = func.returns
+            if returns:
+                yield 'return', returns
+        return {name: self._unstring_annotation(value)
```
moshez, line 517: Couldn't this just have been

```
return {name: self._unstring_annotation(value) for name, value in get_function_signature_annotation(func).items()
```
And not required moving the whole function in here?

Behavioral change diffs (even regression fix ones) are easier to review when they're separated from a zero-behavior-change refactoring review.

pydoctor/astbuilder.py
```diff
@@ -512,36 +485,81 @@ def _annotation_from_attrib(self, expr, ctx):
...
+    def _unstring_annotation(self, node: ast.expr) -> ast.expr:
         """Replace all strings in the given expression by parsed versions.
-        Returns the resulting node, or None if parsing failed.
+        @return: The unstringed node. If parsing fails, an error is logged
+            and the original node is returned.
         """
         try:
-            return _AnnotationStringParser().visit(node)
+            expr = _AnnotationStringParser().visit(node)
+            assert isinstance(expr, ast.expr), expr
```
moshez, line 527: When is this `assert` expected to fail?
I'm not sure I understand the reason for this diff

pydoctor/astbuilder.py
```diff
@@ -512,36 +485,81 @@ def _annotation_from_attrib(self, expr, ctx):
...
+            return expr
         except SyntaxError as ex:
             self.builder.currentMod.report(
                     f'syntax error in annotation: {ex}',
                     lineno_offset=node.lineno)
-            return None
```
moshez, line 525: This won't trigger for the `AssertionError`, so is this supposed to be triggered from the `expr = _Annotation...` line? In this case, let's have the `except` only around the line and use `else`.

pydoctor/astbuilder.py
```diff
@@ -512,36 +485,81 @@ def _annotation_from_attrib(self, expr, ctx):
...
 
 
 class _AnnotationStringParser(ast.NodeTransformer):
+    """Implementation of L{ModuleVistor._unstring_annotation()}."""
+
+    def _parse_string(self, value: str) -> ast.expr:
+        stmt, = ast.parse(value).body
+        if isinstance(stmt, ast.Expr):
+            expr = self.visit(stmt.value)
+            assert isinstance(expr, ast.expr), expr
```
moshez, line 543: This seems to be the same assertion in the call above. Why are both necessary?

Also, I think this `assert` depends on the Python parser only returning expressions, and not statements, from parsing annotations. In that case, I don't think this should be an assert, I think we should throw an explicit `SyntaxError`, since this is a problem with the *caller*, not our internal logic.

What do we expect to happen with code like

```
def foo(a: "raise ValueError()"):
     pass
```

This is legal Python code (it will run and produce valid results), but will cause an `AssertionError` here. This does not seem like desirable behavior?

pydoctor/astbuilder.py
```diff
@@ -512,36 +485,81 @@ def _annotation_from_attrib(self, expr, ctx):
...
-            return None
+            return node
 
 
 class _AnnotationStringParser(ast.NodeTransformer):
+    """Implementation of L{ModuleVistor._unstring_annotation()}."""
+
+    def _parse_string(self, value: str) -> ast.expr:
+        stmt, = ast.parse(value).body
+        if isinstance(stmt, ast.Expr):
```
moshez, line 541: I think `ast.Expr` and `ast.expr` are aliases, but it would probably be better to be consistent.

pydoctor/astbuilder.py
```diff
@@ -512,36 +485,81 @@ def _annotation_from_attrib(self, expr, ctx):
...
                     lineno_offset=node.lineno)
-            return None
+            return node
 
 
 class _AnnotationStringParser(ast.NodeTransformer):
+    """Implementation of L{ModuleVistor._unstring_annotation()}."""
+
+    def _parse_string(self, value: str) -> ast.expr:
+        stmt, = ast.parse(value).body
```
moshez, line 540: Can we surround only this with the `try/except` SyntaxError?

pydoctor/astbuilder.py
```diff
@@ -512,36 +485,81 @@ def _annotation_from_attrib(self, expr, ctx):
...
+    def visit_Constant(self, node: ast.Constant) -> ast.expr:
         value = node.value
         if isinstance(value, str):
-            expr, = ast.parse(value).body
-            return self.visit(expr.value)
+            return self._parse_string(value)
         else:
-            return self.generic_visit(node)
+            const = self.generic_visit(node)
+            assert isinstance(const, ast.Constant), const
```
moshez, line 556: Why is this assert here? A potential bug in `generic_visit`?

moshez (changes requested):
I left some comments, but the one thing I want to make sure:

* Add test for some behavior on weird stringy type definitions
* Change the `assert` in `visit_Constant`: either remove it if there is no way for user input to trigger it (or if it's defensive coding against a specific likely bug document the bug and make sure there's also a test that makes an effort to dig it out), or change it to a `SyntaxError` if user input can trigger it.

pydoctor/astbuilder.py
mthuurne: It was moved in commit 3ac25183, which doesn't change the logic. It only becomes hard to figure out when reviewing all changes at once...

In any case, the reason for the move is to be able to call `_unstring_annotation()`, which is a method of `ModuleVistor`.

The only thing `_unstring_annotation()` needs from `ModuleVisitor` is `self.builder.currentMod.report()`, so an alternative would be to pass the logger as an argument instead. But that is not the style in most of the pydoctor code is currently written.

pydoctor/astbuilder.py
mthuurne: Without making the function into a method, there is no `self`, so I don't see how that would work.

I agree with separating the changes, but that's why I put the move in a commit by itself. I'm not sure what more I could have done.

pydoctor/astbuilder.py
mthuurne: It's not expected to fail; it is expressing a non-trivial property of the code, both for the benefit of the reader and of mypy.

Annotations are always expressions, but there are more types of AST nodes, such as statements. So from the type system's perspective, it's not certain that `visit()` will return an expression. But we know that `_AnnotationStringParser` is written in such a way that it will only return an expression.

pydoctor/astbuilder.py
mthuurne: Yes, it will only catch `SyntaxError` raised by `_AnnotationStringParser`. We can indeed use an `else:` clause; if that makes the code easier to read, it's worth one extra line.

pydoctor/astbuilder.py
mthuurne: > This seems to be the same assertion in the call above. Why are both necessary?

It's a similar assertion, but not redundant, since there are a few intermediate steps inside `NodeTransformer` before the value we return here ends up in `_unstring_annotation()`. Also for mypy we need to have either an assert or cast here to be able to return `expr`. When something is not trivially true, I prefer asserts over casts, since we want to know when theory and practice don't align.

> Also, I think this `assert` depends on the Python parser only returning expressions, and not statements, from parsing annotations. In that case, I don't think this should be an assert, I think we should throw an explicit `SyntaxError`, since this is a problem with the _caller_, not our internal logic.

The `if isinstance(stmt, ast.Expr):` check guarantees that if the top-level node is a statement that is not an expression evaluation, we do in fact raise `SyntaxError`.

> What do we expect to happen with code like
> 
> ```
> def foo(a: "raise ValueError()"):
>      pass
> ```
> 
> This is legal Python code (it will run and produce valid results), but will cause an `AssertionError` here. This does not seem like desirable behavior?

This will result in `isinstance(stmt, ast.Expr)` being false and therefore `SyntaxError` will be raised. The `test_bad_string_annotation()` test case checks this, when `pass` is used as the annotation.

This was in fact an error that existed before, but caught because mypy forced me to check everywhere that the AST nodes we expect to be expressions can only be expressions, even on invalid inputs.

pydoctor/astbuilder.py
mthuurne: They are not the same thing: `ast.Expr` is a statement that contains an expression, while `ast.expr` is the base class for all expressions.

For example, in AST for the following code:
```py
print('hello')
f(2)
```
The second statement `s` is of type `ast.Expr` node and `s.value` contains the expression `f(2)` which is a subsclass of `ast.expr`.

pydoctor/astbuilder.py
mthuurne: No, we need to raise `SyntaxError` ourselves as well, in case the parsed code is valid Python but not an expression.

pydoctor/astbuilder.py
mthuurne: This one is mostly for mypy. If `generic_visit()` changes the type of node, that would indeed be a bug in `NodeTransformer`, I think. So I could change this into a cast, but I still prefer an assert for more certainty.

mthuurne:
> * Add test for some behavior on weird stringy type definitions

I added a test case for an annotation string containing more than one statement, since that wasn't handled yet. There already was a test case for an annotation string containing a non-expression statement.

> * Change the `assert` in `visit_Constant`: either remove it if there is no way for user input to trigger it (or if it's defensive coding against a specific likely bug document the bug and make sure there's also a test that makes an effort to dig it out), or change it to a `SyntaxError` if user input can trigger it.

User input cannot trigger it. But mypy needs either an assert or a cast to guide it, so removing it completely is not an option. I could change it to a cast, but in my opinion that doesn't improve the code.

moshez (approved):

Issue #223, opened 2020-10-08
Difficulties with --docformat=restructuredtext and links


tristanlatr:
Hello,

Thank for this software, it really produce a clear website.

I'm trying to generate documentation from ReST docstrings and I cant make the exemple from the reStructuredText docs work... ([here](https://docutils.sourceforge.io/docs/user/rst/quickref.html#hyperlink-targets))

```
External hyperlinks, like Python_.

.. _Python: http://www.python.org/
```

It outputs everything as plain text, but when I remove the linking part it outputs as normal text.  
Here is the warning I get : `bad docstring: TypeError: sequence item 0: expected str instance, bytes found`

Any help would be much appreciated !

Thanks

mthuurne:
I'll do a more thorough investigation tomorrow, but my first guess is that this is a bug that was introduced in the porting from Python 2 to 3, since Python 2 considered `bytes` to be a string type but Python 3 does not. If that is the case, it should be fairly easy to fix.

tristanlatr:
Thanks,

Here is another error I get when setting only a standalone link as the docstring:

```
bad docstring: RuntimeError: dictionary changed size during iteration
```

mthuurne:
I fixed both issues you found. If you'd like to test the fixes, you can find them in the [20.7.x bugfix branch](https://github.com/twisted/pydoctor/commits/20.7.x):
```
pip install git+https://github.com/twisted/pydoctor.git@20.7.x
```
The ReST support doesn't get as much exposure as epytext support and unfortunately the unit tests are incomplete. If you find more issues, please keep reporting them.

tristanlatr:
Hello, 

Thanks it looks like it's working but I can't make the inline links work

```
`GitHub<https://github.com/mfesiem/msiempy>`_
```

Also the python syntax hightlight doesn't seems to work, do we need to enable it somehow? Well... the HTML class are attributed to the right portion of the Python code but it seems there is no CSS to set a special colour. 

See the docs I'm working on here: https://mfesiem.github.io/docs/test/msiempy/msiempy.html

Thanks for your help

mthuurne:
It seems that the CSS for syntax highlighting was lost in [a big style update in 2015](https://github.com/twisted/pydoctor/commit/620c5070d1ae61d552f6aa542a6b2d9946bd0166). I'll resurrect it.

mthuurne:
The syntax highlighting is back, see #226 for details.

[...]

PR #224, opened 2020-10-09
Fix parsing of docstrings in restructured text format

mthuurne:
Some things were overlooked in the porting to Python 3 and we didn't have enough unit test coverage to detect that. See #223 for details.


twm (approved):
LGTM 👍

PR #226, opened 2020-10-13
Add CSS for syntax highlighting and doctest blocks

mthuurne:
The styling for this got lost in [the big makeover of 2015](https://github.com/twisted/pydoctor/commit/620c5070d1ae61d552f6aa542a6b2d9946bd0166).

I tried resurrecting the old CSS that originated from epydoc, but it didn't match very well with the rest of our styling. So I made new rules instead, using some existing colors from our CSS and using GitHub and Kate as inspiration for new styles.

### DIFF
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -241,3 +241,36 @@ body.private-hidden #splitTables .private, body.private-hidden #childList .priva
 .deprecationNotice {
     margin: 10px;
 }
+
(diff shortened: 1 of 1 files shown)


mthuurne:
This is what it looks like in `twisted.trial.util`:

image

And in `twisted.words.xish.domish.Element`:

image

Julian (approved):
Nice, definite improvement, lgtm.

PR #227, opened 2020-10-15
Fix reported source location in error messages

mthuurne:
This fixes two issues with error reporting:
- when a function or class is reparented, we now report the original source file (where the docstring resides) instead of the new parent's source file
- the reported line number for cross-reference errors is now the line number of the cross-reference itself, rather than the line number of the documented object

I only managed to get the second item to work for epytext: for restructured text I did figure out how it should work in theory, but in practice the line number reported by docutils is always `None`. Instead of the exact line we now report the start of the docstring, which shouldn't be too far off in most cases.

What triggered me to make these fixes was this error reported on Twisted:
```
src/twisted/logger/__init__.py:22: invalid ref to 'twisted.python.log.ILegacyLogObserver' resolved as 'ILegacyLogObserver'
```
which should actually read:
```
src/twisted/logger/_legacy.py:25: invalid ref to 'twisted.python.log.ILegacyLogObserver' resolved as 'ILegacyLogObserver'
```

### DIFF
--- a/pydoctor/epydoc/markup/restructuredtext.py
+++ b/pydoctor/epydoc/markup/restructuredtext.py
@@ -393,8 +393,10 @@ class _EpydocHTMLTranslator(HTMLTranslator):
         if m: text, target = m.groups()
         else: target = text = node.astext()
         label = tags.code(text)
+        # TODO: 'node.line' is None for some reason.
+        lineno = 0
         try:
-            url = self._linker.resolve_identifier_xref(target)
+            url = self._linker.resolve_identifier_xref(target, lineno)
         except LookupError:
             xref = label
         else:
--- a/pydoctor/epydoc/markup/restructuredtext.py
+++ b/pydoctor/epydoc/markup/restructuredtext.py
@@ -394,6 +394,7 @@ class _EpydocHTMLTranslator(HTMLTranslator):
         else: target = text = node.astext()
         label = tags.code(text)
         # TODO: 'node.line' is None for some reason.
+        #       https://github.com/twisted/pydoctor/issues/237
         lineno = 0
         try:
             url = self._linker.resolve_identifier_xref(target, lineno)
--- a/pydoctor/test/test_epydoc2stan.py
+++ b/pydoctor/test/test_epydoc2stan.py
@@ -402,6 +402,11 @@ def test_EpydocLinker_resolve_identifier_xref_order(capsys: CapSys) -> None:
 
 
 def test_xref_not_found_epytext(capsys: CapSys) -> None:
+    """
+    When a link in an epytext docstring cannot be resolved, the reference
+    and the line number of the link should be reported.
+    """
+
     mod = fromText('''
     """
     A test module.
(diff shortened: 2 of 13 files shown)


pydoctor/test/test_epydoc2stan.py
```diff
@@ -395,7 +395,41 @@ class C:
...
     linker = epydoc2stan._EpydocLinker(mod)
 
-    url = linker.resolve_identifier_xref('socket.socket')
+    url = linker.resolve_identifier_xref('socket.socket', 0)
 
     assert 'https://docs.python.org/3/library/socket.html#socket.socket' == url
     assert not capsys.readouterr().out
+
+
+def test_xref_not_found_epytext(capsys: CapSys) -> None:
```
adiroiban, line 404: It would be nice to have a docstring explaining the purpose of this test.

```suggestion
def test_xref_not_found_epytext(capsys: CapSys) -> None:
   """
   When reporting an epytext error, the report contains the line number for the reference that triggered the error and not the start line number of the docstring.
   """
```

pydoctor/test/test_templatewriter.py
```diff
@@ -69,8 +69,10 @@ def test_basic_package(tmp_path: Path) -> None:
     w.writeDocsFor(root, False)
     w.writeModuleIndex(system)
     for ob in system.allobjects.values():
-        if ob.documentation_location is model.DocLocation.OWN_PAGE:
-            assert (tmp_path / f'{ob.fullName()}.html').is_file()
+        url = ob.url
+        if '#' in url:
```
adiroiban, line 73: I don't understand this change

pydoctor/test/test_packages.py
```diff
@@ -47,3 +47,6 @@ def test_allgames() -> None:
     # InSourceAll is not moved into mod2, but NotInSourceAll is.
     assert 'InSourceAll' in system.allobjects['allgames.mod1'].contents
     assert 'NotInSourceAll' in system.allobjects['allgames.mod2'].contents
+    moved = system.allobjects['allgames.mod2'].contents['NotInSourceAll']
```
adiroiban, line 50: It would be nice to have a docstring. I find it hard to read the test.

I see that the test data is at https://github.com/twisted/pydoctor/blob/master/pydoctor/test/testpackages/allgames/mod2.py

and that for pydoctor mod2 is considered as being sourced from mod1.py

adiroiban (approved):
Looks good. I think this is a big improvement. Thanks.

I have left only minor comments.

Feel free to address or ignore them and merge on the PR is green.

Thanks again

pydoctor/test/test_epydoc2stan.py
mthuurne: I'll add one, also for the other new test case.

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -393,8 +393,10 @@ def visit_title_reference(self, node):
         if m: text, target = m.groups()
         else: target = text = node.astext()
         label = tags.code(text)
+        # TODO: 'node.line' is None for some reason.
```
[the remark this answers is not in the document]
mthuurne: Done. Issue is #237.

pydoctor/test/test_templatewriter.py
mthuurne: The old code would construct a file path from the full name. The new code takes the opposite approach and deconstructs the relative URL into a file path, by stripping off the fragment part.

pydoctor/test/test_packages.py
mthuurne: I've added some explanations in the test case.

The new checks basically verify the "when a function or class is reparented, we now report the original source file (where the docstring resides) instead of the new parent's source file" item from the PR summary.

PR #228, opened 2020-10-16
Perform sanity checks on field arguments

mthuurne:
The following sanity checks are added:
- Warn about missing parameter name in `@param` and `@type` fields
- When parameter names in `@param` and `@type` fields include asterixes, warn and strip them
- Warn about missing exception type in `@raise` fields
- Warn on non-empty `@return` and `@rtype` arguments

These checks can spot mistakes in docstrings that can lead to inferior HTML output. In the case of parameters, inconsistencies between documented names and actual names can cause problems when merging the information extracted from annotations with the information from the docstring.

### DIFF
--- a/pydoctor/epydoc/markup/__init__.py
+++ b/pydoctor/epydoc/markup/__init__.py
@@ -36,9 +36,11 @@ each error.
 """
 __docformat__ = 'epytext en'
 
+from typing import List
 import re
 
-from twisted.web.template import XMLString, flattenString
+from twisted.python.failure import Failure
+from twisted.web.template import Tag, XMLString, flattenString
 
 ##################################################
 ## Contents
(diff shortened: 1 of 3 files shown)


twm (approved):
This looks great &mdash; much more helpful messages.. :+1: I'm looking forward to fixing these issues in Twisted's docs!

mthuurne:
> I'm looking forward to fixing these issues in Twisted's docs!

I have a work-in-progress PR for Twisted that fixes issues reported by these checks and additional checks that aren't ready for merging yet: twisted/twisted#1453

The main additional check that I'm working on is reporting when a non-existing parameter is documented. This is mostly working, but it generates false positives when constructor arguments are documented on the class instead of on a method (`__init__()` usually).

PR #230, opened 2020-10-21
enable tox bootstrap

graingert:


mthuurne:
As explained by @graingert on IRC, this forces whatever version of tox is initially started to bootstrap a configured version of tox. So it gives us more control over the version of the build/test tooling and even though we don't need newer functionality right now, having that ability may become useful in the future.

mthuurne (approved):

PR #232, opened 2020-10-23
apply setup-py-upgrade

graingert:


pydoctor/__init__.py
```diff
@@ -1,5 +1,9 @@
 """PyDoctor, an API documentation generator for Python libraries."""
 
+__all__ = ["__version__"]
+
 from ._version import __version__
 
-__all__ = ["__version__"]
+
+def _setuptools_version():
+    return __version__.public()
```
mthuurne, line 9: It would be good to have type annotations on all new code: (even if it's as isolated as this fragment)
```diff
-def _setuptools_version():
-    return __version__.public()
+def _setuptools_version() -> str:
+    return __version__.public() # type: ignore[no-any-return]
```

setup.cfg
```diff
@@ -1,16 +1,63 @@
...
+packages = find:
+setup_requires = incremental
+python_requires = >=3.6
+install_requires =
+    incremental
+    appdirs
+    CacheControl[filecache]
+    Twisted
+    requests
+    astor
```
mthuurne, line 39: Please add `attrs`: that was missing in `setup.py` as well, but since pydoctor imports `attrs` directly, it should declare it and not rely on getting it indirectly via Twisted.

mthuurne (approved):
I added two small changes that I'd like to be applied, but other than that I think it's good to merge. Thanks!

PR #235, opened 2020-10-27
Switch to GitHub Actions for CI

mthuurne:
The latency of builds on Travis has become unworkable: it can take hours before PRs are tested. This may have something to do with the recent merge of travis-ci.org into travis-ci.com: my guess is that their total capacity is insufficient and/or the open source jobs are given a very low priority.

Note that Travis is still listed as a required check for this PR. I think this is due to the branch protection configuration on GitHub still having Travis enabled. I did remove all Travis config files from the source tree.

Release automation isn't covered by this PR, but all other CI functionality we currently have is migrated.


adiroiban:
@mthuurne 

I have removed Travis-CI as a required from here

image

You should also have access to the branch protection page as I see you as an admin for the repo.

.github/workflows/system.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  system_tests:
+    runs-on: ubuntu-latest
```
adiroiban, line 11: Note that in GitHub Action, Ubuntu latest is not yet ... 20.04 ... but I think that is fine for pydoctor and eventually `ubuntu-latest` will get there,

Note that you can also have all your tests into a single workflow and set a custom name for each job.
The test will still be run in parallel.

```suggestion
    name: name: System tests
    runs-on: ubuntu-latest
```

The main advantage is that you have all the test related info for GitHub Action into a single file.

The disadvantage is that if a job fail, when you retry it , it will run all the others from the same workflow.

.github/workflows/unit.yaml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  unit_tests:
+    runs-on: ubuntu-20.04
+
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9, pypy3]
+        include:
+        - python-version: 3.8
+          publish-coverage: true
```
adiroiban, line 18: I guess that for pydoctor this works as there is no python specific code.

But you should be able to publish all version to codecov and it will take care of merging then.

adiroiban (approved):
All good. Thanks. Only very minor comment ...and most probably you are already aware of all the things I mentioned in the comments.

I just wanted to make sure this is configured as you like.

mthuurne:
> I have removed Travis-CI as a required from here
> 
> ...
> 
> You should also have access to the branch protection page as I see you as an admin for the repo.

Thanks.

I had found that page already, but didn't want to actually remove Travis until its replacement was ready to merge, since slow CI is still better than no CI.

.github/workflows/unit.yaml
mthuurne: There are tests that only run on specific Python versions, but those are active in 3.8.

In Travis coverage uploads were also tied to a specific Python version. It seems it was introduced in commit 32fe9bcd, but that was 4 years ago, so I can imagine you don't remember the details anymore.

.github/workflows/system.yaml
mthuurne: > Note that in GitHub Action, Ubuntu latest is not yet ... 20.04 ... but I think that is fine for pydoctor and eventually `ubuntu-latest` will get there,

I wasn't sure whether Python 3.9 was available on 18.04, so for the unit tests I stuck to the version Twisted uses, which was version 20.04.

> Note that you can also have all your tests into a single workflow and set a custom name for each job.
> The test will still be run in parallel.
> 
> The main advantage is that you have all the test related info for GitHub Action into a single file.

How much duplication could be removed by merging the files? If it's only the `on:` block, it doesn't help that much, but if the setup, install tox and system info logging steps could be included only once it would simplify things. I wasn't sure that would be an option though, because the unit test should use a matrix while doing that for the system test and static code checks would be overkill.

> The disadvantage is that if a job fail, when you retry it , it will run all the others from the same workflow.

pydoctor currently doesn't have any flaky tests and I'd like to keep it that way ;) But the CI environment or services we contact can also be flaky, so we might sometimes need reruns nevertheless.

.github/workflows/unit.yaml
adiroiban: I think that the thing `test-py27-codecov-travis` was done not for coverage but for PyPI publishing.

so this is not about publish coverage, but about publish the PyPI package.

I think that you can remove the whole `include` section.

GitHub Action has a much nicer API and tools and we no longer need that hack.

Issue #212, opened 2020-07-10
Irregularities in standard library URLs


mthuurne:
The `stdlib_doc_link_for_name()` function in pydoctor assumes that the URL of a class or function in the standard library can be computed from its full name. However:

- the documentation for the module `xml.sax.xmlreader` is in `xml.sax.reader.html` (not `xmlreader`)
- the class `logging.FileHandler` is documented in `logging.handlers.html` (not `logging.html`)
- there are probably more cases like this

Should we hardcode such irregular mappings individually or should we use a Sphinx objects inventory for the standard library instead?

twm:
Intersphinx, definitely.

mthuurne:
Now that #236 is merged, this should be fixed.

PR #236, opened 2020-10-27
Drop stdlib_doc_link_for_name()

mthuurne:
Give up trying to construct deeplinks to the standard library, since there is no consistent mapping. Instead, rely on intersphinx inventories. See #212 for details.


pydoctor/test/test_epydoc2stan.py
```diff
@@ -370,6 +371,13 @@ def test_EpydocLinker_resolve_identifier_xref_intersphinx_link_not_found(capsys:
     assert expected == captured
 
 
+class FakeInventory:
+
+    def getLink(self, name: str) -> Optional[str]:
+        return {
+            'socket.socket': 'https://docs.python.org/3/library/socket.html#socket.socket',
+            }.get(name)
+
```
adiroiban, line 380:
```suggestion
class InMemoryInventory:
    """
    A simple inventory implementation which has an in memory API link mapping
    """
    INVENTORY = {
            'socket.socket': 'https://docs.python.org/3/library/socket.html#socket.socket',
            }
    
    def getLink(self, name: str) -> Optional[str]:
        return self.INVENTORY.get(name)

```

adiroiban (approved):
Looks good. Thanks.

I guess that the README.rst is updated with the release notes on the release branch/PR.

mthuurne:
> I guess that the README.rst is updated with the release notes on the release branch/PR.

Yes, that's all still manual work. Until 20.7.x, pydoctor didn't have any release notes, so that's still a step up ;)

PR #239, opened 2020-10-28
Fix file names for function pages

mthuurne:
This fixes a regression that slipped in as part of #227. See the commit comment for details.

### DIFF
--- a/pydoctor/templatewriter/writer.py
+++ b/pydoctor/templatewriter/writer.py
@@ -5,7 +5,8 @@ import shutil
 
 from pydoctor import model
 from pydoctor.templatewriter import DOCTYPE, pages, summary
-from pydoctor.templatewriter.util import link, templatefile
+from pydoctor.templatewriter.util import templatefile
+from twisted.python.filepath import FilePath
 from twisted.web.template import flattenString
 
 
@@ -63,9 +64,9 @@ class TemplateWriter:
             if self.dry_run:
                 self.total_pages += 1
             else:
-                f = open(os.path.join(self.base, link(ob)), 'wb')
-                self.writeDocsForOne(ob, f)
-                f.close()
+                path = FilePath(self.base).child(f'{ob.fullName()}.html')
+                with path.open('wb') as out:
+                    self.writeDocsForOne(ob, out)
         for o in ob.contents.values():
             self.writeDocsFor(o, functionpages)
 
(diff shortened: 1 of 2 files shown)


pydoctor/templatewriter/writer.py
```diff
@@ -63,9 +64,9 @@ def writeDocsFor(self, ob, functionpages):
             if self.dry_run:
                 self.total_pages += 1
             else:
-                f = open(os.path.join(self.base, link(ob)), 'wb')
-                self.writeDocsForOne(ob, f)
-                f.close()
+                path = FilePath(self.base).child(f'{ob.fullName()}.html')
+                with path.open('wb') as f:
+                    self.writeDocsForOne(ob, f)
```
adiroiban, line 69:
```suggestion
                with path.open('wb') as stream:
                    self.writeDocsForOne(ob, stream)
```

I am new to py3 and I was a bit confused by the previous `f'{ob.fullName()}.html` 

I like the `f` literal ...but I think that is a 3.6 thing...and in twisted 3.5 is still supported ... but I am happy to remove support for 3.5 :)

adiroiban (approved):
It looks good. Thanks.

One medium comment is that I don't see the test for this change .

if writing a test is complicated, I think that we can skip that and merge it as it is.

Thanks!

pydoctor/templatewriter/writer.py
mthuurne: I'll use `out`: that's the name I typically use in new code, but here the old code was using `f` already so that stuck.

> I am new to py3 and I was a bit confused by the previous `f'{ob.fullName()}.html`
> 
> I like the `f` literal ...but I think that is a 3.6 thing...and in twisted 3.5 is still supported ... but I am happy to remove support for 3.5 :)

Yes, it's a 3.6 feature, which is fine for pydoctor but not yet for Twisted. I think the plan was to drop 3.5 in Twisted after the next release.

mthuurne:
> One medium comment is that I don't see the test for this change .
> 
> if writing a test is complicated, I think that we can skip that and merge it as it is.

There is a test that checks the presence of output files: `test_basic_package()`, but it doesn't enable function pages. However, I consider function pages a bit of a misfeature, so once Twisted no longer needs them, I want to remove that option. Therefore I'm not motivated to write new tests for it.

PR #241, opened 2020-10-29
[Fixes #240] Add support for source href with Windows paths.

### LINKED ISSUE #240: Use forward slash when creating the source href on Windows.
In the following code source_path can be a windows path like `src\some\path` and then the url wil be `http://base.com/url\src\some\path'

https://github.com/twisted/pydoctor/blob/678441a505b4ee8700b29dfd6c264e1d21e294a8/pydoctor/model.py#L572

As a fix maybe alwasys run source_path = source_path.replace('\\', '/') when creating the HREF

Once this is fixed a new release can be made and enable the Twisted release scripts on Windows 

This was detected here https://github.com/twisted/twisted/pull/1473

adiroiban:
This updates the source HREF generation to generate valid HREF when the source path is on windows.

As a drive by, it updates the usage of imp lib with importlib.... as imp is deprecated.

### DIFF
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -9,7 +9,7 @@ being documented -- a System is a bad of Documentables, in some sense.
 import ast
 import builtins
 import datetime
-import imp
+import importlib
 import inspect
 import os
 import platform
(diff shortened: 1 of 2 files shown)


pydoctor/model.py
```diff
@@ -656,22 +656,21 @@ def addPackage(self, dirpath, parentPackage=None):
...
                     continue
                 if package is not None:
                     module_full_name = f'{package.fullName()}.{module_name}'
                 else:
                     module_full_name = module_name
-                py_mod = imp.load_module(
-                    module_full_name, open(path, 'rb'), path,
-                    (suffix, mode, impl))
+                py_mod = importlib.util.spec_from_file_location(
+                    module_full_name, path)
```
adiroiban, line 671: The problem is that we don't have automated tests for this branch

mthuurne:
I'm taking a different approach in #242: by using `pathlib.Path`, we don't have to deal with the difference in separators explicitly.

The replacement of `imp` by `importlib` is still interesting though.

adiroiban:
Good :) I will close this PR. Feel free to reuse the unit tests and the importlib change.

The importlib should be in a dedicated commit.

Thanks!

mthuurne:
I'll extract the `importlib` changes once #242 is done, to make sure it plays well with the added type annotations.

pydoctor/model.py
mthuurne: To test the full flow would require using a C extension, which is a pain to set up. However, we could move the import handling into `introspectModule()`, which can also be used on pure Python and is in fact unit tested that way. Then adding type annotations will let mypy verify the bit of code that remains without unit test coverage.

PR #242, opened 2020-10-29
Run unit tests on Windows

### LINKED ISSUE #240: Use forward slash when creating the source href on Windows.

mthuurne:
This modifies the `unit` workflow to include a test on Windows. It also fixes the tests and the code to make the tests pass on Windows.

The use of `pathlib.Path.as_posix()` fixes #240.

### DIFF
--- a/.github/workflows/static.yaml
+++ b/.github/workflows/static.yaml
@@ -26,8 +26,8 @@ jobs:
       run: |
         test -r /etc/os-release && sh -c '. /etc/os-release && echo "OS: $PRETTY_NAME"'
         python --version
-        python -c 'print("\nENVIRONMENT VARIABLES\n=====================\n")'
-        python -c 'import os; [print(f"{k}={v}") for k, v in os.environ.items()]'
+        python -c "print('\nENVIRONMENT VARIABLES\n=====================\n')"
+        python -c "import os; [print(f'{k}={v}') for k, v in os.environ.items()]"
 
     - name: Run mypy
       run: |
--- a/.github/workflows/system.yaml
+++ b/.github/workflows/system.yaml
@@ -26,8 +26,8 @@ jobs:
       run: |
         test -r /etc/os-release && sh -c '. /etc/os-release && echo "OS: $PRETTY_NAME"'
         python --version
-        python -c 'print("\nENVIRONMENT VARIABLES\n=====================\n")'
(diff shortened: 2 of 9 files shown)


adiroiban (approved):
This looks good. Thanks!

I would like to also have the unit test that I have created in the previous PR for Windows path.

With the updated tests, I find it hard to read where the Windows path separator is tests.

Also, with the unit test from my branch I was able to test that Windows path are still handled even if I was running the tests on Linux...so is much easier to debug and detect regression.

But feel free to merge it as it is...as long as the issue is fixed, I am fine with that.

Thanks!

mthuurne:
> I would like to also have the unit test that I have created in the previous PR for Windows path.

I couldn't use `test_setSourceHrefOptionWindows()` as-is, since it's not compatible with the `pathlib.Path` approach: while `pathlib.WindowsPath` exists under Linux, it can't be instantiated there. And `PosixPath` does not treat backslashes as separators.

However, I did refine the existing `test_setSourceHrefOption()` test to be more representative of how actual execution will happen on Windows.

> With the updated tests, I find it hard to read where the Windows path separator is tests.

The goal of using `Path` was to move the separators out of the pydoctor code, so I guess that qualifies as a success ;)

> Also, with the unit test from my branch I was able to test that Windows path are still handled even if I was running the tests on Linux...so is much easier to debug and detect regression.

There is something to be said for allowing `WindowsPath` to be instantiated on Linux for testing purposes. But that would require a change in the standard library.

As for strings vs Paths, I think there is a less chance of bugs with Paths, especially when combined with mypy. So while debugging might be harder, it won't be needed as often.

mthuurne:
I just read this in the `pathlib` docs:

> You cannot instantiate a WindowsPath when running on Unix, but you can instantiate PureWindowsPath.

So it might be possible to test Windows paths on Linux after all. I don't think pure paths are a limitation for the `setSourceHref` test.

PR #243, opened 2020-10-29
Replace use of deprecated 'imp' module by 'importlib'

mthuurne:
This was extracted from a drive-by fix in #241.

### DIFF
--- a/pydoctor/test/test_model.py
+++ b/pydoctor/test/test_model.py
@@ -165,8 +165,8 @@ class Dummy:
     def crash(self) -> None:
         """Mmm"""
 
-def test_introspection() -> None:
-    """Find docstrings from this test using introspection."""
+def test_introspection_python() -> None:
+    """Find docstrings from this test using introspection on pure Python."""
     system = model.System()
     system.introspectModule(Path(__file__), __name__)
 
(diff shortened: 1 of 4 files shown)


mthuurne:
@adiroiban: Your idea to use `cython-test-exception-raiser` to test introspection of extensions works. I did have to add a workaround to make it pass on pypy3 as well.

pydoctor/test/test_model.py
```diff
@@ -166,19 +165,45 @@ class Dummy:
...
     assert method.docstring == "Mmm"
+
+
+def test_introspection_extension() -> None:
+    """Find docstrings from this test using introspection of an extension."""
+
+    try:
+        import cython_test_exception_raiser.raiser
+    except ImportError:
+        pytest.skip("cython_test_exception_raiser not installed")
```
adiroiban, line 191: I guess that we can assume that the extension is always installed and don't handle the import error.

We have wheels for major OSes so it should be easy to install it.

adiroiban (approved):
Wow! Excellent work.

Happy to see it merged.

Thanks

pydoctor/test/test_model.py
mthuurne: I thought pypy might need it, but it seems that the extension is available in pypy too.

Still, I think it's good to be able to run the tests on any platform, even if the extension is not available or the user hasn't installed it.

adiroiban: my concern is to not end up with accidentally not installing the extension and then the test is always skipped... like we had with twisted tests and pydoctor not being installed.

To mitigate this, I am using codecov and a separate target that makes sure the tests code have 100%  coverage... if someone some tests are always skipped then the coverage will no longer be 100% and end up with a red flag :)

mthuurne: That's a good idea. I want to work on other things right now though, so I filed #245 to make sure we remember it.

PR #244, opened 2020-10-29
Windows paths follow-up

mthuurne:
After #242, it is possible for Hypothesis to generate directory names that are not valid in Windows, which leads to random failures in `test_prepareCache()`. This adds a few `assume()` checks to filter out invalid directory names.

Additionally, `test_setSourceHrefOption()` was modified to use pure paths, which allows testing Windows paths on Linux and vice versa.

### DIFF
--- a/pydoctor/driver.py
+++ b/pydoctor/driver.py
@@ -1,5 +1,7 @@
 """The command-line parsing and entry point."""
 
+from optparse import Option, OptionParser, OptionValueError
+from pathlib import Path
 import datetime
 import os
 import sys
@@ -32,9 +34,18 @@ def findClassFromDottedName(dottedname, optionname):
 
 MAKE_HTML_DEFAULT = object()
 
+def check_path(option, opt, value):
+    try:
+        return Path(value)
+    except Exception as ex:
+        raise OptionValueError(f"{opt}: invalid path: {ex}")
+
+class CustomOption(Option):
+    TYPES = Option.TYPES + ("path",)
+    TYPE_CHECKER = dict(Option.TYPE_CHECKER, path=check_path)
+
 def getparser():
-    from optparse import OptionParser
-    parser = OptionParser(version=__version__.public())
+    parser = OptionParser(option_class=CustomOption, version=__version__.public())
     parser.add_option(
         '-c', '--config', dest='configfile',
         help=("Use config from this file (any command line"
@@ -49,10 +60,9 @@ def getparser():
         '--project-url', dest='projecturl',
         help=("The project url, appears in the html if given."))
     parser.add_option(
-        '--project-base-dir', dest='projectbasedirectory',
-        help=("Absolute path to the base directory of the "
-              "project.  Source links will be computed based "
-              "on this value."))
+        '--project-base-dir', dest='projectbasedirectory', type='path',
+        help=("Path to the base directory of the project.  Source links "
+              "will be computed based on this value."))
     parser.add_option(
         '--testing', dest='testing', action='store_true',
         help=("Don't complain if the run doesn't have any effects."))
--- a/pydoctor/driver.py
+++ b/pydoctor/driver.py
@@ -34,7 +34,10 @@ def findClassFromDottedName(dottedname, optionname):
 
 MAKE_HTML_DEFAULT = object()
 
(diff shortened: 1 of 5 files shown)


pydoctor/driver.py
```diff
@@ -32,9 +34,18 @@ def findClassFromDottedName(dottedname, optionname):
 
 MAKE_HTML_DEFAULT = object()
 
+def check_path(option, opt, value):
+    try:
```
adiroiban, line 38: it would be nice to have a docstring explining what is this checking .

the check_path is a generic name in the context of pydoctor.driver

pydoctor/model.py
```diff
@@ -573,7 +573,7 @@ def setSourceHref(self, mod: _ModuleT, source_path: Path) -> None:
         if self.sourcebase is None:
             mod.sourceHref = None
         else:
-            projBaseDir = Path(mod.system.options.projectbasedirectory)
+            projBaseDir = mod.system.options.projectbasedirectory
```
adiroiban, line 576: instead of validating the value at the high level config level, I would preferred to validate it at this lower level.

In the case in which we will pass a different options implementation, the value will still be validated.

At the same time, in this way we can keep the configuration simple with just a standard string value.

I am ok with having `projBaseDir` as a Path instance.
Is just that I prefer that the configuration not to be a path instance, but just a string instance.

pydoctor/test/test_model.py
```diff
@@ -33,17 +34,16 @@ class FakeDocumentable:
 
 
 
-def test_setSourceHrefOption() -> None:
+@pytest.mark.parametrize('projectBaseDir', [
+    PurePosixPath("/foo/bar/ProjectName"),
+    PureWindowsPath("C:\\foo\\bar\\ProjectName")]
```
adiroiban, line 39: I see what the config was converted into a Path instance... and I guess is ok.

pydoctor/test/test_sphinx.py
```diff
@@ -636,6 +636,14 @@ def test_prepareCache(
     L{cachecontrol.CacheControlAdapter} for C{http} and C{https} URLs.
     """
 
+    # Windows doesn't like paths ending in a space or dot.
+    assume(cacheDirectoryName[-1] not in '. ')
+
+    # These DOS device names still have special meaning in modern Windows.
+    assume(cacheDirectoryName.upper() not in {'CON', 'PRN', 'AUX', 'NUL'})
+    assume(not cacheDirectoryName.upper().startswith('COM'))
+    assume(not cacheDirectoryName.upper().startswith('LPT'))
```
adiroiban, line 645: I guess that is ok...but I think this is not the right way to fix it.

Instead of st.text() I think that we should have an st.path or some helper which will always return a valid path.

adiroiban (approved):
Thanks for the followup.

This looks good and can be merged.

I see why we ended with the Path object at the configuration level.

From my experience, is best to keep the configuration level as dumb as possible and do the validation at the lower level possible... but I understand that for pydoctor this should not be a problem.

pydoctor/driver.py
mthuurne: I'll add one.

pydoctor/model.py
mthuurne: We don't actually validate the path at all; the name `check_path` is a bit misleading. I wasn't even able to trigger the exception handling without modifying the code to raise something, but I figured that if there is a string that `Path()` won't accept, we should report it without crashing.

If we would validate the path, I agree that it would be best to not do that in the config parser. However, I also think `setSourceHref()` is too deep and called too late. A better place could be `System.process()`.

I do like having `Path` instances in the config object though. I think it's similar to storing numeric values as `int`.

pydoctor/test/test_sphinx.py
mthuurne: I looked for a file name generating strategy in Hypothesis' docs, but couldn't find any. One could be written of course, but I think that's overkill for the one test case in pydoctor that uses this strategy.

The Hypothesis docs do suggest that as long as `assume()` rejects only a small portion of the generated values, it's a valid approach.

adiroiban: Yep. Thanks.

PR #246, opened 2020-10-30
Fix system test

mthuurne:
Make sure our system test actually turns red when pydoctor crashes.

The system test broke in #242, but we didn't notice because the error wasn't shown at the top level.

### DIFF
--- a/.github/workflows/system.yaml
+++ b/.github/workflows/system.yaml
@@ -30,6 +30,5 @@ jobs:
         python -c "import os; [print(f'{k}={v}') for k, v in os.environ.items()]"
 
     - name: Generate Twisted API docs
-      continue-on-error: true
       run: |
         tox -e twisted-apidoc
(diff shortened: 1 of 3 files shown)


adiroiban (approved):
Thanks. 

I think that is important to have this fix now.

Note that I plan to update the twisted admin script to exit with -1 on error .

Once that lands in Twisted the tox config pydoctor needs to be updated.

But this should land now...and I don't know how long it would take.

Or maybe I / you can create a separate PR / ticket in twisted and land the script change in twisted 

https://github.com/twisted/twisted/pull/1466/files#diff-80440324a82eacdeebc2fa7d4a5ca54c02d02a4d481fcfc5714f5da7738692a6

mthuurne:
> Note that I plan to update the twisted admin script to exit with -1 on error .

Aren't exit codes typically small but non-negative numbers?

It seems that uncaught exceptions lead to an exit code of 1. It is a convention to return 2 on invalid command line options. So perhaps 3 can be used to indicate content problems?

adiroiban:
true... sorry. 1 for error exit code.

PR #250, opened 2020-11-01
Improve parsing of intersphinx inventories

mthuurne:
I noticed that `L{dict}` was being linked to the 2to3 documentation, rather than the built-in type. This is due to there being 3 pages with the name `dict` and our inventory parser remembers the last one it sees, even if that is not the most relevant one.

While investigating that problem, I also found that `L{key}` in Twisted's documentation was not flagged as an error despite there not being any function or constant named `key` in the Python standard library. It turns out the name `key function` was split in two by the parser. (To be fair to our parser, using spaces as separators while also allowing spaces in fields is a very unfortunate choice for a format.)

### DIFF
--- a/pydoctor/sphinx.py
+++ b/pydoctor/sphinx.py
@@ -117,7 +117,8 @@ class SphinxInventory:
                     'Failed to parse line "%s" for %s' % (line, base_url),
                     )
                 continue
-            result[name] = (base_url, location)
+            if typ.startswith('py:'):
+                result[name] = (base_url, location)
         return result
 
     def getLink(self, name: str) -> Optional[str]:
--- a/pydoctor/sphinx.py
+++ b/pydoctor/sphinx.py
@@ -117,8 +117,12 @@ class SphinxInventory:
                     'Failed to parse line "%s" for %s' % (line, base_url),
                     )
                 continue
-            if typ.startswith('py:'):
-                result[name] = (base_url, location)
+
+            if not typ.startswith('py:'):
+                # Non-Python references are ignored.
+                continue
+
+            result[name] = (base_url, location)
         return result
 
     def getLink(self, name: str) -> Optional[str]:
--- a/pydoctor/sphinx.py
+++ b/pydoctor/sphinx.py
@@ -109,14 +109,15 @@ class SphinxInventory:
         """
         result = {}
         for line in payload.splitlines():
-            parts = line.split(' ', 4)
-            if len(parts) != 5:
+            try:
+                name, typ, prio, location, display = _parseInventoryLine(line)
+            except ValueError:
                 self.error(
                     'sphinx',
                     'Failed to parse line "%s" for %s' % (line, base_url),
                     )
                 continue
-            result[parts[0]] = (base_url, parts[3])
+            result[name] = (base_url, location)
         return result
 
     def getLink(self, name: str) -> Optional[str]:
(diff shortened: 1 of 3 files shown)


pydoctor/sphinx.py
```diff
@@ -134,6 +136,38 @@ def getLink(self, name: str) -> Optional[str]:
         return f'{base_url}/{relative_link}'
 
 
+def _parseInventoryLine(line: str) -> Tuple[str, str, int, str, str]:
```
adiroiban, line 139:
```suggestion
def _parseSphinxInventoryLine(line: str) -> Tuple[str, str, int, str, str]:
```

pydoctor/test/test_sphinx.py
```diff
@@ -402,6 +402,33 @@ def test_parseInventory_single_line(inv_reader_nolog: sphinx.SphinxInventory) ->
     assert {'some.attr': ('http://base.tld', 'some.html')} == result
 
 
+def test_parseInventory_spaces() -> None:
+    """
+    Check that columns that contain spaces are not split up.
```
adiroiban, line 407:
```suggestion
    When parsing an Sphinx inventory line, it will always return 5 values, even when some inventory column values contains spaces.
```

pydoctor/sphinx.py
```diff
@@ -109,14 +109,16 @@ def _parseInventory(
...
+                name, typ, prio, location, display = _parseInventoryLine(line)
+            except ValueError:
                 self.error(
                     'sphinx',
                     'Failed to parse line "%s" for %s' % (line, base_url),
                     )
                 continue
-            result[parts[0]] = (base_url, parts[3])
+            if typ.startswith('py:'):
+                result[name] = (base_url, location)
```
adiroiban, line 121:
```suggestion
            if not typ.startswith('py:'):
                # Non python references are ignored.
                continue
            result[name] = (base_url, location)
```

adiroiban (approved):
Great work. Thanks. Only minor comments.

it's ugly that we can have spaces in the first column...  but this is life :)

mthuurne:
> it's ugly that we can have spaces in the first column... but this is life :)

I guess they didn't realize there could be spaces in the first column until the format was already in use. And then wanted to keep version 2 of the protocol backwards compatible with the version 1 parser, so they couldn't change the separator. That's the only explanation I can think of.

pydoctor/sphinx.py
mthuurne: Since this is a private function in a module named `pydoctor.sphinx`, I think it's clear enough from the context that it's a Sphinx inventory we're parsing.

pydoctor/test/test_sphinx.py
mthuurne: I'll update the docstring.

pydoctor/sphinx.py
mthuurne: We already have a `continue` statement in the method, so I guess this is consistent.

adiroiban: this is a minor comment.

is a question of  style - https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

the idea is that the happy path handling is non indented and not "hidden" in a conditional branch.

mthuurne: I'm not sure this should be considered a happy path, in the sense that there is nothing wrong with non-Python references, but we're currently just not interested in them.

Issue #163, opened 2019-11-17
Move sources into "src" directory


mthuurne:
This reduces the risk of accidentally testing against the source files when we want to test packaged code instead. Twisted made [a similar move](https://twistedmatrix.com/trac/ticket/8214).

PR #253, opened 2020-11-04
Add pydoctor help directive.

adiroiban:
Scope
=====

It adds a sphinx directive for producing the help output.

it also updates the tox docs env to fail on Sphinx warnings and the run is integrated into the Static CI build. 

With this change, we no longer have Sphinx warnings... all clean :)

There are no unit tests for the directive... but this should be private.

How to test
=========

RTD page https://pydoctor--253.org.readthedocs.build/en/253/help.html  ... you might see an old cache.

Local build
```
tox -e docs
firefox build/docs/help.html
```


adiroiban:
@tristanlatr  you might want to take a look and do a review :)

I think that this is the right way (tm) to inject pydoctor help output on the fly.

It now uses the default block literal syntax highlight ...maybe I should remove that...but I wanted to keep the code simple for the first review.

Thanks

docs/source/conf.py
```diff
@@ -43,7 +45,7 @@
 
 # Configure intersphinx magic
 intersphinx_mapping = {
-    'pydoctor': ('https://pydoctor.readthedocs.io/api/', None),
+    #'pydoctor': ('https://pydoctor.readthedocs.io/api/', None),
```
adiroiban, line 48: objects.inv is not yet created.... I plan to add it in a future PR

docs/source/conf.py
```diff
@@ -30,7 +30,9 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    "sphinx_rtd_theme", "sphinx.ext.intersphinx"
+    "sphinx_rtd_theme",
+    "sphinx.ext.intersphinx",
+    "pydoctor.sphinx_ext.help_output",
```
adiroiban, line 35: Feel free to suggest a better name.

Note that we already have pydoctor.sphinx which is the code to generate the Sphinx inventory.

pydoctor/driver.py
```diff
@@ -365,6 +365,10 @@ def p(msg):
                 for fn in sorted(system.docstring_syntax_errors):
                     p('    '+fn)
 
+            if system.warnings:
+                # Update exit code if the run has produced warnings.
+                exitcode = 1
```
tristanlatr, line 370: This will break some workflows. Is it in the scope of the PR? 
Anyway it's definitely something to document in the next version changelog!

adiroiban: Yes... it will break stuff.

But I think that returning 0 on warnings is a bug :)

I can revert it.... but I wanted to have a simple and useful tox environment that will fail on warnings :)

mthuurne: Please use an exit code other than 1, since 1 is returned on uncaught exceptions. I think 3 could be a good choice, so we can reserve 2 for dealing with invalid command line arguments.

Compilers typically return 0 on warnings and only return non-zero when they weren't able to generate code. I think it's useful to have an option (`-Werror` equivalent) to return a non-zero exit code when warnings are reported, but it shouldn't be the default.

When building embedded Linux systems, I have been really annoyed with projects that use `-Werror` in their default release configuration, since it breaks their builds if you use more recent (and therefore often stricter) tooling. We shouldn't inflict that upon anyone if we can avoid it.

pydoctor/sphinx_ext/help_output.py
```diff
@@ -0,0 +1,38 @@
...
+from pydoctor.driver import parse_args
+
+
+class HelpOutputDirective(Directive):
+    """
+    Directive that will generate the pydoctor help as block literal.
+    """
+    has_content = True
+
+    def run(self):
```
mthuurne, line 18: Please add type annotations on new code. They do help a lot finding bugs and making maintenance safer.

tox.ini
```diff
@@ -68,3 +68,31 @@ commands =
...
         {tty:--pretty:}                        \
         {posargs:pydoctor}
+
+
+[testenv:apidocs]
+description = Build only the API documentation
+
+whitelist_externals =
+    cat
+    test
```
mthuurne, line 78: Do we use `cat` and `test`?

setup.cfg
```diff
@@ -31,6 +31,7 @@ packages = find:
 setup_requires = incremental
 python_requires = >=3.6
 install_requires =
+    docutils
```
mthuurne, line 34: Users that only need epytext support and don't build the docs don't need `docutils` installed. So I think it would be better to have it as an extra.

tox.ini
```diff
@@ -68,3 +68,31 @@ commands =
...
+commands =
+    pydoctor \
+    --add-package=pydoctor \
+    --project-name=pydoctor \
+    --project-url=https://github.com/twisted/pydoctor/ \
+    --html-viewsource-base=https://github.com/twisted/pydoctor/tree/master \
+    --html-output=build/apidocs \
+    --project-base-dir="{toxinidir}" \
```
mthuurne, line 86: Perhaps it is a good moment to move our source code to a `src` directory (#163), now that there are few open PRs and before making more changes to our build.

tox.ini
```diff
@@ -68,3 +68,31 @@ commands =
...
+
+whitelist_externals =
+    cat
+    test
+commands =
+    pydoctor \
+    --add-package=pydoctor \
+    --project-name=pydoctor \
+    --project-url=https://github.com/twisted/pydoctor/ \
+    --html-viewsource-base=https://github.com/twisted/pydoctor/tree/master \
```
mthuurne, line 84: Source links that point to a branch, in particular the master branch, will get outdated. We should use a tag or commit ID instead.

pydoctor/sphinx_ext/help_output.py
```diff
@@ -0,0 +1,38 @@
...
+
+        text = ['pydoctor --help'] + stream.getvalue().splitlines()[1:]
+        return [nodes.literal_block(text='\n'.join(text))]
+
+
+def setup(app):
+    app.add_directive('help_output', HelpOutputDirective)
+
+    return {
+            'version': '0.1',
```
mthuurne, line 35: Maybe use the pydoctor version instead?

pydoctor/driver.py
adiroiban: > Compilers typically return 0 on warnings and only return non-zero when they weren't able to generate code. I think it's useful to have an option (-Werror equivalent) to return a non-zero exit code when warnings are reported, but it shouldn't be the default.

True. I will revert it.

And I think that something like  `-Werror` is better.

After another review it looks like this branch is rather experimental ... with many changes :+1: 
So it might be better to break it into smaller PRs.

I wanted to get the dynamic help ouput...but I was missing the basic dev tools... so I ended up with this PR . Sorry.

mthuurne: No problem, it happens to me all the time that when working on one thing, I run into other issues that turn out to be complex enough to become PRs of their own. I still don't have the PR that warns about documented non-existing function parameters in a state where I can submit it...

adiroiban: I will move this to https://github.com/twisted/pydoctor/pull/256

tox.ini
adiroiban: This was moved to https://github.com/twisted/pydoctor/pull/256

This is an tox env for manual testing...so I went with just master as the default link.

This tox env is not designed to be used for production / public site

tox.ini
adiroiban: I can do it when there are not many PR.

I have a few PR that I would like to merge and then we can move to src.

For pydoctor, I don't see as a critical task... I assume that people are using pydoctor for development and for non-critical environments.

I am full supported of automated tests for anything that is production and customer facing/interaction ... but for development tools I prefer a  more Carpe Diem / YOLO approach. 

But we can move to src as it shouldn't be an issue.

pydoctor/sphinx_ext/help_output.py
adiroiban: This is only to be use for internal / private pydoctor documentation... I would keep it simple and just use a static version.
I have no idea how this version is supposed to be used.

pydoctor/sphinx_ext/help_output.py
```diff
@@ -0,0 +1,53 @@
...
+
+        stream = StringIO()
+        try:
+            with redirect_stdout(stream):
+                parse_args(['--help'])
+        except SystemExit:
+            # The stdlib --help handling triggers system exit.
+            pass
+
+        text = ['pydoctor --help'] + stream.getvalue().splitlines()[1:]
```
adiroiban, line 39: Let me know if you want unit tests for this.

tox.ini
mthuurne: It's not so much for the users but for ourselves: it's easy to accidentally test the source tree when you think you're testing in a tox env, which can lead to incorrect conclusions and wasting a lot of time.

Another reason is a bit more selfish: since many projects are now using `src`, if I'm searching in the source I'll type `grep -r something src` automatically and then have to correct the directory name.

mthuurne: It doesn't need to happen right now, but the sooner we do it, the fewer non-source files we need to update, so it would be nice to do it before releasing.

pydoctor/sphinx_ext/help_output.py
mthuurne: If it shows up in the logging, having an actual version there might be useful. But if it's entirely invisible, it doesn't matter. Will Sphinx accept it if you remove this item altogether?

mthuurne:
Do we want to ship this as part of the wheel? (`bdist`)

It should be part of the source tarball (`sdist`), since Linux distributions like to build software from an official source archive rather than from Git. But when installing pydoctor with `pip`, people are going to use the online docs rather than generate a set of offline docs on their machine.

The same question applies to the tests: I know Twisted deliberately keeps the tests inside the same source tree, but I have some doubts about how often that is useful. And unlike Twisted, we use pytest instead of shipping our own test tool, so people who want to run the pydoctor tests will need to take additional steps anyway, at which point they might as well just clone our Git.

One potential problem with mixing production code and tests is that if we add code in `pydoctor/test/testpackages/` that is either invalid Python or only valid in particular Python versions, we might break packaging systems that try to pre-compile installed modules to bytecode.

pydoctor/sphinx_ext/help_output.py
mthuurne: Maybe a system test is more appropriate: run the extension under Sphinx, check the exit code and check the presence of a few options that are unlikely to ever disappear in the output.

In any case, I don't think it needs detailed functional tests, but it would be good to execute it at least once as part of a CI cycle.

mthuurne (approved):
With some kind of sanity check test added, I think it's good to merge.

pydoctor/sphinx_ext/help_output.py
adiroiban: It accepts without a version

The code is here https://github.com/sphinx-doc/sphinx/blob/4c1b8b8db6fa25afcab1a6a3634776009cfafddf/sphinx/extension.py#L25

The version is used together with `needs_extensions`

I will use  the same version as pydoctor.

pydoctor/sphinx_ext/help_output.py
adiroiban: OK. I am building the whole sphinx docs and then run a grep on the help.html page.

adiroiban:
> Do we want to ship this as part of the wheel? (`bdist`)

There is no requirement for this code to be part of the wheel.
This is an internal / private extension to help with Sphinx docs.

The idea is that I plan to add a public extension for Sphinx as part of https://github.com/twisted/pydoctor/pull/254

It will be an public extension for Sphinx to make it easy to generate pydoctor as part of Sphinx build system.... mostly to help with Read The docs build.

Since we will have to include at least one public extension extension in the wheel, I decided to keep it simple and also include this extension

But I have renamed it to an underscore module name to highlight that it is private.

adiroiban:
Thanks @mthuurne for the review.

I think that I have included all your feedback. Let me know if something is not OK and I will send a followup PR.

I think that this is an overall improvement over master.

PR #254, opened 2020-11-04
Build pydoctor apidocs together with Sphinx.

adiroiban:
Scope
=====

This add a Sphinx extension that will trigger pydoctor apidocs build together with sphinx build.

Changes
=======

Add a pydoctor extension that is called at the end of the build.

pydoctor warnings are converted into sphinx warnings.

Follow up work
===========

1. Once this is merge, we can have a public pydoctor sphinx inventory object hosted by Read The Docs and use intersphinx to link from Sphinx to pydoctor  API (See also 6)

2. We need to check how this behaves on RTD tags / stable / latest builds.

3. We should also check RTD stabe vs latest and decide which is which.
* https://pydoctor.readthedocs.io/en/latest/
* https://pydoctor.readthedocs.io/en/stable/

4. Fix all pydoctor warnings from inside the pydoctor own code #269

5. Create some "pydoctor RTD" theme that will provide better integrating with the RTD theme.

6 Build pydoctor intersphinx object before calling RTD sphinx build so that we can reference new pydoctor API included in the branch.

How to test
=========

```
tox -e docs
firefox build/docs/index.html
```

On RTD you can use https://pydoctor--254.org.readthedocs.build/en/254/

Intersphinx generated at https://pydoctor--254.org.readthedocs.build/en/254/api/objects.inv

Check the magic inside the "Quicklinks" TOC.


adiroiban:
@mthuurne can you please take a look at this PR.

I think that once this is merged it will make it easier to review UI / CSS  / HTML PRs like https://github.com/twisted/pydoctor/pull/265

Thanks!

docs/source/conf.py
```diff
@@ -33,6 +36,7 @@
     "sphinx_rtd_theme",
     "sphinx.ext.intersphinx",
     "pydoctor.sphinx_ext._help_output",
+    "pydoctor.sphinx_ext.api_output",
```
tristanlatr, line 39: Thanks a lot, this is a game changer for Sphinx users to use PyDoctor!

May be the Sphinx plugin could be named "pydoctor.sphinx_ext.build_docs" or  "pydoctor.sphinx_ext.build_api_docs"?

Just to emphasis that the goal of the thing is to build API references as part of the Sphinx build.

adiroiban: maybe `build_apidocs` ?

I agree that `api_output` is a stupid name. Thanks for the feedback.

tristanlatr (approved):
> So, if you have time, please send your feedback / review for PR #254 so that we can have it accepted.

This is all great stuff to me. 
`pydoctor.sphinx_ext.build_apidocs` would be a perfect name for such plugin.

An addition to the Follow up work:
- document the process in `docs/source/usage.rst` section "With Sphinx and Read The Docs" part which is commented for now.

docs/source/conf.py
```diff
@@ -10,10 +10,13 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
```
mthuurne, line 13: Removing that line makes the example incomplete. In my opinion the example should either be removed entirely or left as-is.

docs/source/conf.py
```diff
@@ -59,3 +63,39 @@
...
+    _git_reference = subprocess.getoutput('git rev-parse HEAD')
+
+if os.environ.get('READTHEDOCS', '') == 'True':
+    rtd_version = os.environ.get('READTHEDOCS_VERSION', '')
+    if '.' in rtd_version:
+        # It looks like we have a tag build.
+        _git_reference = rtd_version
+
+
+pydoctor_cwd = str(pathlib.Path(__file__).parent.parent.parent)
```
mthuurne, line 89: I'd prefer a name like `pydoctor_root` or `pydoctor_top`. The term `cwd` has the specific meaning of the current OS working dir, which is not what we're dealing with here. (Later I saw that we actually do make it the working dir, but it would still be better if that's an implementation decision that doesn't spill over into this configuration file.)

You don't need to convert to `str` here: in case we might want to use this directory for referencing other files later, it is more convenient to have it as a `Path` object.

docs/source/conf.py
```diff
@@ -59,3 +63,39 @@
...
+    '--quiet',
+    '--add-package=pydoctor',
+    '--project-name=pydoctor',
+    '--project-url=https://github.com/twisted/pydoctor/',
+    '--docformat=epytext',
+    '--intersphinx=https://docs.python.org/3/objects.inv',
+    '--make-html',
+    '--html-viewsource-base=https://github.com/twisted/pydoctor/tree/' + _git_reference,
+    '--html-output={outdir}/api',
+    '--project-base-dir=' + pydoctor_cwd,
```
mthuurne, line 100:
```suggestion
    f'--project-base-dir={pydoctor_root}',
```
The f-string automatically converts the `Path` (or any other value) to `str`.

docs/source/conf.py
```diff
@@ -59,3 +63,39 @@
...
+
+if os.environ.get('READTHEDOCS', '') == 'True':
+    rtd_version = os.environ.get('READTHEDOCS_VERSION', '')
+    if '.' in rtd_version:
+        # It looks like we have a tag build.
+        _git_reference = rtd_version
+
+
+pydoctor_cwd = str(pathlib.Path(__file__).parent.parent.parent)
+pydoctor_args = [
```
mthuurne, line 90: Should we be escaping the values inserted here? Some of them come from user inputs, after all. I don't know how `optparse` deals with these arguments if they contain for example spaces.

docs/source/index.rst
```diff
@@ -22,10 +22,11 @@ Installation
...
 .. toctree::
    :maxdepth: 1
    :caption: Quick links
-   
+
+   api/index
    GitHub <https://github.com/twisted/pydoctor>
-   PyPi <https://pypi.org/project/pydoctor>
\ No newline at end of file
+   PyPi <https://pypi.org/project/pydoctor>
```
mthuurne, line 32:
```suggestion
   PyPI <https://pypi.org/project/pydoctor>
```
Capital "i" for "Index" (verified on pypi.org).

pydoctor/sphinx_ext/api_output.py
```diff
@@ -0,0 +1,69 @@
...
+as otherwise any extra output is converted into Sphinx warnings.
+"""
+import os
+from sphinx.util import logging
+from sphinx.errors import ConfigError
+
+from contextlib import redirect_stdout
+from io import StringIO
+from pydoctor import __version__
+from pydoctor.driver import main
```
mthuurne, line 19: Please group imports in PEP8 style: standard lib, other libs, internal. I want to adopt `isort` at some point to get it exactly right everywhere, but in the mean time having at least the grouping helps making the code easier to read.

pydoctor/sphinx_ext/api_output.py
```diff
@@ -0,0 +1,69 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def on_build_finished(app, exception):
```
mthuurne, line 24: I'd really like to have type annotations on all new code. Prod me on IRC if you have trouble figuring out the right annotations.

pydoctor/sphinx_ext/api_output.py
```diff
@@ -0,0 +1,69 @@
...
+
+    args = []
+    for argument in app.config.pydoctor_args:
+        args.append(argument.format(**placeholders))
+
+    stream = StringIO()
+
+    old_cwd = os.getcwd()
+    try:
+        os.chdir(app.config.pydoctor_cwd)
```
mthuurne, line 46: Do we need to change the working dir? If passed absolute paths, pydoctor should be able to run from any directory, right?

pydoctor/sphinx_ext/api_output.py
```diff
@@ -0,0 +1,69 @@
...
+
+    placeholders = {
+        'outdir': app.outdir,
+        }
+
+    args = []
+    for argument in app.config.pydoctor_args:
+        args.append(argument.format(**placeholders))
+
+    stream = StringIO()
```
mthuurne, line 42: I think it's best if we make sure the stream is closed, by using a `with` block. While the garbage collector will probably do it otherwise, that might not happen immediately.

If I recall correctly, closing the stream does discard the buffered data, so it needs to happen after `getvalue()`.

mthuurne (approved):
Approved with comments: there are some details that still need attention, but I don't think another review round is necessary.

docs/source/conf.py
mthuurne: > Thanks a lot, this is a game changer for Sphinx users to use PyDoctor!

Is this usable for other projects? I was under the impression that, at least right now, it only works for pydoctor's own documentation.

mthuurne:
> @mthuurne can you please take a look at this PR.
> 
> I think that once this is merged it will make it easier to review UI / CSS / HTML PRs like #265

Yes, it will be very useful for PRs like that. I actually tried to find the output of #265 on RTD when I first submitted it and was slightly disappointed that it didn't work yet ;)

docs/source/conf.py
tristanlatr: > Is this usable for other projects?

Yes I think so. 

[This is a documentation draft]
As I understand it's pretty straight  forward to integrate pydoctor's build into sphinx with this:

- enable `pydoctor.sphinx_ext.build_apidocs` in `conf.py`
- configure `pydoctor_args`, `pydoctor_cwd`
- insert a `api/index.rst` file with a title and reference it from the Sphinx TOC tree

That's it, the API reference should be under the `api` folder when running your sphinx build!

docs/source/conf.py
adiroiban: I don't want to make things more complicated.... if escaping is required, the value should be passed by the developer.

This is a developer facing extension and not an end user extension...

yes... for pydoctor, our end users are the developers

docs/source/conf.py
adiroiban: Old habits die hard....

pydoctor/sphinx_ext/api_output.py
adiroiban: Yep... The buffer is freed https://github.com/python/cpython/blob/4a21e57fe55076c77b0ee454e1994ca544d09dc0/Modules/_io/stringio.c#L567

I will move it into a context manger. Thanks

pydoctor/sphinx_ext/api_output.py
adiroiban: We don't need it. Great call. Thanks!

pydoctor/sphinx_ext/api_output.py
adiroiban: Is there a way to run mypy on the diff code and complain if the code has missing / invalid type annotations ?

Something like `diff_cover` but for mypy ? :)

pydoctor/sphinx_ext/api_output.py
adiroiban: I think that we should adopt isort and black  ASAP so that the developers could forget about this and just focus on the code.

docs/source/conf.py
adiroiban: Well... this is no longer needed :)

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -0,0 +1,70 @@
...
+    from sphinx.application import Sphinx
+
+logger = logging.getLogger(__name__)
+
+
+def on_build_finished(app: 'Sphinx', exception: Exception) -> None:
+    """
+    Called when Sphinx build is done.
+    """
+    if not app.config.pydoctor_args:
```
adiroiban, line 38: @mthuurne 

Can you help with this warning ?

```
pydoctor/sphinx_ext/build_apidocs.py:38: error: Cannot determine type of 'config'  [has-type]
```

Thanks!

pydoctor/sphinx_ext/api_output.py
mthuurne: Not that I know of. What I want to do in the long term is set `disallow_untyped_defs=True`, so mypy will flag missing annotations as errors. But we're still some time away from that being an option for the whole code base.

We could add that option though and then exempt existing modules that don't have full annotations. That would at least force annotations to be added in new modules and in new code in modules that are fully annotated already.

mthuurne: As of #274, `disallow_untyped_defs=True` is now the default; several modules have overrides to turn it off, but at least new modules will have mandatory annotations from the start.

Issue #128, opened 2016-09-27
"invalid ref" to symbol that is present in module namespace


wsanchez:
In this build: https://buildbot.twistedmatrix.com/builders/documentation/builds/1752

There are numerous errors noted by pydoctor of this form:

{{{
twisted.application.runner._exit.ExitStatus.EX_CANTCREAT: invalid ref to 'ValueConstant' resolved as 'constantly.ValueConstant'
}}}

I find this confusing, as `ValueConstant` is a symbol defined in the module namespace, so I'm unsure why it's being called an "invalid ref" here.

mthuurne:
This message is logged when a short identifier is successfully resolved to a full name, but the full name isn't found in any Sphinx inventory. So the most likely cause is that pydoctor was run without a Sphinx inventory covering Constantly.

I agree that the message is confusing though. The problem isn't that the reference is invalid, but that it could not be found in the project itself or any of the given inventories.

PR #256, opened 2020-11-05
[Fixes #248] Non zero exit code on warnings.

### LINKED ISSUE #248: Report content problems with non-zero exit code
Once we have a robust logging system (see #112), it would be nice if pydoctor could return a non-zero exit code if it encountered content problems when generating docs.

For the system test, it is important though that we have an option to disable this feature, since during system tests we don't want content problems to fail the build, but we do want to know if pydoctor crashed or couldn't do its job for non-content reasons.

adiroiban:
Scope
======

This adds a '-W' CLI option to have pydoctor exit with non zero code on warnings.

Tries to fix #248 

Changes
=======

Use warning when guessing the project name. This was done to simplify testing.

Add sphinx doc and apidocs tox env for easy manual / CI testings.

I am still using the private `_warning` method as I think that it should be made public in a separate PR.

Drive by changes
==============

I have fixed many of the sphinx warnings.

I have added the documentation build process to static checks. It is red, but I plan to fix it with the on the fly generation of help output.

How to test
=========

Change settings in tox.ini and run `tox -e apidocs` and see the results.


.github/workflows/static.yaml
```diff
@@ -36,3 +36,7 @@ jobs:
     - name: Run pyflakes
       run: |
         tox -e pyflakes
+
+    - name: Run docs
+      run: |
+        tox -e docs
```
adiroiban, line 42: I can remove this from this PR and add it later...

As a first run, I wanted to see that Sphinx docs are failing on errors.

docs/source/conf.py
```diff
@@ -43,7 +43,6 @@
 
 # Configure intersphinx magic
 intersphinx_mapping = {
-    'pydoctor': ('https://pydoctor.readthedocs.io/api/', None),
```
adiroiban, line 46: this will be added later once we publish the API docs on Read The docs....I have a separate PR.

The Proof of Concept is already done

pydoctor/test/test_commandline.py
```diff
@@ -1,10 +1,14 @@
+from contextlib import redirect_stdout
```
mthuurne, line 1: pytest has the `capsys` fixture for capturing stdout/stderr. You can see examples of it in `pydoctor/test/test_epydoc2stan.py`.

pydoctor/driver.py
```diff
@@ -134,6 +134,10 @@ def getparser():
         '--buildtime', dest='buildtime',
         help=("Use the specified build time over the current time. "
               "Format: %s" % BUILDTIME_FORMAT))
+    parser.add_option(
+        '-W', action='store_true', dest='warnings_as_errors',
```
mthuurne, line 138: For consistency, I think it would be good to have a long form of this option as well. Maybe just `--warnings-as-errors`.

pydoctor/test/test_commandline.py
```diff
@@ -59,3 +63,45 @@ def test_cache_disabled_by_default() -> None:
...
+
+def test_cli_warnings_on_error() -> None:
+    """
+    The -W option is disabled by default.
+    """
+    options, args = driver.parse_args([])
+    assert options.warnings_as_errors == False
+
+    options, args = driver.parse_args(['-W'])
+    assert options.warnings_as_errors == True
```
mthuurne, line 76: I'm not sure we need a test case for this, especially since the other two test cases already check the behavior with and without the `-W` command line argument.

pydoctor/test/test_commandline.py
```diff
@@ -59,3 +63,45 @@ def test_cache_disabled_by_default() -> None:
...
+    """
+    output = StringIO()
+    with redirect_stdout(output):
+        exit_code = driver.main(args=[
+            '-v', '--testing',
+            'pydoctor/test/testpackages/basic/'
+            ])
+
+    assert exit_code == 0
+    assert '<None> guessing basic for project name' in output.getvalue()
```
mthuurne, line 91: The `<None>` doesn't look great; the user might think they hit a bug.

In general, I also prefer quotes around names (`guessing "basic" for`), so it's easier for the user to see which parts of the message are static and which are dynamic. I know this message isn't new, but if we're going to test the exact output it produces, now would be a good time to update it.

I notice that in the pydoctor code, sometimes inserted strings are not quoted, sometimes they are quoted in single quotes and sometimes in double quotes. It would be good to pick a single style and stick to it. I have a strong preference for quoted, but no real preference for single vs double quotes; what do you think?

pydoctor/driver.py
```diff
@@ -316,17 +320,20 @@ def main(args=sys.argv[1:]):
...
-        system.process()
 
         if system.options.projectname is None:
             name = '/'.join(ro.name for ro in system.rootobjects)
-            system.msg(
-                'warning',
-                'WARNING: guessing '+name+' for project name', thresh=-1)
+            system._warning(
+                current=None,
+                message='guessing {} for project name'.format(name),
```
mthuurne, line 328: You can use f-strings in pydoctor.

pydoctor/driver.py
```diff
@@ -316,17 +320,20 @@ def main(args=sys.argv[1:]):
...
+                detail='',
+                )
             system.projectname = name
         else:
             system.projectname = system.options.projectname
 
+        system.process()
```
mthuurne, line 335: Why is this statement moved? The initialization is still quite fragile, so changing the order of things should be done very carefully and only when needed.

pydoctor/driver.py
```diff
@@ -134,6 +134,10 @@ def getparser():
         '--buildtime', dest='buildtime',
         help=("Use the specified build time over the current time. "
               "Format: %s" % BUILDTIME_FORMAT))
+    parser.add_option(
+        '-W', action='store_true', dest='warnings_as_errors',
+        default=False,
+        help=("Exit with non zero code on warnings."))
```
mthuurne, line 140:
```suggestion
        help=("Return exit code 3 on warnings."))
```

mthuurne (changes requested):
I would like the CI and drive-by changes that are unrelated to warnings to be excluded from this PR, since it's complicated enough already.

pydoctor/driver.py
mthuurne: By the way, should this actually be a warning?

How likely is it that we guess the project name correctly? If it's unlikely, then it's good to warn, since the name is probably incorrect (but we have to put something there). If it's likely, then warning will either create noise in the logging or force the user to pass an additional command line option while auto-detection would have sufficed.

I think there are quite a few Python projects, especially tools and libraries, where the project name is equal to the top-level package name. In that case, the message should be at the "info" rather than "warning" level, which unfortunately would make this a poor choice to base the test case on.

pydoctor/test/test_commandline.py
adiroiban: I will use the pytest helpers... but I think that the current code is much easier to read.

I have no idea what is going on in https://github.com/twisted/pydoctor/blob/master/pydoctor/test/__init__.py#L25

And this can it harder for people to contribute to a public project.

I am ok to use pytest and all kind of bleending edge libs for private projects, but I think that the public libs should be as accessible as possible.


... I know... I am getting older :)

pydoctor/driver.py
adiroiban: I will go for `--warnings-as-errors`

pydoctor/test/test_commandline.py
adiroiban: I have created this tests as an unit test for the CLI arguments parsing.
The other tests are for CLI behaviour.

But with the introduction of the long format, I can reuse this test for the long form.

pydoctor/test/test_commandline.py
adiroiban: +1 for the quotes.

Any suggestion for a replacement for "<None>" ?

pydoctor/driver.py
adiroiban: Regarding f-string: long time python2.5 user here :) ... still learning the latest things from py3.

But I will update it.

Maybe use the pyupgrade tool to automatically complain about these things.

This was previously a "WARNING" .... so I keep it as such, since it makes it easier to write the tests :)
 
But I still that it should still be a warning, as is best for users to explicitly set the project name.
pydoctor did its best to guess the name, but an user should double check it.
Also, some people might want different cases for the name.

But if you have a strong feeling against making it a warning, I can make it a normal INFO message and will update the tests. I not hard to generate a warning in another way.

pydoctor/driver.py
adiroiban: It was just moved to have the project name during processing.

I think that we should not be afraid of developing pydoctor... if something breaks, an error is reported and fixed and more tests are added along the way

adiroiban:
Thanks for the review. 
I have moved the sphinx changes here https://github.com/twisted/pydoctor/pull/258

pydoctor/model.py
```diff
@@ -518,7 +518,7 @@ def _warning(self, current, message, detail):
         if current is not None:
             fn = current.fullName()
         else:
-            fn = '<None>'
+            fn = '.'
```
adiroiban, line 521: I went with a '.' instead of <None>

This is expected to be the full path to a file for which the report is reported...but we don't have any file for top level project warnings.

pydoctor/epydoc2stan.py
```diff
@@ -139,11 +139,11 @@ def resolve_identifier_xref(self, identifier: str, lineno: int) -> str:
 
         if identifier == fullID:
             self.obj.report(
-                "invalid ref to '%s' not resolved" % (identifier,),
+                "invalid ref to \"%s\" not resolved" % (identifier,),
```
adiroiban, line 142: not sure if you want single or double quoted in the human readable output.

I prefer double quotes for human readable as they stand out and are easier to read.

pydoctor/test/test_commandline.py
adiroiban: @mthuurne I have no idea what capsys is doing but I have lost a lot of time with it without any success so I went with simple redirect_stdout

Feel free to try to make the tests green using capsys ... some are working, other don't :(

pydoctor/model.py
```diff
@@ -500,6 +500,10 @@ def msg(self, section, msg, thresh=0, topthresh=100, nonl=False, wantsnl=True, o
                 return
             else:
                 self.once_msgs.add((section, msg))
+
+        if thresh < 0:
+            self.warnings.setdefault(section, []).append((section, msg))
```
adiroiban, line 505: maybe this needs a comment... or a reference to the ticket in which we implement a reported.

Anything below 0 is generated a  runtime warning / report

```
$ pydoctor --html-writer=pydoctor.test.InMemoryWriter pydoctor/test/testpackages/report_trigger -q
/home/adi/chevah/pydoctor/pydoctor/test/testpackages/report_trigger/__init__.py:8: Unknown field "bad_field"
/home/adi/chevah/pydoctor/pydoctor/test/testpackages/report_trigger/report_module.py:9: invalid ref to "BadLink" not resolved
```

pydoctor/driver.py
mthuurne: I agree that fear shouldn't be a reason to not make changes. However, many attributes in the model classes are initialized at undocumented times, so I have been bitten before by changes that seemed innocent.

I do think that it makes sense to initialize as much as possible before starting the processing, so let's keep this change.

pydoctor/epydoc2stan.py
mthuurne: Double quotes is fine with me.

I do prefer to single-quote the string in that case, since backslashes draw a lot of unnecessary attention when reading the code. Even if we choose to adopt Black at some point, it won't convert string literals if that leads to more backslashes.

This kind of thing is subjective though, so if you have a strong preference for escaped double quotes, keep them.

pydoctor/epydoc2stan.py
```diff
@@ -139,11 +139,11 @@ def resolve_identifier_xref(self, identifier: str, lineno: int) -> str:
...
                 'resolve_identifier_xref', lineno)
         else:
             self.obj.report(
-                "invalid ref to '%s' resolved as '%s'" % (identifier, fullID),
+                "invalid ref to \"%s\" resolved as '%s'" % (identifier, fullID),
```
mthuurne, line 146: There are two quoted values in this string, but only the first one was updated.

Note that #128 is about this message being unclear, so it would be a good idea to either improve the phrasing now as well or to move these message updates into a new PR.

pydoctor/model.py
mthuurne: Maybe something like this:
```py
    if thresh < 0:
        # Count messages with a negative threshold as warnings.
        self.warnings.setdefault(section, []).append((section, msg))
```
When we move to a logger/reporter, it makes sense to derive the log level from the threshold, since they serve similar purposes.

Instead of `setdefault` with an empty list, you can use `defaultdict(list)` as the collection type, then you can just write `warnings[section].append(...)`.

pydoctor/model.py
mthuurne: Do we still need this change? I didn't like how the "guessed project name" message contained `<None>`, but that message is no longer a warning.

I much prefer actual file paths + line numbers where possible and have been slowly making more and more code to log those instead of `fullName`. So I think that in the long run, the `_warning()` method is going to disappear.

pydoctor/test/__init__.py
```diff
@@ -25,3 +27,42 @@ class CapSys(Protocol):
         def readouterr(self) -> CaptureResult: ...
 else:
     CapLog = CaptureResult = CapSys = object
+
+
+class InMemoryWriter:
+    """
+    Minimal template writer that doesn't touches the filesystem.
```
mthuurne, line 34:
```suggestion
    Minimal template writer that doesn't touch the file system and doesn't record output.
```
This is useful to have. For other tests we might want to record the output into an in-memory buffer, but we can add that when we actually need it.

pydoctor/test/test_commandline.py
mthuurne: It has always worked well for me. I'll have a look.

pydoctor/test/test_commandline.py
mthuurne: The argument parsing is done by the standard library though, so what we're testing is that `default=False` is present in the option's definition and that `optparse` works as documented. It just seems like a lot of effort to test something low-risk.

But as it's already written and I don't expect this to require a lot of maintenance, it's fine to keep it.

tox.ini
```diff
@@ -70,6 +70,22 @@ commands =
...
+    pydoctor \
+    --add-package=pydoctor \
+    --project-name=pydoctor \
+    --project-url=https://github.com/twisted/pydoctor/ \
+    --html-viewsource-base=https://github.com/twisted/pydoctor/tree/master \
+    --html-output=build/apidocs \
+    --project-base-dir="{toxinidir}" \
+    --docformat=epytext \
+    --intersphinx=https://docs.python.org/3/objects.inv \
+    --make-html -W
```
mthuurne, line 86:
```suggestion
    --make-html --warnings-as-errors
```

pydoctor/test/test_commandline.py
```diff
@@ -59,3 +65,76 @@ def test_cache_disabled_by_default() -> None:
...
+            ])
+
+    assert exit_code == 0
+    assert '__init__.py:8: Unknown field "bad_field"' in stream.getvalue()
+    assert 'report_module.py:9: invalid ref to "BadLink" not resolved' in stream.getvalue()
+
+
+def test_main_return_non_zero_on_warnings() -> None:
+    """
+    When `-W` is used it returns 3 as exit code even when there are warnings.
```
mthuurne, line 128:
```suggestion
    When `-W` is used it returns 3 as exit code when there are warnings.
```

mthuurne (approved):
Some comments, but I agree with the overall approach.

pydoctor/epydoc2stan.py
adiroiban: I will revert to single quotes. Thanks for the feedback.

pydoctor/epydoc2stan.py
adiroiban: I will revert this change. It was a stupid change. Thanks!

pydoctor/model.py
adiroiban: I will revert this. Thanks!

pydoctor/test/test_commandline.py
adiroiban: This test is more about the fact that we have defined the  `--warnings-as-errors` ... I will keep it.

Since this test is easy to write, I prefer to have them as later it might help detect some strange side effects that we might introduce with "advances" command line parsing options.

So it's just a simple regression  tests since we don't use `--warnings-as-errors` in the other tests.

pydoctor/test/__init__.py
adiroiban: For me this is useful as it shows the public interface of the writer :)

I think that most probably this will be extended to record the actions.... or output.

In the current format it is only epydoc2stan... I don't think that it will handle RST.

adiroiban:
Thanks @mthuurne  for the review.

I tried to include all your feedback... InMemoryWriter ... is complicated. I tried to keep the name short.

It is not only a writed, but also a rendered...and it only renders epydoc.

I guess that will be extended later...maybe when support for RST will get serious :)

PR #258, opened 2020-11-05
Clean sphinx docs to reduce the number of warnings.

adiroiban:
This is just a cleanup of the current sphinx docs to reduce the number of warnings

I have added a tox env to make it easy to run the docs on the local system.

The help text output is generating errors, but it will be fixed in the PR in which the text is generated on the fly.


mthuurne (approved):
Looks good.

Issue #259, opened 2020-11-05
AttributeError: 'Package' object has no attribute 'all'


tristanlatr:
Hello, 

As the ultimate test and demo for people, I'm trying to generate the whole CPython standard Library API docs. 

And I ran into this error:

```
Traceback (most recent call last):
  File "/usr/local/bin/pydoctor", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/site-packages/pydoctor/driver.py", line 308, in main
    system.process()
  File "/usr/local/lib/python3.8/site-packages/pydoctor/model.py", line 726, in process
    self.processModule(mod)
  File "/usr/local/lib/python3.8/site-packages/pydoctor/model.py", line 710, in processModule
    builder.processModuleAST(ast, mod)
  File "/usr/local/lib/python3.8/site-packages/pydoctor/astbuilder.py", line 639, in processModuleAST
    self.ModuleVistor(self, mod).visit(ast)
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/usr/local/lib/python3.8/site-packages/pydoctor/astbuilder.py", line 78, in visit_Module
    self.default(node)
  File "/usr/local/lib/python3.8/site-packages/pydoctor/astbuilder.py", line 67, in default
    self.visit(child)
  File "/usr/local/Cellar/python@3.8/3.8.5/Frameworks/Python.framework/Versions/3.8/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/usr/local/lib/python3.8/site-packages/pydoctor/astbuilder.py", line 169, in visit_ImportFrom
    mod = self.system.getProcessedModule(modname)
  File "/usr/local/lib/python3.8/site-packages/pydoctor/model.py", line 690, in getProcessedModule
    return self.getProcessedModule(modname + '.__init__').parent
  File "/usr/local/lib/python3.8/site-packages/pydoctor/model.py", line 695, in getProcessedModule
    self.processModule(mod)
  File "/usr/local/lib/python3.8/site-packages/pydoctor/model.py", line 710, in processModule
    builder.processModuleAST(ast, mod)
  File "/usr/local/lib/python3.8/site-packages/pydoctor/astbuilder.py", line 639, in processModuleAST
    self.ModuleVistor(self, mod).visit(ast)
[...]

Issue #261, opened 2020-11-05
TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'


adiroiban:
It looks like pydoctor is not happy with the following top module docstring which is found in cpython/Lib/idlelib/debugger_r.py

```
"""Support for remote Python debugging.

Some ASCII art to describe the structure:

       IN PYTHON SUBPROCESS          #             IN IDLE PROCESS
                                     #
                                     #        oid='gui_adapter'
                 +----------+        #       +------------+          +-----+
                 | GUIProxy |--remote#call-->| GUIAdapter |--calls-->| GUI |
+-----+--calls-->+----------+        #       +------------+          +-----+
| Idb |                               #                             /
+-----+<-calls--+------------+         #      +----------+<--calls-/
                | IdbAdapter |<--remote#call--| IdbProxy |
                +------------+         #      +----------+
                oid='idb_adapter'      #

The purpose of the Proxy and Adapter classes is to translate certain
arguments and return values that cannot be transported through the RPC
barrier, in particular frame and traceback objects.

"""
```

to reproduce this, get the cpython code and run

```
$ pydoctor --docformat=restructuredtext --add-package Lib/idlelib/debugger_r.py --project-base-dir=.

```

Issue #184, opened 2020-01-28
__all__ re-exports don't work for CramMD5ClientAuthenticator


glyph:
`CramMD5ClientAuthenticator` is documented here:

https://twistedmatrix.com/documents/19.10.0/api/twisted.mail._cred.CramMD5ClientAuthenticator.html

except that should be a private API.

It's re-exported in the `__all__` of `twisted.mail.imap4` and `twisted.mail.smtp`, but doesn't show up in either of those places.

I'm not sure what the actual issue is here (it might be that this is kind of a nonsense thing to do?) but we should probably do something more coherent, or at least report an error.

tristanlatr:
I think I'm having the same issue with

https://tristanlatr.github.io/wpscan_out_parse/wpscan_out_parse.parser.html#parse_results_from_file which defined in `wpscan_out_parse/parser/__init__.py` and exported in `wpscan_out_parse/__init__.py` but it only shows in `wpscan_out_parse/parser/__init__.py`.

mthuurne:
I think there are actually 3 issues here:
- `CramMD5ClientAuthenticator` does not get re-exported at all
- pydoctor doesn't have a strategy for things that get re-exported more than once
- the module index doesn't differentiate between public and private modules (package pages do)

I don't know if re-exporting a name in multiple locations is something we should support: it seems confusing to just have the same class documentation at different places in the tree. Maybe one of them could be the official location and the others could link to it? That would be the preferred approach when something is moved and the old location is deprecated but still around for backwards compatibility.

mthuurne:
@tristanlatr `parse_results_from_file()` is re-exported when using the latest pydoctor; I think the fixes from #262 have helped here.

tristanlatr:
@mthuurne Thanks

mthuurne:
A long-term improvement discussed in #277 was to drop the reparenting of re-exported names and instead build a separate documentation layer on top of the code model. Doing so would allow one name from the code model to be published in more than one location in the documentation.

I'm still not sure whether it's a good idea to have one name published in more than one location though. For example if an exception type is documented in more than one place, it might surprise the user if catching one documented type will also catch a type documented in a different module. But perhaps that decision can be made by the developer on a case by case basis, instead of the tool deciding for them.

tristanlatr:
Still true with latest pydoctor (https://pydocbrowser.github.io/twisted/latest/twisted.mail._cred.CramMD5ClientAuthenticator.html).

PR #262, opened 2020-11-05
Fix getProcessedModule() for package

mthuurne:
When called with a package name, `getProcessedModule()` would force the `__init__` module to be processed, but then returned the `Package` object instead of the `Module` object. This is contrary to what its name suggests and what `visit_ImportFrom()` expects.

I also added type annotations to narrow down the place where the wrong type originated.

The purpose of this PR is to fix the bug described in #259.

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -5,7 +5,7 @@ import sys
 from functools import partial
 from itertools import chain
 from pathlib import Path
-from typing import Dict, Iterator, Mapping, Optional, Tuple
+from typing import Dict, Iterable, Iterator, Mapping, Optional, Tuple
 
 import astor
 from pydoctor import epydoc2stan, model
@@ -161,57 +161,79 @@ class ModuleVistor(ast.NodeVisitor):
                     return
                 parent = parent.parent
             if parent is not None:
-                modname = parent.fullName() + '.' + modname
+                modname = f'{parent.fullName()}.{modname}'
+
+        if node.names[0].name == '*':
+            self._importAll(modname)
+        else:
+            self._importNames(modname, node.names)
+
+    def _importAll(self, modname: str) -> None:
+        """Handle a C{from <modname> import *} statement."""
+
+        mod = self.system.getProcessedModule(modname)
+        if mod is None:
+            # We don't have any information about the module, so we don't know
+            # what names to import.
+            self.builder.warning("import * from unknown", modname)
+            return
+
+        self.builder.warning("import *", modname)
+
+        # Get names to import: use __all__ if available, otherwise take all
+        # names that are not private.
+        names = mod.all
+        if names is None:
+            names = (
+                name
+                for name in chain(mod.contents.keys(),
+                                  mod._localNameToFullName_map.keys())
+                if not name.startswith('_')
+                )
+
+        # Add imported names to our module namespace.
+        _localNameToFullName = self.builder.current._localNameToFullName_map
+        expandName = mod.expandName
+        for name in names:
+            _localNameToFullName[name] = expandName(name)
+
+    def _importNames(self, modname: str, names: Iterable[ast.alias]) -> None:
+        """Handle a C{from <modname> import <names>} statement."""
 
         mod = self.system.getProcessedModule(modname)
         if mod is not None:
-            assert mod.state in [model.ProcessingState.PROCESSING,
-                                 model.ProcessingState.PROCESSED]
             expandName = mod.expandName
         else:
             expandName = lambda name: f'{modname}.{name}'
+
         _localNameToFullName = self.builder.current._localNameToFullName_map
-        for al in node.names:
-            fromname, asname = al.name, al.asname
-            if fromname == '*':
-                if mod is None:
-                    self.builder.warning("import * from unknown", modname)
-                    return
-                self.builder.warning("import *", modname)
-                if mod.all is not None:
-                    names = mod.all
-                else:
-                    names = (
-                        k
-                        for k in chain(mod.contents.keys(),
-                                       mod._localNameToFullName_map.keys())
-                        if not k.startswith('_')
-                        )
-                for n in names:
-                    _localNameToFullName[n] = expandName(n)
-                return
+        for al in names:
+            orgname, asname = al.name, al.asname
             if asname is None:
-                asname = fromname
+                asname = orgname
+
+            # Handle re-exporting of imported names.
             if isinstance(self.builder.current, model.Module) and \
-                   self.builder.current.all is not None and \
-                   asname in self.builder.current.all and \
-                   modname in self.system.allobjects:
+                    self.builder.current.all is not None and \
+                    asname in self.builder.current.all and \
+                    modname in self.system.allobjects:
                 mod = self.system.allobjects[modname]
                 if isinstance(mod, model.Module) and \
-                       fromname in mod.contents and \
-                       (mod.all is None or fromname not in mod.all):
+                        orgname in mod.contents and \
+                        (mod.all is None or orgname not in mod.all):
                     self.system.msg(
                         "astbuilder",
                         "moving %r into %r"
-                        % (mod.contents[fromname].fullName(),
+                        % (mod.contents[orgname].fullName(),
                            self.builder.current.fullName()))
-                    ob = mod.contents[fromname]
+                    ob = mod.contents[orgname]
                     ob.reparent(self.builder.current, asname)
                     continue
-            if isinstance(
-                self.system.objForFullName(modname), model.Package):
-                self.system.getProcessedModule(modname + '.' + fromname)
-            _localNameToFullName[asname] = expandName(fromname)
+
+            if isinstance(self.system.objForFullName(modname), model.Package):
+                self.system.getProcessedModule(f'{modname}.{orgname}')
+
+            _localNameToFullName[asname] = expandName(orgname)
 
     def visit_Import(self, node):
         """Process an import statement.
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -184,12 +184,12 @@ class ModuleVistor(ast.NodeVisitor):
         # names that are not private.
         names = mod.all
         if names is None:
-            names = (
+            names = [
                 name
                 for name in chain(mod.contents.keys(),
                                   mod._localNameToFullName_map.keys())
                 if not name.startswith('_')
-                )
+                ]
 
         # Add imported names to our module namespace.
         _localNameToFullName = self.builder.current._localNameToFullName_map
@@ -206,29 +206,37 @@ class ModuleVistor(ast.NodeVisitor):
         else:
             expandName = lambda name: f'{modname}.{name}'
 
-        _localNameToFullName = self.builder.current._localNameToFullName_map
+        # Fetch names to export.
+        current = self.builder.current
+        if isinstance(current, model.Module):
+            exports = current.all
+            if exports is None:
+                exports = []
+        else:
+            # Don't export names imported inside classes or functions.
+            exports = []
+
+        _localNameToFullName = current._localNameToFullName_map
         for al in names:
             orgname, asname = al.name, al.asname
             if asname is None:
                 asname = orgname
 
-            # Handle re-exporting of imported names.
-            if isinstance(self.builder.current, model.Module) and \
-                    self.builder.current.all is not None and \
-                    asname in self.builder.current.all and \
-                    modname in self.system.allobjects:
-                mod = self.system.allobjects[modname]
-                if isinstance(mod, model.Module) and \
-                        orgname in mod.contents and \
-                        (mod.all is None or orgname not in mod.all):
-                    self.system.msg(
-                        "astbuilder",
-                        "moving %r into %r"
-                        % (mod.contents[orgname].fullName(),
-                           self.builder.current.fullName()))
+            # Move re-exported objects into current module.
+            if asname in exports and mod is not None:
+                try:
                     ob = mod.contents[orgname]
-                    ob.reparent(self.builder.current, asname)
-                    continue
+                except KeyError:
+                    self.builder.warning("cannot find re-exported name",
+                                         f'{modname}.{orgname}')
+                else:
+                    if mod.all is None or orgname not in mod.all:
+                        self.system.msg(
+                            "astbuilder",
+                            "moving %r into %r" % (ob.fullName(), current.fullName())
+                            )
+                        ob.reparent(current, asname)
+                        continue
 
             if isinstance(self.system.objForFullName(modname), model.Package):
                 self.system.getProcessedModule(f'{modname}.{orgname}')
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -200,14 +200,19 @@ class ModuleVistor(ast.NodeVisitor):
     def _importNames(self, modname: str, names: Iterable[ast.alias]) -> None:
         """Handle a C{from <modname> import <names>} statement."""
 
+        # Process the module we're importing from.
+        # If we're importing from a package, 'mod' will be the __init__ module.
         mod = self.system.getProcessedModule(modname)
-        if mod is not None:
-            expandName = mod.expandName
-        else:
-            expandName = lambda name: f'{modname}.{name}'
+        obj = self.system.objForFullName(modname)
 
         # Are we importing from a package?
-        is_package = isinstance(self.system.objForFullName(modname), model.Package)
+        is_package = isinstance(obj, model.Package)
+        assert is_package or obj is mod or mod is None
+
+        if obj is not None:
+            expandName = obj.expandName
+        else:
+            expandName = lambda name: f'{modname}.{name}'
 
         # Fetch names to export.
         current = self.builder.current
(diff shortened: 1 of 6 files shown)


mthuurne:
This PR causes some regressions when processing Twisted:
```
src/twisted/protocols/ftp.py:2357: probable interface twisted.protocols.ftp.IFinishableConsumer not marked as such
src/twisted/conch/insults/insults.py:453: probable interface twisted.conch.insults.insults.ITerminalTransport not marked as such
src/twisted/cred/test/test_cred.py:73: probable interface twisted.cred.test.test_cred.IDerivedCredentials not marked as such
src/twisted/conch/telnet.py:322: probable interface twisted.conch.telnet.ITelnetProtocol not marked as such
src/twisted/internet/testing.py:944: invalid ref to 'twisted.test.proto_helpers.EventLoggingObserver' resolved as 'testing.EventLoggingObserver'
```
So I'll have to dig deeper.

mthuurne:
It looks like the handling of `ImportFrom` nodes (`from <package> import <something>`) is broken:
- before this PR, imported names would always be resolved from the package, which crashed when "something" is `*`
- in my first fix, imported names would always be resolved from the `__init__` module, which is correct when "something" is `*` or a specific name from `__init__.py`, but will fail to find subpackages or submodules

The name resolving is done using the `expandName()` method. The method implementation seems fine, what matters is which object we call it on: the `Package` itself or the `Module` object for the package's `__init__` module.

From running some tests on Python 3.8, I think it should work like this:
- when importing `*`, we always import from the `__init__` module
- when importing specific names, we have to look them up one at a time, first in the `__init__` module and if not found then in the package

Note that `Package._localNameToFullName()` does delegate to the `__init__` module, but it does so only when there are no hits in its `contents` dictionary, which is the wrong order for `ImportFrom`.

mthuurne:
Perhaps the delegation happens in the wrong direction: if you `import a.b`, the name `a` will be the `a.__init__` module. So `a.b` is actually the name `b` within the module `a.__init__`.

Note that `a.__init__.b` does not appear until the module `a.b` has been imported. When it is imported, it will overwrite any re-existing `a.__init__.b`.

Since pydoctor is doing static analysis, we probably should treat all modules that can be imported as having been imported already, which means that submodules should be given priority over local names in the `__init__` module after all.

mthuurne:
I think I got the implementation correct now, but I still don't have a unit test case that demonstrates the regression.

mthuurne:
I managed to construct a test case that demonstrates the regression.

pydoctor/test/test_astbuilder.py
```diff
@@ -449,6 +458,38 @@ def f(): pass
...
+    ''', modname='__init__', parent_name='a', system=system)
+    mod_b = fromText('''
+    from a import f
+    ''', modname='b', system=system)
+    assert mod_b.resolveName('f') == mod_a.contents['f']
+
+
+@systemcls_param
+def test_import_module_from_package(systemcls: Type[model.System]) -> None:
+    """Importing a module from a package should not look in __init__ module."""
```
adiroiban, line 477: I guess that this is just a  regression tests ?

```suggestion
    """
    Importing a module from a package should not look in __init__ module.
    
    In this test the following hierarchy is constructed:
    a/__init__.py
    a/b.py
    c.py
    
    When c.py has content as below, `f` is resolved as a member if `b.py`:
        from a import b
        f = b.f
    """
```

Or maybe change the name of `f` to a name that describe its purpose in the test.

mthuurne: > I guess that this is just a regression tests ?

In the sense that this is the test case that demonstrates the regression that was introduced halfway the PR, yes. But it also tests a fundamental property of the `from ... import` handling, so in that sense it's more of a test that was missing instead of one that specifically exists to check for a regression.

I'll expand the documentation.

pydoctor/model.py
```diff
@@ -752,18 +749,21 @@ def readd(o):
...
 
-    def getProcessedModule(self, modname):
+    def getProcessedModule(self, modname: str) -> Optional[_ModuleT]:
         mod = self.allobjects.get(modname)
         if mod is None:
             return None
         if isinstance(mod, Package):
-            return self.getProcessedModule(modname + '.__init__').parent
+            initModule = self.getProcessedModule(modname + '.__init__')
+            assert initModule is not None
```
adiroiban, line 758: Is this "out of test code" assertion required ?

I was hoping that we now have enough unit tests to see them fail if this is None

pydoctor/model.py
```diff
@@ -353,18 +352,16 @@ def documentation_location(self) -> DocLocation:
         else:
             return DocLocation.OWN_PAGE
 
-    def setup(self):
+    def setup(self) -> None:
         super().setup()
-        self.all = None
+        self.all: Optional[Collection[str]] = None
```
adiroiban, line 357: I think that it would help to add as a documentation comment all that you know/think so far that `all` should do.

I see that it is used in astbuilder and it looks like it contains the members define in the `__all__`` member of a module.

----------

Also, by reading the code it was not clear why we need a Package and a Module inside model.py

It looks like Package is some sort of "wrapper" for the __init__ module.

pydoctor/astbuilder.py
```diff
@@ -153,65 +153,105 @@ def visit_ImportFrom(self, node):
...
+        """Handle a C{from <modname> import <names>} statement."""
+
+        # Process the module we're importing from.
+        # If we're importing from a package, 'mod' will be the __init__ module.
+        mod = self.system.getProcessedModule(modname)
+        obj = self.system.objForFullName(modname)
+
+        # Are we importing from a package?
+        is_package = isinstance(obj, model.Package)
+        assert is_package or obj is mod or mod is None
```
adiroiban, line 210: A minor question of style... do we need the assertions.
Also, for me `is_package` is a great variable name and the comment is not needed.

adiroiban (approved):
Thanks for working on this.

Sorry... I am not able to do a good review now.

I don't know too much about AST and epydoc in general.

I am not sure why we need the Package class... and why can't  we just have  a
Module.isPackage() method or Module.is_package attribute that is True when the module name is `__init__`

If you can't find another reviewer, I think that this is an improvement over the current state and it should be merged.

My suggestion is to update the system tests based on this diff... the idea is that we can use cpython code as a non-Twisted playground.

```diff
diff --git a/.github/workflows/system.yaml b/.github/workflows/system.yaml
index c54fc4a..236795b 100644
--- a/.github/workflows/system.yaml
+++ b/.github/workflows/system.yaml
@@ -32,3 +32,7 @@ jobs:
     - name: Generate Twisted API docs
       run: |
         tox -e twisted-apidoc
+
+    - name: Generate CPython API docs
+      run: |
+        tox -e cpython-apidoc
diff --git a/tox.ini b/tox.ini
index 425c822..0b852f9 100644
--- a/tox.ini
+++ b/tox.ini
@@ -51,6 +51,17 @@ commands =
     twisted-apidoc: /bin/cat {toxworkdir}/twisted-apidocs.log
 
 
+[testenv:cpython-apidocs]
+description = Build CPython API documentation (for the supported modules)
+
+commands =
+    rm -rf {toxworkdir}/cpython
+    git clone --depth 1 --branch master https://github.com/python/cpython.git {toxworkdir}/cpython
+    pydoctor --add-package={toxworkdir}/cpython/Lib/ctypes --project-base-dir={toxworkdir}/cpython
+    pydoctor --add-package={toxworkdir}/cpython/Lib/tkinter --project-base-dir={toxworkdir}/cpython
+    pydoctor --add-package={toxworkdir}/cpython/Lib/test --project-base-dir={toxworkdir}/cpython
+
+
 [testenv:mypy]
 description = run mypy (static type checker)
```

mthuurne:
> Thanks for working on this.
> 
> Sorry... I am not able to do a good review now.
> 
> I don't know too much about AST and epydoc in general.

Maybe @moshez would be willing to have a look?

> I am not sure why we need the Package class... and why can't we just have a
> Module.isPackage() method or Module.is_package attribute that is True when the module name is `__init__`

I was about to disagree with packages being modules, but then looked up the [definition of a package](https://docs.python.org/3/reference/import.html#packages) and it turns out that packages are in fact special modules.

So it would make sense to change pydoctor's model as you describe. But that would be quite invasive, so I'd rather land this PR first.

> If you can't find another reviewer, I think that this is an improvement over the current state and it should be merged.

I agree: even if the old/current design doesn't match the language definition exactly, the behavior is now more correct and better tested than it was before. It fixes the three crashes from #259 and it might help with #184 as well (unverified).

> My suggestion is to update the system tests based on this diff... the idea is that we can use cpython code as a non-Twisted playground.

I think it's useful to have more system tests than just Twisted, also because Twisted doesn't use RST. On the other hand, we have to keep the CI cycle fast, so I'm a bit hesitant to include a project as large as CPython. Would it be an option to run it as a post-merge test? Then we can see in practice how often the test will find issues we wouldn't otherwise have found, without slowing down CI in the mean time.

pydoctor/model.py
mthuurne: I think the assert was added at some point to tell mypy that `initModule` was safe to call `Module` methods on, but in the method's final form there is nothing called on it, so I'll remove it.

pydoctor/model.py
mthuurne: I agree about more documentation, but that's a problem we'll have to solve bit by bit, since there is quite an amount of technical debt. That said, now is a better time than average to pay off this particular piece of debt, so I'll add a docstring.

pydoctor/astbuilder.py
mthuurne: I think this assertion is useful, as it makes a non-trivial statement about the code. This is illustrated by the fact that originally I forgot about the `or mod is None` part, which can happen if there is a name clash between a module and a variable in `__init__.py` within the same package.

One reason to use asserts in situations like this is that it acts as a formal and verified statement about the code. If the same statement were included as a comment, there would be more room for misinterpretation and if the statement is not or no longer true, we wouldn't know.

Another reason to use asserts is that they're active across all unit and system tests, so they help cover a much larger state space. Despite having around 90% branch coverage, there are still quite a few bugs in the `astbuilder` and `model` modules that we're not catching, because either the test cases don't do enough validations or a bug only occurs in particular combinations of states that are tested separately.

adiroiban: Thanks for the info. We can keep it.

Another matter of style: Move `is_package = isinstance(obj, model.Package)` on line 250 so that it is closer to the place where it is used.

Will make the code easier to read...
I was searching for placer between line 211 and 250 to see in which other place it is used.

adiroiban:
> I think it's useful to have more system tests than just Twisted, also because Twisted doesn't use RST. On the other hand, we have to keep the CI cycle fast, so I'm a bit hesitant to include a project as large as CPython.

It took me 27 second to `run tox -e cpython-apidocs` ... out of which 25 were cloning the repo.

At this point, we don't need to run pydoctor on the full cpython...but just on the parts that failed in the past... to have it as a regression test.

What we can/should do, is run cpython-apidocs in parallel with twisted-apidocs.
We still have enough free concurrent builds (20 in total if we just use linux)

twisted-apidocs are already taking close to  2 minutes... so as long as we run cpythong-apidocs in parallel and they take less than 2 minutes to run, we don't increase the total test time

Diff to run system tests in parallel can look like this:

```diff
diff --git a/.github/workflows/system.yaml b/.github/workflows/system.yaml
index c54fc4a..d6aa2db 100644
--- a/.github/workflows/system.yaml
+++ b/.github/workflows/system.yaml
@@ -10,6 +10,10 @@ jobs:
   system_tests:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        tox_target: [twisted-apidoc, cpython-apidoc]
+
     steps:
     - uses: actions/checkout@v2
 
@@ -31,4 +35,4 @@ jobs:
 
     - name: Generate Twisted API docs
       run: |
-        tox -e twisted-apidoc
+        tox -e ${{ matrix.tox_target }}
```

pydoctor/astbuilder.py
mthuurne: Line 250 is inside the loop, while `is_package` does not depend on the thing we're iterating through. So I don't think it should be inside the loop, both for clarity and for performance.

I thought about moving the re-export code to a separate method, since that takes up the bulk of the loop's body, but doing that would create a method with 3 arguments (`mod`, `orgname` and `asname`), so it I'm not sure it would improve readability.

mthuurne:
> > I think it's useful to have more system tests than just Twisted, also because Twisted doesn't use RST. On the other hand, we have to keep the CI cycle fast, so I'm a bit hesitant to include a project as large as CPython.
> 
> It took me 27 second to `run tox -e cpython-apidocs` ... out of which 25 were cloning the repo.
> 
> At this point, we don't need to run pydoctor on the full cpython...but just on the parts that failed in the past... to have it as a regression test.

Are regressions more likely to come from these three than from other parts of the code though? Now that we have mypy coverage of that part of the code, hitting an unexpected `None` there won't happen again. If we spend 25 seconds on cloning the source, we might as well use more of that code.

> What we can/should do, is run cpython-apidocs in parallel with twisted-apidocs.
> We still have enough free concurrent builds (20 in total if we just use linux)
>
> twisted-apidocs are already taking close to 2 minutes... so as long as we run cpythong-apidocs in parallel and they take less than 2 minutes to run, we don't increase the total test time

That is true: as long as it doesn't take longer than twisted-apidocs, it's essentially free. I'll add your suggestions, but I'd like to try and see how long the full cpython would take. Or maybe we can pick some random files plus a list of files that previously failed.

mthuurne:
By the way, could we put the cpython source in a cached location? Then we would not need to do a full clone every time.

adiroiban:
I think that you can try and see how much faster is caching the cpython code... the problem is what key to invalidate the cache.

I think that you can also run the full cpython documentation... including tests.

I just went with those 3 packages as those were the one triggering these failures.... and I didn't wanted to make the test run longer....
but with a parallel run, I think that we can afford to run more packages. 


The ticket https://github.com/twisted/pydoctor/issues/259 has a list of modules ...

One option is to convert the whole cpython/Lib into a package via `$ touch cpython/Lib/__init__.py` and then run `$ pydoctor --add-package=cpython/Lib --project-base-dir=cpython`

mthuurne:
I'm also running into a problem: if there are docstring syntax errors, pydoctor currently returns exit code 2. Should we consider docstring syntax errors as errors, since the output is likely wrong? Or should we consider them warnings, as we are still able to generate the full output even though it may be wrong in places? Unlike warnings, in these cases we're certain that the input is wrong, but like warnings, the problem has a limited impact on the end result.

In any case, we shouldn't be using exit code 2: it should either be 0/3 (depending on `--warnings-as-errors`) or a new exit code like 4.

We're getting syntax errors reported because we're not selecting RST as the format yet, but even if we do that, there is the question of how to treat docstring errors should they exist. In any case, it's a content problem and it shouldn't fail the system test, but the question is whether pydoctor should return 0 or whether we filter the exit code in the tox rule and only fail the test on exit code 1.

mthuurne:
> I think that you can try and see how much faster is caching the cpython code... the problem is what key to invalidate the cache.

Do we ever need to invalidate the cache? Maybe if it became corrupted, but under normal circumstances we could just `git pull` and follow CPython developments.

mthuurne:
It seems you got a 2 seconds run time because only `ctypes` was processed: tox stops at the first error and `ctypes` is significantly smaller than `tkinter` and `test`. If I process all three, with the proper docformat, it takes about 2.5 minutes on my PC.

adiroiban (changes requested):
I am running the following command based on this branch and it still fails :(

```
pydoctor --add-package=cpython/Lib/test --project-base-dir=cpython
```

adiroiban:
On my laptop, just running `touch cpython/Lib/__init__.py; pydoctor --add-package=cpython/Lib --project-base-dir=cpython`

took `real	1m17.822s`

I had to use this hack to 
```
                try:
                    mod.all
                except:
                    mod.all = []
```

This can help just with the AST traversal.

Using `--docformat=restructuredtext` will not make it to the end :)

So my suggestion is to still run pydoctor on cpython, but at this stage without RST... just to check all the imports

mthuurne:
> I am running the following command based on this branch and it still fails :(
> 
> ```
> pydoctor --add-package=cpython/Lib/test --project-base-dir=cpython
> ```

In what way does it fail? It runs fine here.

mthuurne:
> I had to use this hack to
> 
> ```
>                 try:
>                     mod.all
>                 except:
>                     mod.all = []
> ```
> 
> This can help just with the AST traversal.

Are you sure you're running the right version of pydoctor? Because that sounds like the original bug from #259 that was fixed.

> Using `--docformat=restructuredtext` will not make it to the end :)

Yes, the problem from #261 is still there.

> So my suggestion is to still run pydoctor on cpython, but at this stage without RST... just to check all the imports

In that case, I think we should add a command line option to disable the output generation altogether, since that is where the bulk of the time is spent. Parsing all of CPython takes on 24 seconds on my PC, while generating output takes about a minute extra.

Surprisingly, it doesn't matter all that much which docformat I select: `plaintext` isn't much faster than `epytext`. That would suggest that it is the HTML output rather than the docstring parsing that is the bottleneck.

mthuurne:
The new system test passes in CI, so I think that the problem you ran into is specific to your setup.

adiroiban (approved):
Great work. Thanks!

adiroiban:
I think that this is ready to be merged.... it is a huge improvement (running the regression AST traveral on cpython.

I am happy to see the new tox env, as they allow us to re-run cpython in an easy to reproduce way.

Thanks!

PR #263, opened 2020-11-06
CSS: Add visual contrast between fields and members

### LINKED ISSUE #260: CSS: Better Emphasis
I noticed the "new" pydoctor CSS: 
- Do not emphasis on element definition, it's just regular plaintext
- Do not emphasis on parameters name, it's just regular plaintext
- Do not emphasis on flags titles, i.e. "Parameters", it's just regular text

To me, this is a "layout regression" from older versions of pydoctor 

New CSS: https://twistedmatrix.com/documents/20.3.0/api/twisted.web.client.HTTPClientFactory.html

Older CSS exemples: https://twistedmatrix.com/documents/12.3.0/api/twisted.web.client.HTTPClientFactory.html or https://twistedmatrix.com/documents/12.3.0/api/twisted.web.client.html

Would be great to have the same kind of clarity.
If it's only a matter of CSS, I can try to have a look at it.

tristanlatr:
Closes #260 

It looks like  this now 

by default 2020-11-06 at 12 48 58 AM

### DIFF
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -198,7 +198,7 @@ code > .code {
     border-left-color: #03a9f4;
     border-left-width: 1px;
     border-left-style: solid;
-    background: #f6f6f6;
+    background: #fafafa;
 }
 
 #childList > div .functionHeader {
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -142,6 +142,15 @@ tr.private {
     margin-top: 10px;
 }
 
+.fieldArg {
+    font-weight: bold;
+    color: #c7254e;
+}
+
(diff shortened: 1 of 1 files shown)


adiroiban (approved):
It looks good. Thanks.

'CSS: More clarity' is a very generic name for a PR.

Maybe: 'Add visual contrast between element types.'

tristanlatr:
The functions' background is almost the same color as the doctest's...
I'll change it for something even lighter

by default 2020-11-06 at 2 25 44 PM

tristanlatr:
by default 2020-11-06 at 2 34 12 PM

It's better now

adiroiban:
Thanks. I have merged. I have no idea what rules to use to decide if a CSS change is ok or not..
So this looks good to me :)

PR #265, opened 2020-11-08
Use sticky top bar as navigation bar

mthuurne:
This moves some information elements out of the top bar, so a minimal top bar is left that we can then make sticky. This allows for quick navigation up the hierarchy.

### DIFF
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -67,11 +67,9 @@ class Documentable:
     @ivar parent: ...
     @ivar parentMod: ...
     @ivar name: ...
-    @ivar documentation_location: ...
     @ivar sourceHref: ...
     @ivar kind: ...
     """
-    documentation_location = DocLocation.OWN_PAGE
     docstring: Optional[str] = None
     system: 'System'
     parsed_docstring: Optional[ParsedDocstring] = None
@@ -79,6 +77,13 @@ class Documentable:
     linenumber = 0
     sourceHref: Optional[str] = None
     kind: str
+
+    @property
+    def documentation_location(self) -> DocLocation:
+        """Page location where we are documented.
+        The default implementation returns L{DocLocation.OWN_PAGE}.
+        """
+        return DocLocation.OWN_PAGE
 
     @property
     def css_class(self):
@@ -341,6 +346,13 @@ class Module(CanContainImportsDocumentable):
     kind = "Module"
     state = ProcessingState.UNPROCESSED
 
+    @property
+    def documentation_location(self) -> DocLocation:
+        if self.name == '__init__':
+            return DocLocation.PARENT_PAGE
+        else:
+            return DocLocation.OWN_PAGE
+
     def setup(self):
         super().setup()
         self.all = None
@@ -385,7 +397,10 @@ class Class(CanContainImportsDocumentable):
 
 
 class Inheritable(Documentable):
-    documentation_location = DocLocation.PARENT_PAGE
+
+    @property
+    def documentation_location(self) -> DocLocation:
+        return DocLocation.PARENT_PAGE
 
     def docsources(self):
         yield self
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -8,7 +8,10 @@
     top: 0;
     background-color: #fff;
 }
-
+.page-header h1 {
+    overflow: hidden;
+    max-height: 44px;
+}
 
 ul {
     margin-top: 10px;
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -4,6 +4,9 @@
 
 .page-header {
     margin-top: 22px;
+    position: sticky;
+    top: 0;
+    background-color: #fff;
 }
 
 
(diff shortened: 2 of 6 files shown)


adiroiban:
I have merged master so that it will be easier to test by just running:

```
$ tox -e apidocs
$ firefox build/apidocs/index.html
```

tristanlatr:
This is so so nice!

A few things:
- With long class names it shows only "class", not "class documentation" in the gray text. I.e. for `pydoctor.epydoc.markup.restructuredtext.ParsedRstDocstring`. May be the gray text could be displayed in a second line no matter what, and hidden when scrolling. 
- The "View source" and "View In Hierarchy" should always keep the same design across modules and classes. When it's a class it shows `(source)` but when it's a module it's shows "View Source". 
- Also the "View in Hierarchy" link could be moved up not to change too much from the curent place

pydoctor

- The method definition should not have the same red-ish color because we'll confuse it with actual links.
May be something like 
by default 2020-11-08 at 2 22 46 PM
So we know only the inherited class is clickable :)

tristanlatr:
This can be obtained with adding the following CSS:
```
div.extrasDocstring > p:nth-child(1) > code{
    color: #000000;
}
```
But may be adding a "signature" class is the best?

mthuurne:
> * With long class names it shows only "class", not "class documentation" in the gray text. I.e. for `pydoctor.epydoc.markup.restructuredtext.ParsedRstDocstring`. May be the gray text could be displayed in a second line no matter what, and hidden when scrolling.

At first the "class documentation" text would wrap on long class names, which meant the sticky navigation bar would take up more space for what is mostly redundant information. That's why I did some CSS trickery to make it invisible when it wraps.

Always putting it on a second non-sticky line, as you suggest, might be a simpler and more consistent solution for the same problem.

> * The "View source" and "View In Hierarchy" should always keep the same design across modules and classes. When it's a class it shows `(source)` but when it's a module it's shows "View Source".

I was aiming for consistency with how functions are documented, which have a "(source)" link on their `def` line. However, as modules don't have a definition line, I couldn't do something similar there. Perhaps using the "(source)" syntax for modules as well would be more consistent, even if there is no definition to attach it to.

> * Also the "View in Hierarchy" link could be moved up not to change too much from the curent place

"View In Hierarchy" is a kind of "read more" link for a class's inheritance information, so I thought that placing it below the list of subclasses made the flow more logical: the `class` line shows the superclasses, the next line lists the subclasses and finally there is the link that tells you everything else you might want to know about inheritance.

From a technical point of view, it's easy to put "View In Hierarchy" above the `class` line, but I'm not sure that's actually better.

> * The method definition should not have the same red-ish color because we'll confuse it with actual links.
>       May be something like
>       by default 2020-11-08 at 2 22 46 PM
>       So we know only the inherited class is clickable :)

I agree we have too much red text, but I didn't want to do too much in a single PR. The way the function `def` lines are rendered should also be updated, for example.

adiroiban (approved):
This looks good and can be merged :) Thanks.

I will not argue in terms of colors  :)

I would say that the `module documentation | package documentation | class documentation` text from the current sticky header is not really useful.... especialy the `documentation` page.

We are in apidocs so everything is about documentation.

What I use quite often is "show source" 
My suggestion is the have just class "Module/Class/Package" words at the start... or create an icon for each name...then the "breadcrumbs" and then show source.

image

mthuurne:
> I would say that the `module documentation | package documentation | class documentation` text from the current sticky header is not really useful.... especialy the `documentation` page.
>
> We are in apidocs so everything is about documentation.

Yes, the word is redundant, but having only one short word (class/module/package) in a different font + color doesn't look good. So I think "documentation" is cosmetic here.

> What I use quite often is "show source"

Often enough that you'd want it in the sticky navigation bar? I mostly use it when the documentation is missing or unclear, so ideally I wouldn't have to use it often ;)

I was thinking of moving the "Show Private API" link into the navigation bar, since that is also an element that is always accessible, but currently it can overlap with the body if your browser window isn't very wide. I often work with a two-column window arrangement, with my text editor on one side and docs or terminal on the other side.

> My suggestion is the have just class "Module/Class/Package" words at the start... or create an icon for each name...then the "breadcrumbs" and then show source.
> 
> image

We have dedicated colors for packages, modules and classes, so in theory we could use that as the background color for an additional hint for the documentation category. The current colors for them are very similar though, so this idea would only work if we would pick different category colors.

I think it's useful to have the breadcrumbs start at the same position always, which would work better with an icon than a word, although "Module/Class/Package" don't vary all that much in length. It may not be easy to create good icons for such abstract words though: package is easy, but how do you draw a module? An alternative would be to show a capital letter M/C/P on a panel.

Some full names can be quite long; I'm using `twisted.application.internet.StreamServerEndpointService` as a test case. So I'd rather not put too much other text on that line, especially since it's sticky and therefore takes up a lot of space if it wraps.

Icons take up less space than words, so if we had a "view source" icon, space would be less of an issue. But the word "source" is clearer than any icon can be.

mthuurne:
I'll merge the current state now, so we can (re)base other PRs upon it. That doesn't mean the design is final: we can continue to discuss and fine-tune it in new PRs.

adiroiban:
@mthuurne  thanks for the PR. I undestand that my suggestion is 100% subjective... and is true that "show source" usage is the side effect of bad docs :)

Issue #257, opened 2020-11-05
Sidebar boxes with autogenerated indexes


tristanlatr:
Hello,

I'm using `pydoctor` as the only tool to document a little library https://mfesiem.github.io/docs/msiempy/msiempy.html

As you can see, there are numerous titles, and it would be great if users could be navigate rapidly to one another. 

In this issue I'm proposing to add a sidebar (it could be an option or enabled by default), just like in JavaDoc (i.e. https://docs.oracle.com/javase/7/docs/api/index.html?overview-summary.html) to all pages. 

JavaDoc do not create summary pages for modules, PyDoctor does, but it doesn't change anything...

Several level (boxes) of indexes can be identified:

1/ The table of contents. The result of ``.. contents::``. Would be present in module and classes documentation if titles are defined in the documented element's docstring. 
2/ The Classes / Interfaces / Exceptions / Constants / Functions / etc listed in the current module (As an easy way to navigate to related objects). 
3/ The module index 
Those 3 indexes could be in 3 ajustable boxes (like JavaDoc, again). 

I think this feature would greatly increase `pydoctor` usability and adoption.

tristanlatr:
Here a really simple Frame view I got with

```
<frameset cols="30%,70%" title="Documentation frame">
    <frameset rows="30%,70%" title="Left frames">
    <frame src="moduleIndex.html" name="moduleIndex" title="All modules">
    <frame src="classIndex.html" name="classFrame" title="All classes">
    </frameset>
    <frame src="index.html" name="indexFrame" title="Index" scrolling="yes">
</frameset>
``` 

![by default 2020-11-06 at 9 26 37 PM](https://user-images.githubusercontent.com/19967168/98429775-c21f6800-2076-11eb-9a5a-e45f999d84e5.png)

It's only to show the type of thing that can be done. The module index and the class index files should not be used in the frames, but some other specially generated files, also it should not show the docstrings. The class list should only show classes in the current module (everything in `index.html`). The headers should be removed and the titles made smaller. The table of contents is not there, maybe it could be added on the top-right corner or another frame. 

I guess we'de have to edit the templates: `common.html` and `index.html`. 

My question are: Where do I start ? I'm not very familiar with introspection. Does a custom HTMLWriter is sufficient to do the job ?

mthuurne:
[...]

Issue #121, opened 2016-07-26
Parameter type information visually hard to find


DanielSank:
Consider the documentation found at (screenshot below) https://twistedmatrix.com/documents/current/api/twisted._threads.IWorker.html

At present, the type information is set off
- By the string "type"
- At the end of the parameter description
- In parentheses

Parameter types are essential information. Consider
- Putting the type a the beginning of the parameter description
- Omitted the unnecessary word "type"
- Giving the parameters their own colored box

![pydoc_note](https://cloud.githubusercontent.com/assets/4584456/17146292/9239a640-52f9-11e6-8519-34ef7fc5609b.png)

mthuurne:
I agree that having the type in parentheses at the end is not great. I'm considering to either put the type immediately after the name, similar to PEP 484 annotations (`task: Callable`), or put it in a separate column.

Types can get quite long, for example the full PEP 484 syntax would be `task: Callable[[], None]` (a 0-argument callable that returns nothing). Type information becomes even longer when non-builtin types are used as type arguments, for example a list of workers would be `List[twisted._threads.IWorker]`. Therefore it might be better to have the documentation body on a separate line, below the name + type. It would then look like a definition list rather than a table.

tristanlatr:
I think the parameters types should be right after the param name, in the same "column", like https://geopy.readthedocs.io/en/stable/#nominatim

tristanlatr:
> Therefore it might be better to have the documentation body on a separate line, below the name + type. It would then look like a definition list rather than a table.

This is not exactly what I did, I stuck with the common way to show arguments types: In parenthesis on the same column as the arg name. 
I fix the size of the argument name column to 25%. 
To handle the longer type names, I've add potential line breaks after `[` or `, ` in type rendering: 
<img width="701" alt="by default 2021-01-14 at 10 33 28 AM" src="https://user-images.githubusercontent.com/19967168/104612208-f35a7c80-5653-11eb-8a76-4cb37b3e5c96.png">

tristanlatr:
after second try, it looks like this (on a small screen): 

![Capture d’écran, le 2021-01-30 à 11 14 43](https://user-images.githubusercontent.com/19967168/106361545-88b06e80-62ec-11eb-9b74-72b585ebc291.png)

mthuurne:
I think that latest version is a big improvement over the current implementation.

[...]

PR #266, opened 2020-11-08
CSS: Reserve the use of the red color for links

tristanlatr:
Changes styles:

- functionHeader: bold and black
- fieldArg: black. Add a arrow before args
- margins: a little less, for more readability
- Also add a 35px margins to showPrivate toggle to able to click on "API Documentation" link at the bottom
- Make function's (source) links not bold
- merge the Sphinx build to be able to review the docs

Why?
Because the red color is usually for links and it needed some other CSS to mark function signature specials


adiroiban:
Thanks for the update. It looks good. but first we need to merge #254 

I am happy that you used the code from #254 and I can review it via https://pydoctor--266.org.readthedocs.build/en/266/api/pydoctor.epydoc2stan.html 

So, if you have time, please send your feedback / review for PR #254 so that we can have it accepted.

Thanks again!

tristanlatr:
The best would be to have the functions `def` lines generated with some kind of syntax highlight, like the class signature...

mthuurne:
The `class` and `def` code fragments look much better in black.

I'm not a fan of the italics in the functions and variables definitions. I think they stand out enough already, making them italic as well just makes them harder to read. Also it means that items that were already italic such as "Undocumented" and "overrides" now stand out less.

For formatting the parameters, I think it's better to do a bigger overhaul in a separate PR in which we address #121.

To come back to the core change: instead of patching specific cases, maybe we should reconsider how we want to format code fragments in general. We have the following types of code fragments: (let me know if I missed any)
* definitions: `class X(Y):`, `def func():`, `var =`
* parameter names
* hyperlinked package/module, function, variable and type names
* inline docstring content marked up as code, for example `C{foo}` in epytext
* docstring content blocks, such as doctests

The current styling for code fragments consists of:
* font: monospaced at 90% size
* text color: red
* background: pink rounded rectangle

I think the font should be applied to all code fragments, since monospace is a standard way of hinting that something is code. Additionally, monospace fonts are often less ambiguous when it comes to for example "l vs I" (`l vs I`).

For the red text color, it may be best to reserve that for hyperlinked code. Having to hover over a fragment to know whether it is clickable is not a good user experience.

Regarding the background, I don't like the rectangles in the summary tables: when the background is already colored, having another colored rectangle on top of it makes it look cluttered. Maybe this could be fixed by tweaking the color of the code rectangle, but I think it's fine to just omit it, since most code fragments in summary tables are hyperlinked and therefore stand out already.

If we only make hyperlinked code fragments red, then no-link inline code fragments don't stand out very well. So having some kind of background on them is useful. For example, GitHub puts `code` on a light gray background. I assume the pink color was chosen to match the red text, so it would make sense to use gray when the text is black.

For inline code fragments that are hyperlinked, we could either keep the pink background, have a gray background or have no background. I've been playing around with "Inspect Element" and I think I prefer no background, but I'd have to see it in the full context of all the other changes to be sure.

> The best would be to have the functions `def` lines generated with some kind of syntax highlight, like the class signature...

Yes, it would be best to use the same CSS classes for definition lines as we use for syntax highlighting, since they are mimicking Python code.

tristanlatr:
> I'm not a fan of the italics in the functions and variables definitions. I think they stand out enough already, making them italic as well just makes them harder to read. Also it means that items that were already italic such as "Undocumented" and "overrides" now stand out less.

No extra italics, I agree

> For example, GitHub puts code on a light gray background. I assume the pink color was chosen to match the red text, so it would make sense to use gray when the text is black.

Let's do as github no?

> For the red text color, it may be best to reserve that for hyperlinked code. Having to hover over a fragment to know whether it is clickable is not a good user experience.

I think it's addressed in the latests commits

This is now the 
The current styling for code fragments consists of:
font: monospaced at 90% size
text color: black
background: light gray

The current styling for links with code fragments consists of:
font: monospaced at 90% size
text color: red
background: pink rounded rectangle

+ I've added a border around a link to index.html that was transparent when using project url. It's not pretty but it works, It needs some other work, to address #121 and #257 anyway

:)

pydoctor/templates/apidocs.css
```diff
@@ -18,7 +18,7 @@
 .categoryHeader {
     font-size: 24px;
     color: #777;
-    margin-bottom: 1.8em;
+    margin-bottom: 0.5em;
```
mthuurne, line 21: I like having a bit more margin between different parts of the document. Rather than reducing the margin here, we could increase margins elsewhere.

This is subjective of course, but in any case this change is not necessary to improve the code fragments layout.

pydoctor/templates/apidocs.css
```diff
@@ -108,6 +108,7 @@ ul ul ul ul ul ul ul {
     padding: 5px;
 }
 
+
```
mthuurne, line 111: This extra empty line isn't consistent with the surrounding rules (a leftover of reverted edits?).

pydoctor/templates/apidocs.css
```diff
@@ -157,8 +158,14 @@ tr.private {
 }
 
 .fieldArg {
-    font-weight: bold;
-    color: #c7254e;
+    color: #222222;
```
mthuurne, line 161: Do we need to override the text color here? The standard text color set by bootstrap is `#333333`, which doesn't look all that different from `#222222`.

adiroiban:
There is a strange red border ... looks like an error

https://pydoctor--266.org.readthedocs.build/en/266/api/pydoctor.epydoc2stan.html

image

pydoctor/templates/apidocs.css
```diff
@@ -157,8 +158,14 @@ tr.private {
...
+}
+
+/* Add an arrow before fields */
```
mthuurne, line 164: I prefer to handle the field layout in a separate PR, since we'll be looking at moving the type information to a different place as well.

pydoctor/templates/apidocs.css
```diff
@@ -195,30 +202,90 @@ tr.private {
     margin-bottom: 6px;
 }
 
+/* This is magic to show the transparent index link */
```
mthuurne, line 205: I think it's useful to have a link to the index somewhere, but this is only half a solution, so I'd prefer to solve it completely in a separate PR.

pydoctor/templates/apidocs.css
```diff
@@ -195,30 +202,90 @@ tr.private {
...
+
+/* .page-header > h1 > code > a {
+    background-color:#FCF9FA;
+} */
+
+/* 
+<a><code></code><a> for links
+This is inline docstring content marked up as code, for example C{foo} in epytext
+*/
+a > code {
```
mthuurne, line 230:
```suggestion
a[href] > code {
```
The `<a>` tag is also used to define anchors and that should not cause the text to become red.

pydoctor/templates/apidocs.css
```diff
@@ -195,30 +202,90 @@ tr.private {
...
+    background-color:#f9f2f4;
+    border-radius: 4px;
+}
+
+/* This is for the main top navagation bar */
+.page-header > h1 > code {
+    background-color:#f9f2f4;
+    /* background-color:transparent; */
+}
+
```
mthuurne, line 221: There is some commented out code here (above and below) that can be cleaned up.

pydoctor/templates/apidocs.css
```diff
@@ -195,30 +202,90 @@ tr.private {
...
+    margin-right: 0.3em;
+    border-width: 1.5px;
+    border-color: #c7254e;
+    background-color:#f9f2f4;
+    border-radius: 4px;
+}
+
+/* This is for the main top navagation bar */
+.page-header > h1 > code {
+    background-color:#f9f2f4;
```
mthuurne, line 218: Maybe we should attach the color to the `<h1>` tag rather than the `<code>` tag. Then we could later use the `package`/`module`/`class` styles that are currently applied to `<tr>` also on headers.

pydoctor/templates/apidocs.css
mthuurne: Where do we use this rule? I'd expect most HTML would have the `<a href="">` inside a `<code>` tag and not vice versa.

pydoctor/templates/apidocs.css
```diff
@@ -195,30 +202,90 @@ tr.private {
...
+*/
+a > code {
+    color:#c7254e; 
+    background-color:#f9f2f4;
+    /* background-color:transparent; */
+}
+/* 
+<a class="code"> links that are present in summary tables 
+*/
+a.code {
```
mthuurne, line 238: If we use the same styling in multiple places, we can use the `a, b { ... }` selector syntax to avoid duplicating the rules.

pydoctor/templates/apidocs.css
```diff
@@ -195,30 +202,90 @@ tr.private {
...
+    /* color:#c7254e; 
+    background-color:#FCF9FA;*/
     border-radius:4px
 }
+/* 
+This overwrites the default values inherited from bootstrap min
+*/
+code {
+    color: #222222;
+    background-color: #f9f2f4;
```
mthuurne, line 261: I think it would look better if we used pink background for red text and gray background for near-black text.

pydoctor/templates/apidocs.css
```diff
@@ -195,30 +202,90 @@ tr.private {
...
 
 code > .code {
     padding: 0px;
 }
 
+
+
+a.functionSourceLink {
+    font-weight: normal;
+}
```
mthuurne, line 272: This is good for now. Later we should remove the bold from `functionHeader` and instead style the function definition line like syntax highlighted code.

pydoctor/templates/apidocs.css
```diff
@@ -195,30 +202,90 @@ tr.private {
...
+
+
 #childList > div {
     border-left-color: #03a9f4;
     border-left-width: 1px;
     border-left-style: solid;
-    background: #fafafa;
+    background: #fbfbfb;
```
mthuurne, line 279: That's a very subtle change ;)

pydoctor/templates/apidocs.css
```diff
@@ -250,6 +317,7 @@ body.private-hidden #splitTables .private, body.private-hidden #childList .priva
     left: 0;
     padding: 10px;
     opacity: 0.7;
+    margin-bottom: 35px;
```
mthuurne, line 320: This seems unrelated to the other changes.

mthuurne (changes requested):
To summarize:
- overall this looks much better :)
- I would like changes that are unrelated to styling code fragments to be moved to separate PRs
- the CSS still has several leftovers from experiments that need to be cleaned up before we can merge

mthuurne:
By the way, did you try disabling the background rectangle for code in the summary tables? If I do that using the browser's dev mode, I think the result is easier on the eyes.

pydoctor/templates/apidocs.css
tristanlatr: With `.page-header > h1 > * {` instead of `.page-header > h1 > code {` it works

pydoctor/templates/apidocs.css
tristanlatr: > Where do we use this rule?

Here for instance: file:///Users/tristan/Documents/pydoctor/build/apidocs/pydoctor.epydoc.markup.ParsedDocstring.html

The `Subclasses must implement L{to_stan()}.`

pydoctor/templates/apidocs.css
mthuurne: Hmm, not sure that's an improvement. I was thinking of using just `.page-header > h1 {`, but it seems quite a lot of CSS needs to change for that to work.

Also the `<h1>` spans the entire body width instead of just the title, so it would look differently.

Let's leave it as- is, for now.

pydoctor/templates/apidocs.css
mthuurne: Hmm, we could change the code that outputs this in `pydoctor.epydoc.markup.*`, but I'm not sure whether that would make it easier or harder to write the CSS.

tristanlatr:
> There is a strange red border ... looks like an error

That was't an error, but a link to the index

> By the way, did you try disabling the background rectangle for code in the summary tables? If I do that using the browser's dev mode, I think the result is easier on the eyes.

Yes, it was the `/* background-color:transparent; */`. I find it cool too. 

The requested changes are done.

mthuurne:
Thanks!

The command like `git` client shows some redundant spaces or newline characters at the end of some changed lines, so I added a commit to remove those.

PR #273, opened 2020-11-13
Don't crash if the line number for a docstring error is unknown

### LINKED ISSUE #261: TypeError: unsupported operand type(s) for -: 'NoneType' and 'int'

mthuurne:
Fixes #261.

This is another case where mypy could have spotted the problem if we'd had enough annotation coverage.

### DIFF
--- a/pydoctor/epydoc/markup/__init__.py
+++ b/pydoctor/epydoc/markup/__init__.py
@@ -205,73 +205,66 @@ class DocstringLinker:
 class ParseError(Exception):
     """
     The base class for errors generated while parsing docstrings.
-
-    @ivar _linenum: The line on which the error occured within the
-        docstring.  The linenum of the first line is 0.
-    @type _linenum: C{int}
-    @ivar _descr: A description of the error.
-    @type _descr: C{string}
-    @ivar _fatal: True if this is a fatal error.
-    @type _fatal: C{bool}
     """
-    def __init__(self, descr, linenum=None, is_fatal=True):
+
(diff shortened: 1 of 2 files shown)


adiroiban (approved):
It looks good. Thanks.

It would have been nice to have a new tests... but I hope that by running the cpython system tests we will have some sort of regression coverage.

So I don't think that we should delay this merge.

PR #274, opened 2020-11-13
Annotate more

mthuurne:
Since we're running into bugs that mypy could have spotted given enough annotations, I decided to add a bunch of annotations.

So far I have found:
- one broken feature (`--abbreviate-specialcase`), that I removed since no-one reported it was broken for 5 years
- one "don't use the network" precaution in a test case that didn't work because it patched a removed method

This PR contains the changes from #262 and #273, so it should be merged after those two.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,6 +1,7 @@
 [mypy]
 disallow_any_generics=True
 disallow_incomplete_defs=True
+disallow_untyped_defs=True
 namespace_packages=True
 no_implicit_optional=True
 show_error_codes=True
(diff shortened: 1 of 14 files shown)


adiroiban (approved):
It looks good. Thanks!

Happy to see the TemplateWriter private method rename :)

adiroiban:
I see that the slowest test is pyp3 - 2min 30 seconds... and then there is also Windows tests at 2min 

So, I guess that as long as the cpython doc build tests takes less than 2.5 minutes we are ok.

mthuurne:
Thanks for the quick review!

> I see that the slowest test is pyp3 - 2min 30 seconds... and then there is also Windows tests at 2min
> 
> So, I guess that as long as the cpython doc build tests takes less than 2.5 minutes we are ok.

The `cpython-summary` env is now run as part of the system test. That one parses all of CPython but only generates a handful of HTML pages, instead of one per module and class. If the CI runners aren't super slow, it should stay well under a minute. I think this gives us a good trade-off between run time and chance of discovering regressions.

The `cpython-apidocs` env is there for manual testing, which takes several minutes to run to completion. It does reach completion now though.

PR #275, opened 2020-11-14
Move anchors up to compensate for sticky navbar

### LINKED ISSUE #270: Safari : The sticky navbar hides the function definition focused with "#" in URL
Here: https://mfesiem.github.io/docs/test/msiempy/msiempy.AlarmManager.html#page_size for instance, you won't actually see the definition of the `page_size` property as it's behind de navbar.

The solution would be to apply a margin somewhere but not sure where.

Also that would be good that the element focused with "#" in URL appear highlighted so we know where to look for it.

mthuurne:
When navigating to an anchor, browsers will typically put it at the top of the page, but they don't take the sticky navbar into account, so the thing the user was looking for ends up covered by the navbar.

Since anchors are invisible, we can move their location without problems. However, they don't have an actual location by default. But if we provide invisible content, they do, and the trick works.

Fixes #270.

### DIFF
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -23,7 +23,7 @@
 
 a[name] {
     position: relative;
-    bottom: 70px;
+    bottom: 62px;
 }
 
 ul {
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -21,6 +21,14 @@
     margin-bottom: 1.8em;
 }
 
+a[name] {
+    position: relative;
+    bottom: 70px;
+}
+a[name]:before {
+    content: "";
+}
+
 ul {
     margin-top: 10px;
     padding-left: 10px;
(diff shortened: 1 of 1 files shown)


pydoctor/templates/apidocs.css
```diff
@@ -21,6 +21,14 @@
     margin-bottom: 1.8em;
 }
 
+a[name] {
+    position: relative;
+    bottom: 70px;
+}
+a[name]:before {
+    content: "";
```
adiroiban, line 29: Is this required? Just asking.

adiroiban (approved):
Great work.

I have executed the following tests.

Went to "https://pydoctor.readthedocs.io/en/latest/api/pydoctor.model.Documentable.html" and then click on the "parent" instance variable.

I can see that on master the "parent" is covered... but on this branch it looks ok

Link for master
https://pydoctor.readthedocs.io/en/latest/api/pydoctor.model.Documentable.html#parent

Link for PR
https://pydoctor--275.org.readthedocs.build/en/275/api/pydoctor.model.Documentable.html#parent

pydoctor/templates/apidocs.css
mthuurne: It was probably required back when I found this workaround, but indeed when I test it now, it doesn't seem to make a difference. I'll remove it.

PR #276, opened 2020-11-14
Convert --project-base-dir argument to absolute path

mthuurne:
The `Path.relative_to()` method used in `System.setSourceHref()` will only work on absolute paths.

This fixes a crash when passing for example `--project-base-dir .` or `--project-base-dir src`.

### DIFF
--- a/pydoctor/test/test_commandline.py
+++ b/pydoctor/test/test_commandline.py
@@ -1,6 +1,7 @@
 from contextlib import redirect_stdout
 from io import StringIO
 from pathlib import Path
+import os
 import sys
 
 from pydoctor import driver
@@ -54,7 +55,10 @@ def test_projectbasedir() -> None:
     on the options object, as an absolute path.
     """
 
-    absolute = "/home/name/src/project"
+    if os.name == 'nt':
+        absolute = r"C:\Users\name\src\project"
+    else:
+        absolute = "/home/name/src/project"
     options, args = driver.parse_args(["--project-base-dir", absolute])
     assert str(options.projectbasedirectory) == absolute
 
--- a/pydoctor/test/test_commandline.py
+++ b/pydoctor/test/test_commandline.py
@@ -1,5 +1,6 @@
 from contextlib import redirect_stdout
 from io import StringIO
+from pathlib import Path
 import sys
 
 from pydoctor import driver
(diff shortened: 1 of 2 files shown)


pydoctor/test/test_commandline.py
```diff
@@ -50,12 +52,21 @@ def test_invalid_systemclasses() -> None:
...
+    else:
+        absolute = "/home/name/src/project"
+    options, args = driver.parse_args(["--project-base-dir", absolute])
+    assert str(options.projectbasedirectory) == absolute
+
+    relative = "projbasedirvalue"
+    options, args = driver.parse_args(["--project-base-dir", relative])
+    assert options.projectbasedirectory.is_absolute()
+    assert options.projectbasedirectory.name == relative
+    assert options.projectbasedirectory.parent == Path.cwd()
```
adiroiban, line 69:
```suggestion
    if os.name == 'nt':
        absolute = r"C:\Users\name\src\project"
    else:
        absolute = "/home/name/src/project"

    options, args = driver.parse_args(["--project-base-dir", absolute])
    assert str(options.projectbasedirectory) == absolute
    assert options.projectbasedirectory.is_absolute()

def test_projectbasedir_relative() -> None:
    """
    When --project-base-dir is passed as a relative path,
    the projectbasedirectory is set as an absolute path.
    """
    options, args = driver.parse_args(["--project-base-dir=./relative/base"])

    assert options.projectbasedirectory.is_absolute()
    assert options.projectbasedirectory.name == 'base'
    # The absolute path will differ based on project location, but we
    # can check that it's related to the current working directory.
    assert options.projectbasedirectory.parent.parent == Path.cwd()
```

mthuurne: Done.

adiroiban (approved):
Thanks. All good.

The Path.resolve() difference between Win and Unix is a feature or a bug?

mthuurne:
> The Path.resolve() difference between Win and Unix is a feature or a bug?

I feel that the Windows behavior is either a bug or at least poorly documented, since [it states](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve):

> Make the path absolute,
> ...
> If *strict* is `False`, the path is resolved as far as possible and any remainder is appended without checking whether it exists.

But if the first segment of a relative path doesn't exist, the Windows version will append the remainder to the empty path (resulting in a path that is **not** absolute), instead of appending it to the working directory as Linux does.

There is already a [bug report about improving the documentation](https://bugs.python.org/issue39090) though, as well as one asking for [the undocumented `Path.absolute()` method to be made official](https://bugs.python.org/issue29688), so I think the maintainers are aware of the issues. But it also feels like the existence of lots of nasty edge cases is preventing the changes needed to make a common use case work as expected.

Issue #268, opened 2020-11-09
Param types should be clickable


tristanlatr:
Whether the type is retrieved with

```
:type arg: str
```
or from type hints
```
def get(id:str):
    pass
```

It's not clickable, even if it appears in red when figured from type hints

mthuurne:
I was already working on hyperlinked type annotations :) But as usual, I ran into pre-existing issues (see #250) and then got pulled into other things, so it's not PR-ready yet.

For non-annotation types, we format them just like other field values. For epytext, the `type` field is just a text field, it doesn't have to contain the name of a class or built-in type. What are the semantics for reStructuredText?

Twisted uses explicit links when the type is a class or built-in type:
```
@type arg: L{str}
```

tristanlatr:
With ReST, 
```
:type arg: str
```
The type value should be interpreted 

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists
Says:

Container types such as lists and dictionaries can be linked automatically using the following syntax:

:type priorities: list(int)
:type priorities: list[int]
:type mapping: dict(str, int)
:type mapping: dict[str, int]
:type point: tuple(float, float)
:type point: tuple[float, float]
Multiple types in a type field will be linked automatically if separated by the word “or”:

:type an_arg: int or None
:vartype a_var: str or int
:rtype: float or str

tristanlatr:
> The type value should be interpreted

I have this issue fixed with Google and Numpy docstring parsing, see [code](https://github.com/tristanlatr/pydoctor/blob/42fe0921cfd4dc0a0719223a5bbdf441af42bc39/pydoctor/napoleon/docstring.py#L959). but it should probably be moved to epydoc2stan and adapted to be a ParsedDocstring subclass

PR #277, opened 2020-11-14
Handle cyclic imports better

mthuurne:
When working on adding hyperlinks to presented annotations (#268), I noticed that several names did not become links. After some debugging, it turned out that because of circular imports, the module being imported from (`apetest.report`) wasn't fully processed yet, which led to `expandName('Report')` returning just `'Report'` instead of `apetest.report.Report` (the full name).

I wrote a fix for this particular case, which is in the first commit. It checks whether the module being imported from is already processed, and if not, it will synthesize the full name by combining the module name and the imported name (`apetest.report` and `Report`) instead of using the `expandName()` method.

While this fix works, I don't like it: in this case, the synthetic name is the same as the name we would get if the module had already been processed, but in some projects that might not always be the case. In that situation, the output that pydoctor produces can be different depending on the order in which modules are processed, which is unpredictable and can vary between runs (maybe because of the VM's hash seed, maybe because of the underlying file system).

So in the second commit, I removed the use of the `expandName()` method from the `from <somewhere> import <names>` parsing in the `_importNames()` method altogether. This means we always use the synthetic names, which makes the behavior simple and consistent. It does change the behavior of pydoctor when processing chains of aliases though, so I had to update the expected output in two test cases. I don't think this is a problem, as explained in that second commit:

> Another motivation for removing the feature is that I have some doubt about its usefulness: importing a name that is itself imported from another module is bad practice, except when the name is explicitly re-exported via `__all__`. But in the case of a re-export, the documentation will be using the re-export location, so we don't need to trace it back to the original implementation location.

We can go one step further after this simplification and also move the re-exporting code out of the `_importNames()` method and instead only record re-exported names there and then do the reparenting as a separate step after all modules have been processed. That would simplify the implementation of `_importNames()` considerably and it would remove the need for triggering additional module processing via `getProcessedModule()` from there.

Since this is complex behavior, I didn't want to do too much in this PR though, so I stuck to the changes that we need to be able to reliably hyperlink names in annotations.

### DIFF
--- a/pydoctor/test/test_astbuilder.py
+++ b/pydoctor/test/test_astbuilder.py
@@ -253,7 +253,15 @@ def test_aliasing(systemcls: Type[model.System]) -> None:
     addsrc(system)
     C = system.allobjects['c.C']
     assert isinstance(C, model.Class)
-    assert C.bases == ['a.A']
+    # An older version of this test expected a.A as the result.
+    # The expected behavior was changed because:
+    # - relying on on-demand processing of other modules is unreliable when
+    #   there are cyclic imports: expandName() on a module that is still being
+    #   processed can return the not-found result for a name that does exist
+    # - code should be importing names from their official home, so if we
+    #   import b.B then for the purposes of documentation b.B is the name
+    #   we should use
+    assert C.bases == ['b.B']
 
 @systemcls_param
 def test_more_aliasing(systemcls: Type[model.System]) -> None:
@@ -282,7 +290,9 @@ def test_more_aliasing(systemcls: Type[model.System]) -> None:
     addsrc(system)
     D = system.allobjects['d.D']
     assert isinstance(D, model.Class)
-    assert D.bases == ['a.A']
+    # An older version of this test expected a.A as the result.
+    # Read the comment in test_aliasing() to learn why this was changed.
+    assert D.bases == ['c.C']
 
 @systemcls_param
 def test_aliasing_recursion(systemcls: Type[model.System]) -> None:
--- a/pydoctor/test/test_astbuilder.py
+++ b/pydoctor/test/test_astbuilder.py
@@ -233,35 +233,36 @@ def test_class_with_base_from_module(systemcls: Type[model.System]) -> None:
 @systemcls_param
 def test_aliasing(systemcls: Type[model.System]) -> None:
     def addsrc(system: model.System) -> None:
-        src_a = '''
+        src_private = '''
         class A:
             pass
         '''
-        src_b = '''
-        from a import A as B
+        src_export = '''
+        from _private import A as B
+        __all__ = ['B']
         '''
-        src_c = '''
-        from b import B
+        src_user = '''
+        from public import B
         class C(B):
             pass
         '''
-        fromText(src_a, modname='a', system=system)
-        fromText(src_b, modname='b', system=system)
-        fromText(src_c, modname='c', system=system)
+        fromText(src_private, modname='_private', system=system)
+        fromText(src_export, modname='public', system=system)
+        fromText(src_user, modname='app', system=system)
 
     system = systemcls()
     addsrc(system)
-    C = system.allobjects['c.C']
+    C = system.allobjects['app.C']
     assert isinstance(C, model.Class)
-    # An older version of this test expected a.A as the result.
+    # An older version of this test expected _private.A as the result.
     # The expected behavior was changed because:
     # - relying on on-demand processing of other modules is unreliable when
     #   there are cyclic imports: expandName() on a module that is still being
     #   processed can return the not-found result for a name that does exist
     # - code should be importing names from their official home, so if we
-    #   import b.B then for the purposes of documentation b.B is the name
-    #   we should use
-    assert C.bases == ['b.B']
+    #   import public.B then for the purposes of documentation public.B is
+    #   the name we should use
+    assert C.bases == ['public.B']
 
 @systemcls_param
 def test_more_aliasing(systemcls: Type[model.System]) -> None:
--- a/pydoctor/test/test_packages.py
+++ b/pydoctor/test/test_packages.py
@@ -70,3 +70,10 @@ def test_allgames() -> None:
     assert moved.parentMod is mod2
     assert moved.parentMod.source_path is not None
     assert moved.parentMod.source_path.parts[-2:] == ('allgames', 'mod2.py')
+
+def test_cyclic_imports() -> None:
+    system = processPackage('cyclic_imports')
+    mod_a = system.allobjects['cyclic_imports.a']
+    assert mod_a.expandName('B') == 'cyclic_imports.b.B'
+    mod_b = system.allobjects['cyclic_imports.b']
+    assert mod_b.expandName('A') == 'cyclic_imports.a.A'
--- a/pydoctor/test/test_packages.py
+++ b/pydoctor/test/test_packages.py
@@ -72,6 +72,14 @@ def test_allgames() -> None:
     assert moved.parentMod.source_path.parts[-2:] == ('allgames', 'mod2.py')
 
 def test_cyclic_imports() -> None:
+    """
+    Test whether names are resolved correctly when we have import cycles.
+    The test package contains module 'a' that defines class 'A' and module 'b'
+    that defines class 'B'; each module imports the other. Since the test data
+    is symmetrical, we will at some point be importing a module that has not
+    been fully processed yet, no matter which module gets processed first.
+    """
+
     system = processPackage('cyclic_imports')
     mod_a = system.allobjects['cyclic_imports.a']
     assert mod_a.expandName('B') == 'cyclic_imports.b.B'
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -209,14 +209,6 @@ class ModuleVistor(ast.NodeVisitor):
         is_package = isinstance(obj, model.Package)
         assert is_package or obj is mod or mod is None
 
-        if mod is not None and mod.state is model.ProcessingState.PROCESSED:
-            assert obj is not None
-            expandName = obj.expandName
-        else:
-            # Module information is absent or incomplete;
-            # treat it as a black box.
-            expandName = lambda name: f'{modname}.{name}'
-
         # Fetch names to export.
         current = self.builder.current
         if isinstance(current, model.Module):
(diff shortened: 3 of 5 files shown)


mthuurne:
Some new warnings are introduced when processing Twisted with pydoctor from this PR:
```
src/twisted/mail/pop3client.py:614: ambiguous ref to TLSError, could be twisted.mail._except.TLSError, twisted.mail.smtp.TLSError
src/twisted/mail/pop3client.py:614: ambiguous ref to TLSError, could be twisted.mail._except.TLSError, twisted.mail.smtp.TLSError, twisted.words.protocols.jabber.xmlstream.TLSError
src/twisted/mail/pop3client.py:614: invalid ref to 'TLSError' resolved as 'twisted.mail.pop3client.TLSError'
src/twisted/python/filepath.py:227: invalid ref to '_OSErrorOrFakeWindowsError' resolved as 'twisted.python.win32.WindowsError'
```
The first three warnings all belong to the same cross-reference. The problem seems to be that there are 3 `TLSError` classes in Twisted and pydoctor considers them all equally good candidates. That shouldn't be the case, as resolving the name like Python does should yield `twisted.mail.pop3client.TLSError` as the obvious best candidate. Maybe the problem is that for some reason `POP3Client` is re-exported as `twisted.mail.pop3.AdvancedPOP3Client` while `TLSError` is not re-exported.

We can't do "look up this name like Python does" anymore after models have been reparented. There once were similar problems with the source links and reparenting, which was fixed by giving every `Documentable` a `source_path` attribute instead of only modules. We could do something similar here, keeping both the current and source parent around. Another option would be to reconsider whether implementing re-exports using reparenting in the model a good idea; maybe we should leave the analysis results in the source locations and then build a separate hierarchy for the output locations.

Regarding the fourth warning, `_OSErrorOrFakeWindowsError` is either an alias for `os.WindowsError` or `twisted.python.win32.FakeWindowsError`. The old pydoctor code would pick the latter, while the new code refuses to pick since it won't resolve aliases across modules. Maybe we need a `Documentable` subclass for public aliases, so that we have something to link to in cases like this.

By the way, I think that documenting a private exception name in a public method isn't correct. There is also no need for this: both the real and fake `WindowsError` inherit from `OSError`, so the docstring could just say that the original exception is an `OSError`. In fact, since Python 3.3, `os.WindowsError` is actually an alias for `OSError`.

mthuurne:
The problem with cyclic imports is worse than I thought: for example `addInterfaceInfoToScope()` looks up the interface using `System.objForFullName()` and ignores it if not found. So if the interface's module hasn't been processed yet, the interface is not recorded as being implemented. Because `astbuilder` tries to process every imported module, this is usually not a problem, but in the case of cyclic imports, it's possible for an imported module to not have been fully processed yet when the importer is processed.

I think the safest thing to do would be to not have `objForFullName()` as part of `System`, but instead to build an index of all names after `System` has finished processing. However, that's a design change that's going to touch a lot of code.

adiroiban:
Is this for review or still a draft/in work?
Thanks

pydoctor/test/test_packages.py
```diff
@@ -70,3 +70,10 @@ def test_allgames() -> None:
     assert moved.parentMod is mod2
     assert moved.parentMod.source_path is not None
     assert moved.parentMod.source_path.parts[-2:] == ('allgames', 'mod2.py')
+
+def test_cyclic_imports() -> None:
+    system = processPackage('cyclic_imports')
```
adiroiban, line 75: I would like to see a docstring with the target / scope / purpose of this test :)

pydoctor/test/test_astbuilder.py
```diff
@@ -282,7 +290,9 @@ class D(C):
     addsrc(system)
     D = system.allobjects['d.D']
     assert isinstance(D, model.Class)
-    assert D.bases == ['a.A']
+    # An older version of this test expected a.A as the result.
+    # Read the comment in test_aliasing() to learn why this was changed.
+    assert D.bases == ['c.C']
```
adiroiban, line 295: To make it easier to understand what we have here, maybe use `_private_stuff` instead of just `a`.

Also, instead of `c` use `public_stuff`.

And instead of `b` use "_private_agregator` or something like that.

I hope that in this way it will make it easier to see that we don't want to expose a, since a is a private module

adiroiban (approved):
I think that this is an improvement over master and it should be merged.

I would prefer that instead of `a, b, and c` to have names which add a bit more info about the purpose of each test element...but that is not critical and should not block the merge.

Thanks! Great work!

mthuurne:
> Is this for review or still a draft/in work?

I originally put it up for review, then noticed that it introduced new warnings when processing Twisted. The `WindowsError` warning can be fixed by tidying up the Twisted code (https://github.com/twisted/twisted/pull/1479), but the warnings about `TLSError` will have to be addressed in pydoctor.

I tried addressing the warnings by cleaning up the pydoctor code further, but that only introduces more warnings. It seems that pydoctor is currently in a local maximum, where cyclic imports do cause problems, but very rarely. Partial solutions tend to make the symptoms worse, even though the code is closer to correct.

I'll try to find a fix for the `TLSError` warnings without changing too much in the processing order. Another option could be to merge only the first commit from this PR: although it has downsides in being unpredictable (output can differ depending on processing order), that is not a regression since this is a problem that has existed in pydoctor from the start.

pydoctor/test/test_packages.py
mthuurne: I'll add it.

mthuurne:
The situation with `TLSError` seems to be like this:
- `twisted.mail.pop3` imports `TLSError` from `twisted.mail.pop3client` and exports it via `__all__`.
- `twisted.mail.pop3client` imports `TLSError` from `twisted.mail._except` and has an empty `__all__`.
- `twisted.mail._except` contains the actual class definition for `TLSError` and has no `__all__`.

So we have one implementation module importing things from another implementation module.

If this is something we want to support, we have to drop commit 6982f81. Otherwise, we should modify Twisted to have `twisted.mail.pop3` import `TLSError` directly from `twisted.mail._except`.

While importing directly from `_except` would make things easier for pydoctor, there is something to be said for the internal details of an implementation being hidden from the outside world by a top-level implementation module. So I think the Twisted code is fine and therefore we should drop the "Remove tracing of imported names across module boundaries" change.

As for a long term solution, I'm starting to think that reparenting is a bad idea: it means we lose information about how the code is structured. Instead of altering the code model, we could create layer on top of it that contains the documentation structure. That would also allow us to thin down the interface of the model classes a bit, by moving presentation data out of the code model.

Linking annotations containing imported names would then work as follows:
- look up the original code object in the same way Python does (trace it back recursively to source location)
- look up where the code object ended up in the documentation layer (might not match source location if name was re-exported)
- link to the documentation location

adiroiban:
My understanding that we can/should use the member from `__all__` as the canonical API location

My guess is that Twisted we should have something like this:

Private module implementation:

* `twisted.mail._except` contains the actual class definition for TLSError and has no `__all__`.

Public export of the implementation

* A: `twisted.mail.pop3client` imports `TLSError` from `twisted.mail._except` and exports it via `__all__`
* B: `twisted.mail.pop3` imports `TLSError` from `twisted.mail._except` and exports it via `__all__`.

Public usage:

* A: `twisted.mail.pop3` imports `TLSError` from `twisted.mail.pop3client` and **does not** exports it via `__all__`
* B: `twisted.mail.pop3client` imports `TLSError` from `twisted.mail.pop3` and **does not** exports it via `__all__`.


> As for a long term solution, I'm starting to think that reparenting is a bad idea: it means we lose information about how the code is structured. Instead of altering the code model, we could create layer on top of it that contains the documentation structure. That would also allow us to thin down the interface of the model classes a bit, by moving presentation data out of the code model.

+1 yes. I think that we should always have a layer with the true structure.. and with that we can have a public API layer which does it best to remove duplicates and present the doc links using "canonical" paths.
 

-----------

pydoctor was created to help twised documentation...but if we see that some usage patterns don't make sense in Twisted is ok to change the Twisted to make it easier to document and present the canonical API

mthuurne:
> My understanding that we can/should use the member from `__all__` as the canonical API location

Yes, the module exporting a name via `__all__` is the canonical location for both documentation and for 3rd party code to import when its wants to use the name.

For keeping track of implementations, pydoctor should be using the original name from the definition in its original source file as the canonical location. But that happens only internally; it is not visible to the reader of the documentation.

> My guess is that Twisted we should have something like this:
> 
> Private module implementation:
> * `twisted.mail._except` contains the actual class definition for TLSError and has no `__all__`.
> 
> Public export of the implementation
> * A: `twisted.mail.pop3client` imports `TLSError` from `twisted.mail._except` and exports it via `__all__`
> * B: `twisted.mail.pop3` imports `TLSError` from `twisted.mail._except` and exports it via `__all__`.

`twisted.mail.pop3` tries to export `TLSError` via `__all__`, but pydoctor doesn't move it there for some reason. This is broken even in the older pydoctor versions that were used to generate for example the Twisted 19.2.0 docs, so that problem is unrelated to the modernization efforts. However, it is clear that the intention is to have `TLSError` show up in the documentation as `twisted.mail.pop3.TLSError`. So option B is correct here.

> Public usage:
> * A: `twisted.mail.pop3` imports `TLSError` from `twisted.mail.pop3client` and **does not** exports it via `__all__`
> * B: `twisted.mail.pop3client` imports `TLSError` from `twisted.mail.pop3` and **does not** exports it via `__all__`.

`twisted.mail.pop3client`'s docstring says:
```
A POP3 client protocol implementation.

Don't use this module directly.  Use twisted.mail.pop3 instead.
```
So `pop3client` seems to be intended as a private implementation module, but doesn't comply with the leading underscore naming scheme.

Therefore `twisted.mail.pop3client` is not public usage, so I don't think option B applies here. There would also be practical problems with that option, since `twisted.mail.pop3client` importing anything from `twisted.mail.pop3` would create an import cycle, which could be a problem not just for pydoctor, but also for Python itself.

What would work is:
* C: `twisted.mail.pop3client` imports `TLSError` from `twisted.mail._except` and **does not** export it via `__all__`.

And that is also the current implementation.

> pydoctor was created to help twised documentation...but if we see that some usage patterns don't make sense in Twisted is ok to change the Twisted to make it easier to document and present the canonical API

I agree for cases where such a change does not make the Twisted code worse. I initially thought that having a semi-private module like `twisted.mail.pop3client` has advantages in structuring the code, but the more I think about it, the more I'm convinced that it's not worth the complexity, even when we ignore pydoctor and only think of a developer navigating Twisted's code.

mthuurne:
I filed [a ticket for Twisted](https://twistedmatrix.com/trac/ticket/10054).

PR #278, opened 2020-11-14
Annotation presentation

mthuurne:
Until now, annotations were presented in the output as plain text. This PR changes that by processing the AST for the annotation and linking type names to their internal or external documentation.

This addresses one of the two use cases requested in #268.

The line number preservation changes aren't strictly necessary, since we don't actually report errors when presenting annotations, but having line number information for annotations might be useful in the future.

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -649,7 +649,6 @@ def _annotation_for_value(value: object) -> Optional[ast.expr]:
         return None
     name = type(value).__name__
     if isinstance(value, (dict, list, set, tuple)):
-        name = name.capitalize()
         ann_elem = _annotation_for_elements(value)
         if isinstance(value, dict):
             ann_value = _annotation_for_elements(value.values())
@@ -658,7 +657,7 @@ def _annotation_for_value(value: object) -> Optional[ast.expr]:
             elif ann_elem is not None:
                 ann_elem = ast.Tuple(elts=[ann_elem, ann_value])
         if ann_elem is not None:
-            if name == 'Tuple':
+            if name == 'tuple':
                 ann_elem = ast.Tuple(elts=[ann_elem, ast.Ellipsis()])
             return ast.Subscript(value=ast.Name(id=name),
                                  slice=ast.Index(value=ann_elem))
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -609,14 +609,14 @@ class _AnnotationStringParser(ast.NodeTransformer):
         else:
             # Other subscript; unstring the slice.
             slice = self.visit(node.slice)
-        return ast.Subscript(value, slice, node.ctx)
+        return ast.copy_location(ast.Subscript(value, slice, node.ctx), node)
 
     # For Python >= 3.8:
 
     def visit_Constant(self, node: ast.Constant) -> ast.expr:
         value = node.value
         if isinstance(value, str):
-            return self._parse_string(value)
+            return ast.copy_location(self._parse_string(value), node)
         else:
             const = self.generic_visit(node)
             assert isinstance(const, ast.Constant), const
@@ -625,7 +625,7 @@ class _AnnotationStringParser(ast.NodeTransformer):
     # For Python < 3.8:
 
     def visit_Str(self, node: ast.Str) -> ast.expr:
-        return self._parse_string(node.s)
+        return ast.copy_location(self._parse_string(node.s), node)
 
 def _infer_type(expr: ast.expr) -> Optional[ast.expr]:
     """Infer an expression's type.
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -638,7 +638,11 @@ def _infer_type(expr: ast.expr) -> Optional[ast.expr]:
     except ValueError:
         return None
     else:
-        return _annotation_for_value(value)
+        ann = _annotation_for_value(value)
+        if ann is None:
+            return None
+        else:
+            return ast.fix_missing_locations(ast.copy_location(ann, expr))
 
 def _annotation_for_value(value: object) -> Optional[ast.expr]:
     if value is None:
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -500,7 +500,10 @@ class ModuleVistor(ast.NodeVisitor):
         self.default(node)
         self.builder.popFunction()
 
-    def _annotation_from_attrib(self, expr, ctx):
+    def _annotation_from_attrib(self,
+            expr: ast.expr,
+            ctx: model.Documentable
+            ) -> Optional[ast.expr]:
         """Get the type of an C{attr.ib} definition.
         @param expr: The expression's AST.
         @param ctx: The context in which this expression is evaluated.
@@ -624,7 +627,7 @@ class _AnnotationStringParser(ast.NodeTransformer):
     def visit_Str(self, node: ast.Str) -> ast.expr:
         return self._parse_string(node.s)
 
-def _infer_type(expr):
+def _infer_type(expr: ast.expr) -> Optional[ast.expr]:
     """Infer an expression's type.
     @param expr: The expression's AST.
     @return: A type annotation, or None if the expression has no obvious type.
@@ -637,7 +640,7 @@ def _infer_type(expr):
     else:
         return _annotation_for_value(value)
 
-def _annotation_for_value(value):
+def _annotation_for_value(value: object) -> Optional[ast.expr]:
     if value is None:
         return None
     name = type(value).__name__
@@ -657,7 +660,7 @@ def _annotation_for_value(value):
                                  slice=ast.Index(value=ann_elem))
     return ast.Name(id=name)
 
-def _annotation_for_elements(sequence):
+def _annotation_for_elements(sequence: Iterable[object]) -> Optional[ast.expr]:
     names = set()
     for elem in sequence:
         ann = _annotation_for_value(elem)
--- a/pydoctor/epydoc/markup/__init__.py
+++ b/pydoctor/epydoc/markup/__init__.py
@@ -184,9 +184,22 @@ class DocstringLinker:
     target URL for crossreference links.
     """
 
+    def resolve_identifier(self, identifier: str) -> Optional[str]:
+        """
+        Resolve a Python identifier.
+        This will resolve the identifier like Python itself would.
+
+        @param identifier: The name of the Python identifier that
+            should be linked to.
+        @return: The URL of the target, or L{None} if not found.
+        """
+        raise NotImplementedError()
+
     def resolve_identifier_xref(self, identifier: str, lineno: int) -> str:
         """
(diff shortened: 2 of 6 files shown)


pydoctor/test/test_astbuilder.py
```diff
@@ -71,17 +71,37 @@ def unwrap(parsed_docstring: ParsedEpytextDocstring) -> str:
...
+
+def type2str(type_expr: Optional[ast.expr]) -> Optional[str]:
     if type_expr is None:
         return None
     else:
         src = astor.to_source(type_expr)
         assert isinstance(src, str)
         return src.strip()
 
+def str_and_line(obj: model.Documentable) -> Tuple[str, int]:
```
adiroiban, line 100: maybe rename `annotation_type_and_line` ... and add a short docstring. :)

pydoctor/test/test_astbuilder.py
```diff
@@ -71,17 +71,37 @@ def unwrap(parsed_docstring: ParsedEpytextDocstring) -> str:
...
-    return flatten(parsed_docstring.to_stan(None))
+    return flatten(parsed_docstring.to_stan(NotFoundLinker()))
 
-def type2str(type_expr: object) -> Optional[str]:
+@overload
+def type2str(type_expr: None) -> None: ...
+
+@overload
+def type2str(type_expr: ast.expr) -> str: ...
+
```
adiroiban, line 91: I am not sure why you don't write the implementation when you declare the overloading declaration and remove the definition without the overload decorator

```suggestion
@overload
def type2str(type_expr: None) -> None:
    return None

@overload
def type2str(type_expr: ast.expr) -> str:
    src = astor.to_source(type_expr)
    assert isinstance(src, str)
    return src.strip()
```

pydoctor/test/test_epydoc2stan.py
```diff
@@ -445,3 +446,52 @@ def test_xref_not_found_restructured(capsys: CapSys) -> None:
...
+
+    def resolve_identifier(self, identifier: str) -> Optional[str]:
+        self.requests.append(identifier)
+        return None
+
+    def resolve_identifier_xref(self, identifier: str, lineno: int) -> str:
+        assert False
+
+@mark.parametrize('annotation', (
+    '<bool>',
```
adiroiban, line 467: minor comment:

to make it easy to read and reduce the places were errors can be introduced, when creating parametrized test method I prefer to pass the input and the exact (already resolved) expected value.
So in this case, replace the `expected_lookups` and `expected_text` with actual values passes to `@mark.parametrize`

pydoctor/astbuilder.py
```diff
@@ -635,14 +638,17 @@ def _infer_type(expr):
     except ValueError:
         return None
     else:
-        return _annotation_for_value(value)
+        ann = _annotation_for_value(value)
```
adiroiban, line 641: maybe also pass `expr` as an argument to `_annotation_for_value` so that it can `fix_missing_locations` and return a fully resolved value.

pydoctor/test/test_astbuilder.py
mthuurne: I'd like to keep the name short to keep the asserts that use it readable. But I can rename it to `ann_str_and_line` to be a bit more descriptive.

I'll add a docstring.

pydoctor/test/test_astbuilder.py
mthuurne: [`overload`](https://docs.python.org/3/library/typing.html#typing.overload) doesn't work like that: the overloaded variants cannot be called, they're only for type checking. What you're thinking of is [`singledispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch), but I don't think the extra complexity of that is worth it when we only have 2 variants.

pydoctor/test/test_epydoc2stan.py
mthuurne: I actually had it like that at first, but I thought the compact syntax was more readable. For example:
```py
    '<Callable>[[<str>, <bool>], <None>]',
vs
    ('Callable[[str, bool], None]', ('Callable', 'str', 'bool', 'None')),
```
It does make the test code slightly more complex, but it only adds 2 lines so it's not that big of a risk.

Note that depending on what we pick in #280, the rule that the expected text must be equal to the input may no longer hold. If so, we'd have to switch to a 3-parameter format: input, expected text and expected lookups.

pydoctor/test/test_astbuilder.py
adiroiban: or just `ann_str_line` 

I am old school, and I prefer to write my own assertion helper... like `def assertStrAndLine(self, name, line, actual):`
and have all the ugly processing inside the method.

pydoctor/test/test_astbuilder.py
adiroiban: my bad... this is `typing.overload` and not https://www.python.org/dev/peps/pep-3124/ .. .and I see that PEP-3124 was not implemented.

thanks for the info.

pydoctor/epydoc/markup/__init__.py
```diff
@@ -184,9 +184,22 @@ class DocstringLinker:
...
+    def resolve_identifier(self, identifier: str) -> Optional[str]:
+        """
+        Resolve a Python identifier.
+        This will resolve the identifier like Python itself would.
+
+        @param identifier: The name of the Python identifier that
+            should be linked to.
+        @return: The URL of the target, or L{None} if not found.
+        """
+        raise NotImplementedError()
```
adiroiban, line 196: Since we don't have a test for this...maybe add 
```suggestion
        raise NotImplementedError()  # pragma: no cover
```

pydoctor/epydoc2stan.py
```diff
@@ -82,6 +82,15 @@ def look_for_intersphinx(self, name: str) -> Optional[str]:
         """
         return self.obj.system.intersphinx.getLink(name)
 
+    def resolve_identifier(self, identifier: str) -> Optional[str]:
+        fullID = self.obj.expandName(identifier)
+
+        target = self.obj.system.objForFullName(fullID)
+        if target is not None:
+            return target.url
```
adiroiban, line 90: it looks like there is not test for this line.

pydoctor/epydoc2stan.py
```diff
@@ -552,8 +561,93 @@ def __init__(self, annotation: ast.expr) -> None:
...
+        self.linker = linker
+
+    def _handle_name(self, identifier: str) -> Tag:
+        url = self.linker.resolve_identifier(identifier)
+
+        tag: Tag
+        if url is None:
+            tag = tags.transparent(identifier)
+        else:
+            tag = tags.a(identifier, href=url, class_='code')
```
adiroiban, line 582: it looks like there is no coverage for this line.

pydoctor/test/test_astbuilder.py
mthuurne: A custom assert method would fit well in the unittest/trial model, but with the way pytest reports test failures, I prefer to have the asserts outside the helper functions (when feasible).

pydoctor/epydoc2stan.py
```diff
@@ -552,8 +561,93 @@ def __init__(self, annotation: ast.expr) -> None:
...
+        return self._handle_name(node.id)
+
+    def visit_Attribute(self, node: ast.Attribute) -> Tag:
+        parts = []
+        curr: ast.expr = node
+        while isinstance(curr, ast.Attribute):
+            parts.append(curr.attr)
+            curr = curr.value
+        if not isinstance(curr, ast.Name):
+            return self.generic_visit(node)
```
adiroiban, line 614: missing coverage?

pydoctor/epydoc/markup/__init__.py
mthuurne: I'd rather exclude all `raise NotImplementedError` instances from coverage in the configuration.

pydoctor/epydoc2stan.py
mthuurne: Correct: the `resolve_identifier()` method is implemented by the test helper class `RecordingAnnotationLinker` in the new tests, instead of the implementation in `_EpydocLinker`. And the old tests don't happen to trigger this line. But the old tests don't validate any of the method, so the other lines aren't really tested either, even though they execute.

pydoctor/epydoc2stan.py
mthuurne: Yes, that is also a consequence of the `RecordingAnnotationLinker` test helper always returning "not found".

I don't want to deal with HTML when comparing the test results though, at that is messy and brittle. And just executing the line without comparing its output, to check that it doesn't crash, is not very valuable now that we have type checking, since mypy will catch the majority of issues that you can find this way. It might be useful in some cases, for example when there are non-trivial asserts in the code, but I don't think there are any on this code path.

adiroiban (approved):
It looks good. Thanks. Only minor comments.

pydoctor/epydoc2stan.py
mthuurne: Yes. The expected pattern for attribute lookups is that the sequence ends with a name, but the types from the AST module don't guarantee that. I'll try to find a code fragment that triggers this early return and add it to the tests.

pydoctor/epydoc2stan.py
mthuurne: Maybe some of the old tests that call `resolve_identifier_xref()` could be adapted to call and verify this method as well without too many changes; I'll have a look.

pydoctor/astbuilder.py
mthuurne: While that would simplify `_infer_type()`, it would move the complexity to `_annotation_for_value()`, which is both more complex already and has two non-None returns versus one. Also it can be called recursively via `_annotation_for_value()`, so we'd have to pass `expr` to that as well. All in all, I think this change would be a net negative for readability.

pydoctor/epydoc2stan.py
mthuurne: Partial success: it was fairly easy to adapt the old tests to also call the new method, but the old tests didn't cover the case where an identifier is found within the code being documented.

mthuurne: I've written a new test case to test a successful internal match by full name.

pydoctor/epydoc2stan.py
mthuurne: I can't actually find any code fragment that leads to this line being executed. While `ast.Attribute.value` is documented as having type `ast.expr`, in practice there doesn't seem to be any syntactically valid Python that produces an `ast.Attribute` for which the `value` is neither an `ast.Name` nor another `ast.Attribute`. 

However, if such an expression would exist, perhaps in a future Python version with new grammatical options, I think we should have pydoctor handle it gracefully rather than crashing on it. Note however that "handle it gracefully" currently means "let `astor` deal with it", but `astor` won't be able to format future grammar either without updates to its code.

I'll leave it as-is for now, but this is something we might want to revisit at some point. For example if we want to drop the dependency on `astor`: now that we do custom AST formatting, we don't actually use all that much functionality from that library anymore.

PR #281, opened 2020-11-23
Tweak coverage exclusions

mthuurne:
By excluding things we don't intend to cover, lines that should be covered but are not stand out more.


adiroiban (approved):
Thanks. All good.

PR #282, opened 2020-11-23
Include async functions in output

### LINKED ISSUE #279: PyDoctor seems to ignore "async def" methods
Hello,

As you can see here https://chorsley.github.io/python-Wappalyzer/Wappalyzer.WebPage.html

The function async method `WebPage.new_from_url_async` and `new_from_response_async` do not show in the documentation. 

They are defined like
```
    @classmethod
    async def new_from_response_async(cls, response):
```

Any clue?

Thanks

mthuurne:
An `async def` statement is parsed to an `AsyncFunctionDef` node in the AST, but this node type was ignored until now.

Fixes #279.

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -438,7 +438,13 @@ class ModuleVistor(ast.NodeVisitor):
 
         self.generic_visit(node)
 
+    def visit_AsyncFunctionDef(self, node):
+        self._handleFunctionDef(node, is_async=True)
+
     def visit_FunctionDef(self, node):
+        self._handleFunctionDef(node, is_async=False)
+
+    def _handleFunctionDef(self, node, is_async):
         # Ignore inner functions.
         if isinstance(self.builder.current, model.Function):
             return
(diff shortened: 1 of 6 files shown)


adiroiban (approved):
Looks good. Thanks.

My suggestion is to create a pydoctor.test.sample.epydoc or pydoctor.samples.epydoc package in which we include all the possible usage of pydoctor with epydoc 

In this way, the result of this PR could be observed via
https://pydoctor--282.org.readthedocs.build/en/282/api/pydoctor.test.html

But this can be merged as it is.
Thanks!

mthuurne:
> My suggestion is to create a pydoctor.test.sample.epydoc or pydoctor.samples.epydoc package in which we include all the possible usage of pydoctor with epydoc

I think it would be good to have a package that displays all the supported use cases, but I think it should be part of the documentation rather than part of pydoctor's code.

One reason is that we'd want to showcase syntax that is not available in all Python versions, such as position-only parameters. Having that syntax in the main source dir could cause problems with distros that pre-optimize Python modules on install. 

Another reason is that I wouldn't want samples to be mixed into the indexes of the actual API docs. In the package index it would be just one line, but in the class hierarchy and name index it would produce more noise.

adiroiban:
> Another reason is that I wouldn't want samples to be mixed into the indexes of the actual API docs. In the package index it would be just one line, but in the class hierarchy and name index it would produce more noise.

True. It should not be part of the pydoctor code.

We can build the sample as part if /api-demo/ or something like this... to not mix it with the actual pydoctor API.

We don't need to demonstrate all python version. We can start with just latest python (3.9)

Issue #225, opened 2020-10-10
Better support for restructuredtext directives


tristanlatr:
Hello,

Some of the restructuredtext directives are supported, like `.. contents::` which is great !

But the classic `.. warning::` or `.. note::` are not supported and are triggering the following warning:
```
bad docstring: NotImplementedError: <class 'pydoctor.epydoc.markup.restructuredtext._EpydocHTMLTranslator'> visiting unknown node type: warning
```

What reST directives pydoctor is currently supporting ? 

Would be cool to support `.. warning::`, `.. note::` and others :)

reST docs about directives here: https://docutils.sourceforge.io/0.4/docs/ref/rst/directives.html#admonitions

mthuurne:
We inherited reST support from epydoc and I think epydoc only supports the directives that can be mapped to equivalents in epytext. The supported fields can be found [here in the code](https://github.com/twisted/pydoctor/blob/master/pydoctor/epydoc/markup/restructuredtext.py#L70).

I agree that more support would be good to add, but I'll have to say "patches are welcome" here, since I have too many other things to work on already.

tristanlatr:
Thanks, that ok.

But I don't get something, `.. contents::` works but it does not appear in the code.
So some of the directive are handled in `docutils.parsers.rst` but not all of them ?

mthuurne:
I dug into the docutils source for a bit and it seems the `contents` directive generates a `topic` node. Since [`HTMLTranslator`](https://sourceforge.net/p/docutils/code/HEAD/tree/trunk/docutils/docutils/writers/_html_base.py) implements `visit_topic()`, it won't trigger a "visiting unknown node type" error.

So indeed, whether a directive works out of the box depends on whether docutils provides a default `visit_<node_type>()` implementation for it in `HTMLTranslator`.

tristanlatr:
Pydoctor now suports reST admonitions!

Issue #264, opened 2020-11-07
RST support as an extra


mthuurne:
pydoctor relies on `docutils` to parse and format reStructedText. But other documentation formats such as epytext work fine when `docutils` is not installed. To model this in the package metadata, we can create an extra for RST support and make that extra depend on `docutils`.

adiroiban:
I can take a look at that... this will also mean running a set of tests without sphinx and docutils installed ... `nodeps` here we come :)

tristanlatr:
Hello,

I think rest support should be kept as a native feature.

This will impose the users to install the extra packages with

`pip install pydoctor[rst]` for instance.

This will break workflows that relies on reST. Which is much more largely used than epytext.

Plus it will make hazardous to implement #16 and #123 ( I think, I'm no expert)

mthuurne:
Note that the pydoctor code for reST/ReST/RST (is there an official short form?) support is always included, whether you install the extra or not. The only difference is whether `pip` will install `docutils` if it wasn't already installed.

For supporting new document formats, using extras if they have external dependencies means that we can keep the dependencies of pydoctor itself low and pull in only the dependencies that the user actually needs.

The current situation is that there is no declared dependency on `docutils` at all, so declaring it in an extra is always an improvement over that.

Maybe this fine-grained dependency management is overkill though and we can make all pure Python dependencies unconditional. Are extras still widely used for optional features or only when dealing with native code that is not available everywhere?

tristanlatr:
> The current situation is that there is no declared dependency on docutils at all, so declaring it in an extra is always an improvement over that.

Did not understand that until yesterday, I guess my builds have `docutils` already installed...

> Are extras still widely used for optional features or only when dealing with native code that is not available everywhere?

No idea, But since RST it's much more widely used, I think it should be a native feature to support it with all directives. As well as markdown, google and numpy docstrings when (or if) they will be supported.

mthuurne:
> No idea, But since RST it's much more widely used, I think it should be a native feature to support it with all directives. As well as markdown, google and numpy docstrings when (or if) they will be supported.

I'm in favor of supporting every docstring format that sees significant use, provided that someone is willing to implement the format-specific parts (probably not me any time soon).

My concern is that when we support many formats, we'll also end up pulling in many dependencies for parsing those formats. For example in the case of Markdown, we'd need a parser library.

[...]

Issue #16, opened 2015-01-12
pydoctor should have the equivalent of __docformat__ = "restructuredtext en"


mwhudson:
Originally reported as Twisted issue 2187:

> epydoc has a epytext variant that's valid restructured text. You can either set **docformat** = "restructuredtext en" in your module, or pass --docformat restructuredtext to epydoc to use it.
> 
> pydoctor should support these as well, so that it can be a better drop-in replacement for epydoc. For instance, bzr's source uses restructuredtext format epytext, so pydoctor isn't hugely useful with it.

---

Imported from Launchpad using lp2gh.
- date created: 2009-08-06T18:47:24Z
- owner: exarkun
- the launchpad url was https://bugs.launchpad.net/bugs/410000

mwhudson:
(by mwhudson)
pydoctor has supported --docformat for ages now.  It doesn't support __docformat__ in modules, but I don't know how much I care.

tristanlatr:
This feature could to be actually useful for our own documentation !

PR #284, opened 2020-11-24
Docs update : Better RST specifications

tristanlatr:
This PR is the first step towards more support for RST markup. #225
It imports part le the docs from epydoc.
It also fixes a few documentation issues like broken links and intersphinx.
I've also written some tests for unimplemented RST directives like `.. note::`


tristanlatr:
I've also done a quick RST demo file, it's somewhat linked with #285, but for the RST part.

The RST quick guide is only useful as a demo for now, but as we add some documentation and feature specification it will become useful. It's a shame we can't build it inside the pydoctor package docs for now, this is because #16. But as discussed it's probably better to keep this in a separate project. 

What do you think ?
I'll add a few missing information and mark the PR ready for review

tristanlatr:
Also, any help with the Static code checks failure would be appreciated

mthuurne:
> Also, any help with the Static code checks failure would be appreciated

> pydoctor/test/epydoc/test_restructuredtext.py:6: error: Skipping analyzing 'bs4': found module but no type hints or library stubs  [import]

Beautiful Soup doesn't ship with type annotations (yet?), so it needs the following piece of configuration in `mypy.ini`:
```ini
[mypy-bs4.*]
ignore_missing_imports=True
```

I'll address the other mypy errors using the review UI.

docs/source/index.rst
```diff
@@ -10,6 +10,10 @@ Installation
 
    $ pip install -U pydoctor
 
+Install with *restructuredtext* support::
+
+   $ pip install -U pydoctor docutils
+
```
mthuurne, line 16: I'd prefer to implement #264, for a more reliable way of installing pydoctor with RST support.

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -1,8 +1,11 @@
+import pprint
 from typing import List
-
-from pydoctor.epydoc.markup import ParseError, flatten
+from unittest import skip
```
mthuurne, line 3: The imported `skip` isn't used. But while pytest has some support for the built-in `unittest` framework, I think it's better to use [pytest's native markers](https://docs.pytest.org/en/stable/skipping.html) instead.

If you were looking for a marker for the commented out tests, I think `xfail` (expected failure) is more appropriate here than `skip`: it will force the test case to execute, but considers the result good if it fails rather than if if it succeeds.

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -26,3 +29,80 @@ def test_rst_anon_link_email() -> None:
     assert html.startswith('<a ')
     assert ' href="mailto:postmaster@example.net"' in html
     assert html.endswith('>mailto:postmaster@example.net</a>')
+
+# THIS IS HELPER METHODS
+from pydoctor.epydoc.markup import flatten, restructuredtext
+def to_html(docstring) -> str:
```
mthuurne, line 35:
```suggestion
def to_html(docstring: str) -> str:
```

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -26,3 +29,80 @@ def test_rst_anon_link_email() -> None:
...
+    """
+    Utility method to convert a docstring to html with pydoctor. 
+    """
+    err = []
```
mthuurne, line 39:
```suggestion
    err: List[ParseError] = []
```

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -26,3 +29,80 @@ def test_rst_anon_link_email() -> None:
...
+    p = restructuredtext.parse_docstring(docstring, err)
+    if [ e.__dict__ for e in err ]!=[]:
```
mthuurne, line 41:
```suggestion
    if err:
```
This is testing whether there were any errors, right?

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -26,3 +29,80 @@ def test_rst_anon_link_email() -> None:
...
+# THIS IS HELPER METHODS
+from pydoctor.epydoc.markup import flatten, restructuredtext
+def to_html(docstring) -> str:
+    """
+    Utility method to convert a docstring to html with pydoctor. 
+    """
+    err = []
+    p = restructuredtext.parse_docstring(docstring, err)
+    if [ e.__dict__ for e in err ]!=[]: 
+        raise ValueError("\n".join([ pprint.pformat(e.__dict__) for e in err ]))
```
mthuurne, line 42:
```suggestion
        raise ValueError("\n".join(str(e) for e in err))
```
If the output of `str()` or `repr()` isn't good enough, we can improve the implementation of `ParseError.__str__()` or `ParseError.__repr__()` instead of working around it here.

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -26,3 +29,80 @@ def test_rst_anon_link_email() -> None:
...
+    html=flatten(p.to_stan(None))
+    return html
+
+def parse_and_print(s):
```
mthuurne, line 46:
```suggestion
def parse_and_print(s: str) -> None:
```

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -26,3 +29,80 @@ def test_rst_anon_link_email() -> None:
     assert html.startswith('<a ')
     assert ' href="mailto:postmaster@example.net"' in html
     assert html.endswith('>mailto:postmaster@example.net</a>')
+
+# THIS IS HELPER METHODS
+from pydoctor.epydoc.markup import flatten, restructuredtext
```
mthuurne, line 34: Please put the imports all at the top, unless a late import is absolutely necessary to avoid import cycles (but then it's usually inside a function).

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -26,3 +29,80 @@ def test_rst_anon_link_email() -> None:
     assert html.startswith('<a ')
     assert ' href="mailto:postmaster@example.net"' in html
     assert html.endswith('>mailto:postmaster@example.net</a>')
+
+# THIS IS HELPER METHODS
```
mthuurne, line 33: The fact that their name doesn't start with `test_` already identifies them as helper methods, so this comment is not needed.

pydoctor/test/epydoc/test_restructuredtext.py
mthuurne: If this function is here to stay, it would be best to remove it from `restructuredtext.doctest` and import it from this test module instead, to avoid code duplication.

mthuurne (changes requested):
I did a preliminary review of the code changes. I haven't read the generated documentation yet and haven't looked into the CI tasks yet in detail.

docs/source/index.rst
tristanlatr: I agree, I also discovered that some directive need Sphinx actually not only docutils

pydoctor/test/epydoc/test_restructuredtext.py
mthuurne: Or maybe the doctest should be converted into regular test cases. While the doctests are fairly readable, they lack the flexibility of being able to execute individual test cases and also the reporting when a test fails nowhere near as useful as pytest's usual reporting.

pydoctor/test/epydoc/test_restructuredtext.py
tristanlatr: Yep, I'll change that

docs/source/index.rst
mthuurne: How does that work? Does Sphinx automatically register additional directive handlers for docutils, or should pydoctor make calls to Sphinx to set this up?

tristanlatr: It looks like it's the `.. code:: python` directive the issue:

```
docs/source/PyDoctor_ReST_Quick_Reference.py:169: bad docstring: Cannot analyze code. Pygments package not found.
```

tristanlatr:
Thanks for the insights @mthuurne , I've applied your preliminary comments.
Also, I've commented the link to the RST demo as the publish link is not the good one and the content is all very raw (https://tristanlatr.github.io/pydoctor/rst-quick-ref/PyDoctor_ReST_Quick_Reference.html). 
Also we don't have the same thing for Epytext yet.

docs/source/index.rst
mthuurne: Hmm, we already have a Python syntax highlighter inherited from epydoc and I think Twisted contains another, so pulling in Pygments for that purpose would be a bit redundant. Would it be possible to register our own directive handler?

Another option would be to drop the epydoc highlighter and adopt Pygments ourselves.

docs/source/rst_quick_ref.rst
```diff
@@ -0,0 +1,449 @@
...
+~~~~~~~~~~
+
+line-block
+++++++++++
+
+.. list-table:: 
+   :header-rows: 1
+
+   * - Plain text
+     - Result
```
mthuurne, line 119: I think "Plain text" doesn't really describe what that column is about.
```suggestion
   * - Source
     - Output
```
or
```suggestion
   * - Docstring
     - Result
```

docs/source/rst_quick_ref.rst
```diff
@@ -0,0 +1,449 @@
...
+
+        | But can a bee be said to be
+        |     or not to be an entire bee,
+        |         when half the bee is not a bee,
+        |             due to some ancient injury?
+     -
+        | But can a bee be said to be
+        |     or not to be an entire bee,
+        |         when half the bee is not a bee,
+        |             due to some ancient injury?
```
mthuurne, line 142: This is output with a straight left margin. Is the output incorrect? Or is the example confusing in that it suggests the input would be treated as pre-formatted text while it is not?

docs/source/docformat.rst
```diff
@@ -4,35 +4,276 @@ Documentation Formats
...
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#setting-the-default-interpreted-text-role>`_
+     - Yes
+    
+   * - ``.. line-block::``
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#line-block>`_
+     - eh
+
+   * - ``.. code::``
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#code>`_
+     - Yes (syntax highlight ignored). Use ``.. python::``.
```
mthuurne, line 121: I think the problem might be that the CSS is missing, similar to what happened for the `python` directive before.

Pygments can generate the required CSS, but if nothing is requesting this CSS from Pygments and writing it to file, the browser won't have access to it.

docs/source/rst_quick_ref.rst
```diff
@@ -0,0 +1,449 @@
...
+
+.. list-table:: 
+   :header-rows: 1
+
+   * - Plain text
+     - Result
+     
+   * - :: 
+
+        .. figure:: https://khms1.googleapis.com/kh?v=878&hl=en-US&x=2273&y=3006&z=13
```
mthuurne, line 258: Could we use a local file instead? Unfortunately pydoctor doesn't have a logo, but perhaps we can use the Twisted or Python logo, those should be pretty small.

docs/source/rst_quick_ref.rst
```diff
@@ -0,0 +1,449 @@
...
+    
+     -  .. table:: Truth table for "not"
+            :widths: auto
+
+            =====  =====
+            A      not A
+            =====  =====
+            False  True
+            True   False
+            =====  =====
```
mthuurne, line 301: In Chromium the "not A" column header is spread over two lines, which looks really bad. The problem might be due to missing CSS: there are a number of classes with a `rst-` prefix defined on this table, but no loaded style sheet defines those.

mthuurne (changes requested):
I like where this is going, but it will need a number of changes before it's ready to merge.

tox.ini
```diff
@@ -52,12 +53,31 @@ commands =
...
+commands =
+    rm -rf ./build/rst-quick-ref
+    mkdir -p ./build
+
+    pydoctor \
+        --docformat=restructuredtext \
+        --project-name="PyDoctor ReST Quick Reference" \
+        --project-base-dir=./docs/source/ \
+        --html-output=./build/rst-quick-ref \
+        ./docs/source/PyDoctor_ReST_Quick_Reference.py
```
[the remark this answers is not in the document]
mthuurne: ...unless we decide to make this part of the base docs, of course, but then it wouldn't need its own `tox` env to begin with.

mthuurne:
Instead of having one page that lists supported directives and another that demonstrates the output, maybe we should combine both?

We could implement two custom directives, one which takes a fragment of Python source and another which takes a docstring and both would format that input using pydoctor. That would allow mixing pydoctor output into the documentation at a fine-grained level.

mypy.ini
```diff
@@ -53,3 +53,6 @@ ignore_missing_imports=True
 
 [mypy-urllib3.*]
 ignore_missing_imports=True
+
+[mypy-bs4.*]
+ignore_missing_imports=True
```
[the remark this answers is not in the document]
tristanlatr: I've configured visual studio code to "Insert Final Newline" on save now. No idea why this is so much of an issue but I'll go with the flow...

docs/source/index.rst
tristanlatr: I see. The issue here is if we drop Pygments we need to say `.. code::` directive is not supported yet. And this PR is not about coding fancy directive for pydoctor yet neither (I haven't looked in details how to do so).
So I'll drop Pygments dep and say `.. code::` is not supported yet if that's ok with you

.github/workflows/rst-quick-ref.yaml
```diff
@@ -0,0 +1,46 @@
...
+        tox -e rst-quick-ref || true
+        
+    - name: Publish Pydoctor documentation to the gh-pages branch, folder rst-quick-ref
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: ./build/rst-quick-ref
+        commit_message: "Generate Pydoctor documentation"
+        destination_dir: rst-quick-ref
+        keep_files: true
```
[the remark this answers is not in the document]
tristanlatr: Yes we want it as a chapter of the RTD site, I'll delete the tox env and yml file. 
It's just that for now we can't do that expect with an iframe. But maybe it's enough for the quick ref ? In any case, the `pydoctor_sphinx_ext.build_apidocs` needs an edit to allow several runs of pydoctor with differents configs for that to work.

docs/source/rst_quick_ref.rst
```diff
@@ -0,0 +1,449 @@
...
+  :contributor: `Glyph Lefkowitz  <mailto:glyph@twistedmatrix.com>`__
+  :contributor: `Edward Loper  <mailto:edloper@gradient.cis.upenn.edu>`__
+  :contributor: `Bruce Mitchener  <mailto:bruce@cubik.org>`__
+  :contributor: `Jeff O'Halloran  <mailto:jeff@ohalloran.ca>`__
+  :contributor: `Simon Pamies  <mailto:spamies@bipbap.de>`__
+  :contributor: `Christian Reis  <mailto:kiko@async.com.br>`__
+  :contributor: `Daniele Varrazzo  <mailto:daniele.varrazzo@gmail.com>`__
+  :contributor: `Jonathan Guyer <mailto:guyer@nist.gov>`__
+
+  .. |copy| unicode:: 0xA9 .. copyright sign
```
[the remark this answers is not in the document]
tristanlatr: As I said in the document header, all of this is a working draft, I agree I could have worked a bit more on presenting those. I might just remove it altogether

docs/source/docformat.rst
```diff
@@ -4,35 +4,276 @@ Documentation Formats
...
+   * - ``.. list-table::``
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#list-table>`_
+     - Yes
+
+   * - ``.. warning::`` and other abnomitions
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#specific-admonitions>`_
+     - No 
+
+   * - ``.. versionadded::``
+     - `ref (Sphinx) <https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded>`_
```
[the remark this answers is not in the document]
tristanlatr: > Should we be supporting Sphinx directives?

I think we should support the directive that see significant usage, like ``.. versionadded::``, ``.. note::``, etc...

docs/source/rst_quick_ref.rst
tristanlatr: There is indeed something wrong with line block syntax handling. 
https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#line-blocks
We can read that "line breaks are preserved. Initial indents are also significant, resulting in a nested structure", this is not what pydoctor does

docs/source/rst_quick_ref.rst
tristanlatr: Same here on safari

docs/source/docformat.rst
```diff
@@ -4,35 +4,276 @@ Documentation Formats
...
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#unicode-character-codes>`_
+     - Yes
+ 
+   * - ``.. raw::``
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#raw-data-pass-through>`_
+     - Yes
+  
+   * - ``.. class::``
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#class>`_
+     - eh
```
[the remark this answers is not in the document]
tristanlatr: Something like "I don't know if that's the expected behaviour" and it also mean that I'm lazy. I'll work more on that to show them in the quick ref so we can assess if their supported or not

docs/source/docformat.rst
```diff
@@ -4,35 +4,276 @@ Documentation Formats
...
+   * - ``.. line-block::``
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#line-block>`_
+     - eh
+
+   * - ``.. code::``
+     - `ref (docutils) <https://docutils.sourceforge.io/docs/ref/rst/directives.html#code>`_
+     - Yes (syntax highlight ignored). Use ``.. python::``. 
+   
+   * - ``.. python::``
+     - `ref (pydoctor) <https://pydoctor.readthedocs.io/en/latest/api/pydoctor.epydoc.markup.restructuredtext.PythonCodeDirective.html>`_
```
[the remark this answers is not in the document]
tristanlatr: >  I think it would be best to link to pydoctor's version of these directives.

Some of them are not supported yet, so the links are directed to epydoc apidocs for now. For now it's unclear what will be the links to the "pydoctor reference manual" so I'll just put pydoctor as plaintext and we'll link to the quick ref once we have it published.

tristanlatr (commented):
> We could implement two custom directives, one which takes a fragment of Python source and another which takes a docstring and both would format that input using pydoctor. That would allow mixing pydoctor output into the documentation at a fine-grained level.

This is a great idea, but it's not in the scope of this PR. We would need to make the pydoctor CSS concurrent with Read the Docs theme's. 

In the same topic:
For now, the best way to include pydoctor's output into sphinx docs is [with an iframe](https://github.com/tristanlatr/pydoctor/blob/pydoctor-sphinx-iframe/docs/source/api.rst) (it's not the best because we loose the links to the html files). We would need to think of a directive that would generates pydoctor's output and include Sphinx toctree into the generated HTML or something. That would be theme dependant I suppose. 

What I can do is merge the table and the quick guide together into one "rst" file, then we'de need to find a way to build it https://github.com/twisted/pydoctor/pull/284#discussion_r530534193

tristanlatr:
Ok I manage to build the ReST quick ref and link it inside sphinx with an iframe. I didn't want to get into creating new directives yet.

I've edited `pydoctor_sphinx_ext.build_apidocs` to allow several runs of pydoctor with differents configs.
BUT the ReST quick reference module won't build inside sphinx... It throws error `'Values' object has no attribute 'env'` (It might be the same cause of https://github.com/sphinx-doc/sphinx/issues/2922). So I'm using the the `tox -e docs` command to generate the docs for now (it works). 
Any help with fixing this issue would be appreciated.

Meanwhile, I'll continue to merge the table and the output exemples. 


Here is what it looks like with the iframe, what do you think ?

by default 2020-11-26 at 12 27 22 PM

tristanlatr:
What I think is happening is sphinx is replacing some of the docutils default directives (supported by pydoctor) by their own, which are not supported.

tristanlatr:
I decided to remove altogether the not very useful yet long list of directives to focus on porting the part of the epydoc documentation , for the reST part, to our docs. We have a list of supported directives and pydoctor do not support any of the sphinx or epydoc directive (except `.. python::`) yet so we don't need this long list...

Ideally we would port all the docs from http://epydoc.sourceforge.net/epydoc.html to read the docs since almost everything applies to pydoctor too.

mthuurne (approved):
Looks good to me.

I've synced with the current master branch and added a few commits. Please let me know if you're OK with these changes.

tristanlatr:
> Please let me know if you're OK with these changes

Of course I'm ok

I've added a few minor changes to documentation.

It's ready to be merged

tristanlatr:
I don't understand all the failures, the tests are passing locally (except the one comparing the absolute path).
May be due to the current github action issue https://www.githubstatus.com/incidents/sxthxvdlljk4

mthuurne:
One of them reads "GitHub Actions has encountered an internal error when running your job." so it's likely that it's indeed an infrastructure problem.

PR #286, opened 2020-11-24
Improve error message when a link target is not found

### LINKED ISSUE #128: "invalid ref" to symbol that is present in module namespace

mthuurne:
Closes #128

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -146,14 +146,10 @@ class _EpydocLinker(DocstringLinker):
         if target is not None:
             return target.url
 
-        if identifier == fullID:
-            self.obj.report(
-                "invalid ref to '%s' not resolved" % (identifier,),
-                'resolve_identifier_xref', lineno)
-        else:
-            self.obj.report(
-                "invalid ref to '%s' resolved as '%s'" % (identifier, fullID),
-                'resolve_identifier_xref', lineno)
+        message = f'Cannot find link target for "{fullID}"'
+        if identifier != fullID:
+            message = f'{message} (resolved from "{identifier}")'
+        self.obj.report(message, 'resolve_identifier_xref', lineno)
         raise LookupError(identifier)
 
 
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -148,7 +148,10 @@ class _EpydocLinker(DocstringLinker):
 
         message = f'Cannot find link target for "{fullID}"'
         if identifier != fullID:
-            message = f'{message} (resolved from "{identifier}")'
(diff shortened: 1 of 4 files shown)


pydoctor/epydoc2stan.py
```diff
@@ -146,14 +146,10 @@ def resolve_identifier_xref(self, identifier: str, lineno: int) -> str:
...
 
-        if identifier == fullID:
-            self.obj.report(
-                "invalid ref to '%s' not resolved" % (identifier,),
-                'resolve_identifier_xref', lineno)
-        else:
-            self.obj.report(
-                "invalid ref to '%s' resolved as '%s'" % (identifier, fullID),
-                'resolve_identifier_xref', lineno)
+        message = f'Cannot find link target for "{fullID}"'
```
adiroiban, line 149: Maybe also hint for a possible solution

```suggestion
        message = f'Cannot find link target for "{fullID}" (maybe missing intersphinx inventory)'
```

adiroiban (approved):
All good. Thanks!

pydoctor/epydoc2stan.py
mthuurne: That's only one possible reason though. Other reasons could be a typo or pydoctor resolving the identifier incorrectly.

Would this qualify as a FAQ? I've seen the old error message come up in discussions several times, but the confusing phrasing was likely a contributing factor in that.

Another option would be to only display the intersphinx hint if `fullID` has a different root than the software being documented. I'll do a quick experiment to see if this is feasible.

adiroiban: > Would this qualify as a FAQ? I've seen the old error message come up in discussions several times, but the confusing phrasing was likely a contributing factor in that.

Yes. I think hat a FAQ can help. It would be the second best... but we can add the link into the error message.

In the idea case, the error message should contain enough information so that you know what step to take to fix it.

Thanks!

mthuurne: I ended up implementing the conditional message before merging, see 7095e72.

adiroiban: Yep. I saw that. I think that is ok.
But for the future we can keep the error message logic simple and link to a FAQ.

I see this type of error reporting is common with Google / MS / GitHub products.
You can an error message and a web link.

PR #289, opened 2020-11-25
Remove --html-write-function-pages option

mthuurne:
This feature existed to produce pages that third party tools could link to, but intersphinx inventories are a much better solution.

I'm leaving this PR as a draft for now, since we shouldn't merge this before [Twisted stops using function pages](https://twistedmatrix.com/trac/ticket/10057).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -63,7 +63,7 @@ What's New?
 pydoctor dev
 ^^^^^^^^^^^^
 
-* Removed the ``--html-write-function-pages`` option. Please use the generated Intersphinx inventory (objects.inv) for deep-linking your documentation.
+* Removed the ``--html-write-function-pages`` option. As a replacement, you can use the generated Intersphinx inventory (objects.inv) for deep-linking your documentation.
 
 pydoctor 20.12.1
 ^^^^^^^^^^^^^^^^
--- a/README.rst
+++ b/README.rst
@@ -60,6 +60,11 @@ You can select a different format using the ``--docformat`` option.
 What's New?
 ~~~~~~~~~~~
 
+pydoctor dev
+^^^^^^^^^^^^
+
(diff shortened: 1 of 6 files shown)


adiroiban:
I think that this can be merged now.

Since there is no new release of pydoctor,  for Twisted I plan to install pydoctor based on a sha commit.
I want to use a newer version of pydoctor as it has the Read The Docs integration  and in this way, I don't have to reinvent the wheel for Twisted.

This is why I would like to have pydoctor release now :)

mthuurne:
> I think that this can be merged now.
> 
> Since there is no new release of pydoctor, for Twisted I plan to install pydoctor based on a sha commit.
> I want to use a newer version of pydoctor as it has the Read The Docs integration and in this way, I don't have to reinvent the wheel for Twisted.

Even if it doesn't cause problems for Twisted, there is the issue that we use Twisted in our system test, which is currently failing because of the removal of the `--html-write-function-pages` option.

> This is why I would like to have pydoctor release now :)

I do want to release soon (and without this PR), but the parameter annotation support isn't finished yet, which makes the generated documentation significantly less useful.

README.rst
```diff
@@ -60,6 +60,11 @@ You can select a different format using the ``--docformat`` option.
 What's New?
 ~~~~~~~~~~~
 
+pydoctor dev
+^^^^^^^^^^^^
+
+* Removed the ``--html-write-function-pages`` option. Please use the generated Intersphinx inventory (objects.inv) for deep-linking your documentation.
```
adiroiban, line 66: No need for the "please" part :)

There was a guide (by Google or maybe Gnome) in which it was described how to write user facing text  / messages.
I remember that one of the now as to not use "Please"

```suggestion
* Removed the ``--html-write-function-pages`` option. As a replacement, you can use the generated Intersphinx inventory (objects.inv) for deep-linking your documentation.
```

adiroiban (approved):
All good. Thanks!

PR #291, opened 2020-11-26
Replace argspec by inspect.Signature

### LINKED ISSUE #283: Implementation of function arguments
In the Python 3.9 AST, function arguments are defined as follows:

arguments = (arg* posonlyargs, arg* args, arg? vararg, arg* kwonlyargs, expr* kw_defaults, arg? kwarg, expr* defaults)
arg = (identifier arg, expr? annotation, string? type_comment)

Position-only arguments were added in Python 3.8, while keyword-only arguments exist since Python 3.0. While type annotations are gathered from both these categories, in the `def` line they are not presented yet, because the code for that was written during the Python 2.x days and hasn't been updated yet.

The information about function arguments currently stored in `pydoctor.model.Function` as follows:
```py
argspec: Tuple[Sequence[str], Optional[str], Optional[str], Sequence[str]]
         (args, varargname, kwargname, defaults)
annotations: Mapping[str, Optional[ast.expr]]
             {name: annotation}
```
The `argspec` is similar to an outdated version of the AST (outdated because pos-only and kw-only are missing).

As this format is not particularly easy to work with, `getBetterThanArgspec()` in `pydoctor.templatewriter.pages` matches up argument names and default values. This is again missing support for position-only arguments (inevitably, since `argspec` is the input). The "better than argspec" return value is structured like this:

```
(name*, (name, default)*)
```

I have some doubts whether this is the ideal data structure:
- it would have to be extended with support for position-only and keyword-only arguments
- it doesn't include information about `varargname` and `kwargname`, so it's not a standalone replacement for `argspec`
- in general I prefer minimal classes (dataclasses/attrs) over tuples, especially when they're part of a larger data structure

We could take advantage of the fact that dictionaries are order-preserving since Python 3.6, by having a dictionary that maps each name to a class containing the information we have on that parameter.

I think that if we prefer a different format to `argspec`, it would be better to store that friendly format inside the code model. Which means the conversion should happen in the `astbuilder` module rather than in the `templatewriter` package. One advantage would be that it integrates better with the unit tests: the conversion would be easier to test, while the tests would be easier to read and write since they can use a friendlier format.

The `signature()` function in `pydoctor.templatewriter.pages` takes the "better than argspec" data and presents the `def` line in the output. This function still has support for detupling of arguments, a language feature that was dropped in Python 3, so there is some opportunity for simplifying the code.

Currently `signature()` tries to mimic the way the line looks in the source code. Is that the best possible presentation though? Enumerating the positional parameters is useful, in any case. But for keyword-only parameters, having them in the `def` line isn't as useful.

mthuurne:
This replaces the `argspec` data structure by the `inspect.Signature` class, as suggested by @mwhudson. Besides simplifying the code, this PR also adds support for position-only and keyword-only parameters and it improves the presentation of parameter default values.

Closes #283

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -481,24 +481,34 @@ class ModuleVistor(ast.NodeVisitor):
                 func.kind = 'Class Method'
 
         args = [arg.arg for arg in node.args.args]
+        defaults = node.args.defaults
+        if defaults:
+            withdefaults = list(zip(reversed(args), reversed(defaults)))
+            withdefaults.reverse()
+            nodefaults = args[:-len(withdefaults)]
+        else:
+            withdefaults = []
+            nodefaults = args
+
+        parameters = []
+
+        for name in nodefaults:
+            parameters.append(Parameter(name, Parameter.POSITIONAL_OR_KEYWORD))
 
         vararg = node.args.vararg
-        varargname = None if vararg is None else vararg.arg
+        if vararg is not None:
+            parameters.append(Parameter(vararg.arg, Parameter.VAR_POSITIONAL))
+
+        for name, default in withdefaults:
+            parameters.append(Parameter(name, Parameter.POSITIONAL_OR_KEYWORD,
+                                        default=_ValueFormatter(default)))
 
         kwarg = node.args.kwarg
-        kwargname = None if kwarg is None else kwarg.arg
+        if kwarg is not None:
+            parameters.append(Parameter(kwarg.arg, Parameter.VAR_KEYWORD))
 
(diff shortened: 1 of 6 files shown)


mwhudson:
I'm not really competent to review the construction of the Signature objects, but this seems nice overall.

adiroiban (approved):
Looks good. thanks. Happy to see the move to inspect.

I was expecting that there was already a way to move from AST to Signature

mthuurne:
> I was expecting that there was already a way to move from AST to Signature

I did look for it in the standard library, but couldn't find anything.

Issue #295, opened 2020-11-27
Lookup of name in annotation fails on reparented object


mthuurne:
`twisted.logger.extractField()` has return type `typing.Any`, but in the output this is not linked to the standard library documentation, despite the intersphinx inventory being available and other types being linked successfully.

What I think is happening here is that `extractField()` is implemented in `twisted.logger._flatten`, which imports `Any` from the `typing` module in the standard library. However, by the time the link is resolved, the `extractField()` has been moved into the `twisted.logger` module, it's public location. But this module does not import `Any`, so the lookup fails.

Note that doing the lookup before the move isn't correct either: when we link to other objects that get moved around, that link would be broken after the move. A similar problem occurs if we'd keep a reference to the original parent location and attempt to do a lookup there after all moves have happened, since the target would have been moved away then. I think the only way to do it correctly is to look up the target `Documentable` in a code model that reflects the actual source and when presenting ask that `Documentable` where in the output it resides and link there.

In an earlier discussion of problems related to cyclic imports, [I wrote](https://github.com/twisted/pydoctor/pull/277#issuecomment-729671525):

> As for a long term solution, I'm starting to think that reparenting is a bad idea: it means we lose information about how the code is structured. Instead of altering the code model, we could create layer on top of it that contains the documentation structure. That would also allow us to thin down the interface of the model classes a bit, by moving presentation data out of the code model.
>
> Linking annotations containing imported names would then work as follows:
> - look up the original code object in the same way Python does (trace it back recursively to source location)
> - look up where the code object ended up in the documentation layer (might not match source location if name was re-exported)
> - link to the documentation location

@adiroiban replied:

> +1 yes. I think that we should always have a layer with the true structure.. and with that we can have a public API layer which does it best to remove duplicates and present the doc links using "canonical" paths.

I think that's a good way to look at it: objects have an implementation location, being the source file they're defined in, and a public API location, which is where the user will read about them in the documentation and import them from.

There are more issues related to reparenting, for example #184, in which the same name is re-exported in more than one location. So the mapping between implementation layer and public API layer might not be 1:1.

tristanlatr:
Keeping a layer with the "true structure" will also allow to display the Class in the module summary where they are actually defined, but link to where they are exported. It can be confusing to see an empty module because all classes are exported.

mthuurne:
[...]

Issue #294, opened 2020-11-26
Format integer literals like the source


mthuurne:
The source of `twisted.internet.posixbase.PosixReactorBase` reads:
```py
    def listenUNIX(self, address, factory, backlog=50, mode=0o666, wantPID=0):
```
But this is formatted in pydoctor as:
```py
    def listenUNIX(self, address, factory, backlog=50, mode=438, wantPID=0):
```
While decimal 438 is the same value as otcal 666, the meaning of `0o666` is immediately clear to someone familiar with POSIX file modes while that is not the case for `438`.

The AST does not contain the original formatting of a constant, only the value. However, with accurate line number information we could pick the right slice from the original source file and present that instead.

In theory we could do the same for other literals, but I'm not sure if it would be as useful there as it is for integers.

tristanlatr:
> with accurate line number information we could pick the right slice from the original source file

No idea how you would do that!

This sounds like a very hard task for limited benefits. But maybe I'm mistaken.

Issue #229, opened 2020-10-21
Warn on duplicate fields


mthuurne:
If for example a return value or a parameter is documented more than once, it would be useful to warn the user about that.

Note that fields such as `author`, `note` and `seealso` can be used multiple times, so not all duplicate fields are a problem.

mthuurne:
Duplicate `type` fields are already warned about, see `FieldHandler.redef()`. But the warning message itself still uses the old-style full name format, not source location.

mthuurne:
#296 implements a warning for duplicate `param` fields. This warning found [several issues in Twisted](https://github.com/twisted/twisted/pull/1486).

It would be good to have a warning for duplicate `type` fields as well. But the way the `self.types` dictionary is filled makes this messy to implement, since it is pre-filled with the presentations of the type annotations, so we can't easily detect when a type is defined twice.

The same `keyword` occurring twice is also something we can warn about, but as keywords aren't used often, so I don't consider this a priority.

PR #296, opened 2020-11-27
Do more with parameter inspection

### LINKED ISSUE #217: Integrate function AST inspection and doc tags better
Currently, the function parameter list in the output contains parameters documented using `@param` tags or the returned value using the `@return` tag, supplemented by type annotations gathered from the AST. This leads to sub-optimal output:

- parameters/return values that are annotated but not documented are not shown, while we could show type information for those
- documentation for non-existing parameters (typically typos or removed parameters) is included in the output

The code that handles parameters is a bit of a mess though, since the data is reshuffled several times. Ideally we'd gather info in a single data structure shared by epydoc, astbuilder and epydoc2stan.

Implementing #216 would allow even more accurate output, but is not required.

mthuurne:
Fixes #217

Known issues: (to be fixed after release)
- #229: we don't warn about duplicate `type` or `keyword` fields yet: `type` is not easy to implement, `keyword` is barely used
- #294: doesn't happen all that often
- #295: annoying but not easy to fix

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -257,7 +257,13 @@ class FieldHandler:
         ret_type = formatted_annotations.pop('return', None)
         self.types.update(formatted_annotations)
         if ret_type is not None:
-            self.return_desc = FieldDesc(kind='return', type=ret_type)
+            # In most cases 'None' is not an actual return type, but the absence
+            # of a returned value. Not storing it is the easiest way to prevent
+            # it from being presented.
+            ann_ret = annotations['return']
+            assert ann_ret is not None  # ret_type would be None otherwise
+            if not _is_none_literal(ann_ret):
+                self.return_desc = FieldDesc(kind='return', type=ret_type)
 
     def handle_return(self, field: Field) -> None:
         if field.arg is not None:
@@ -426,6 +432,11 @@ class FieldHandler:
             return tags.table(class_='fieldTable')(r) # type: ignore[no-any-return]
         else:
             return tags.transparent # type: ignore[no-any-return]
+
+
+def _is_none_literal(node: ast.expr) -> bool:
+    """Does this AST node represent the literal constant None?"""
+    return isinstance(node, (ast.Constant, ast.NameConstant)) and node.value is None
 
 
 def reportErrors(obj: model.Documentable, errs: Sequence[ParseError]) -> None:
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -5,8 +5,8 @@ Convert epydoc markup into renderable content.
 from collections import defaultdict
 from importlib import import_module
 from typing import (
-    Callable, DefaultDict, Dict, Iterable, Iterator, List, Mapping, Optional,
-    Sequence, Tuple
+    Callable, ClassVar, DefaultDict, Dict, Iterable, Iterator, List, Mapping,
+    Optional, Sequence, Tuple
 )
 import ast
 import itertools
@@ -158,6 +158,7 @@ class _EpydocLinker(DocstringLinker):
 
 @attr.s(auto_attribs=True)
 class FieldDesc:
+    _UNDOCUMENTED: ClassVar[Tag] = tags.span(class_='undocumented')("Undocumented")
 
     kind: str
     name: Optional[str] = None
@@ -165,7 +166,7 @@ class FieldDesc:
     body: Optional[Tag] = None
 
     def format(self) -> Tag:
-        formatted: Tag = tags.transparent if self.body is None else self.body
+        formatted = self.body or self._UNDOCUMENTED
         if self.type is not None:
             formatted = tags.transparent(formatted, ' (type: ', self.type, ')')
         return formatted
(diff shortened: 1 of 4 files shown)


pydoctor/epydoc2stan.py
```diff
@@ -305,8 +312,35 @@ def handle_param(self, field: Field) -> None:
...
+                    # Constructor parameters can be documented on the class.
+                    return
+            field.report('Documented parameter "%s" does not exist' % (name,))
     handle_arg = handle_param
-    handle_keyword = handle_param
+
+    def handle_keyword(self, field: Field) -> None:
+        name = self._handle_param_name(field)
+        if name is not None:
+            # TODO: How should this be matched to the type annotation?
```
mthuurne, line 340: There are two functions in Twisted that use `@keyword`, but neither of them have type annotations at the moment, so I think it's fine to resolve this TODO later.

adiroiban: Do we want to support `@keyword` ... can't we just use `@param` ? :)

I see that in the previous code keyword was the same as param... so maybe deprecated `keyword` and just use `@param` everywhere.

pydoctor/test/test_epydoc2stan.py
```diff
@@ -128,6 +159,83 @@ def f(a):
...
+        def __init__(self):
+            super().__init__(1)
+    ''')
+    epydoc2stan.format_docstring(mod.contents['Base'].contents['__init__'])
+    assert capsys.readouterr().out == ''
+    epydoc2stan.format_docstring(mod.contents['Sub'].contents['__init__'])
+    assert capsys.readouterr().out == ''
+
+def test_func_keyword(capsys: CapSys) -> None:
+    """Test handling of @keyword."""
```
adiroiban, line 227:
```suggestion
def test_func_parameter_as_keyword(capsys: CapSys) -> None:
    """A warning is raised when a parameter is documented as a @keyword."""
```

pydoctor/test/test_epydoc2stan.py
```diff
@@ -128,6 +159,83 @@ def f(a):
...
+    ''')
+    epydoc2stan.format_docstring(mod.contents['f'])
+    captured = capsys.readouterr().out
+    assert captured == (
+        '<test>:4: Documented parameter "x" does not exist\n'
+        '<test>:6: Documented parameter "y" does not exist\n'
+        )
+
+def test_func_arg_not_inherited(capsys: CapSys) -> None:
+    """Do not warn about non-existing parameters from inherited docstrings."""
```
adiroiban, line 209:
```suggestion
    """No warning is emitted when a subclass is implementing an inherited function with different parameters."""
```

pydoctor/epydoc2stan.py
mthuurne: The annotation would be useful only if all keywords have the same type: if for example one keyword is a `str` and another keyword is an `int`, the annotation will be `object` or `Any` or `Union[int, str]`, which isn't very helpful to the end user. So the author will have to use `type` fields to document the types of the individual keywords.

pydoctor/test/test_epydoc2stan.py
```diff
@@ -128,6 +159,83 @@ def f(a):
     classic_fmt = docstring2html(classic_mod.contents['f'])
     assert annotation_fmt == classic_fmt
 
+def test_func_param_duplicate(capsys: CapSys) -> None:
+    """Warn when the same parameter is documented more than once."""
```
adiroiban, line 163: Do we already have a test when the same parameter has multiple `@type` definitions?

pydoctor/test/test_epydoc2stan.py
```diff
@@ -59,6 +59,37 @@ def test_html_empty_module() -> None:
...
+        pass
+    ''')
+    func = mod.contents['nop']
+    lines = docstring2html(func).split('\n')
+    assert '<td class="fieldName">Returns</td>' not in lines
+
+
+def test_func_undocumented_return_something() -> None:
+    """When the returned value is undocumented and its type is not None,
+    include a the "Returns" entry in the output.
```
adiroiban, line 78: I don't understand this docstring.... 

So if the return type is not None and is not documented, it is added as an undocumented span.

I guess that it handles this case https://pydoctor--296.org.readthedocs.build/en/296/api/pydoctor.astbuilder.ASTBuilder.html#parseFile

Can we also include `<tr class="fieldStart"><td class="fieldName">Returns</td><td colspan="2">` in the expected output?

adiroiban (approved):
It looks good. Only minor comments.

It would have been nice to have the sample API doc enabled so that we could see in HTML in a browser all the cased for there is a new test

For now, I have used https://pydoctor--296.org.readthedocs.build/en/296/api/.

pydoctor/epydoc2stan.py
mthuurne: > Do we want to support `@keyword` ... can't we just use `@param` ? :)
> 
> I see that in the previous code keyword was the same as param... so maybe deprecated `keyword` and just use `@param` everywhere.

The purpose of `@keyword` is to document keywords that can be captured by the `**kwargs` parameter. I think it's useful to have a separate field for those, since otherwise we wouldn't be able to warn about `@param` being used with a non-existing parameter name. That warning led to several documentation fixes in Twisted.

pydoctor/test/test_epydoc2stan.py
mthuurne: Yes, that is the case that it handles. I'll add a bit more detail to the docstring, in the hope that it will be easier to understand.

I added both `<td>` tags in the expected output, that indeed makes the test easier to read. However, I left out the `<tr>`, since that is not really part of what is being tested here and adding more output increases the chance that the test will require maintenance on unrelated changes.

Issue #290, opened 2020-11-25
Release 2020.12.0 tracking


mthuurne:
While there is only one week left in 2020.11, I want to try and get a pydoctor release out soon, so let's stick to that as a target version number for now.

Things that need to be done:
- parameter annotations: #136, #217, #229 and #283
- #255, since this includes a deprecation
- #233?
- #153?
- check whether deprecations are documented correctly
- update release notes
- automatically release on tag push (we had this ability in Travis, but Travis itself became unworkable)

While there are 4 issues related to parameter annotations, it's not as much work as it may seem:
- #136 doesn't require a PR, only the closing of other tickets
- #229 could be implemented for parameter fields only for now
- #283 could be limited to what we need to support pos-only and kw-only params, without major changes in presentation
- I already have a local branch for #217, but it was stranded several times as other issues had to be resolved first (like #283)

~~I have some doubt about whether deprecations are documented correctly, since there is Twisted-specific code for documenting deprecations in `pydoctor.templatewriter.pages.functionchild` that isn't executed when generating Twisted's API docs. It's possible this is obsolete code, but it could also mean that some deprecations are currently omitted from the output.~~ Also Twisted now supports [parameter deprecations](https://twistedmatrix.com/trac/ticket/9844) and I haven't seen any code that handles documentation generation for that.
 
I'd like to hear opinions on other tickets that really need to go into the next release, or tickets that are listed here but aren't crucial. Also help on resolving the to-do items is welcome of course.

mthuurne:
I was looking at the wrong coverage report (for the unit tests); when generating Twisted's API docs the deprecation code is executed.

adiroiban:
A lot of tickets here.... I think that this will be 2020.12.0 ... or maybe even 2021

mthuurne:
Some tickets take multiple days, while others can be solved in an hour. For the ones with question marks, I want to give them a go, but if complications turn up, we can postpone them.

The parameter annotations is the item I really wanted in this release and that item is almost finished.

adiroiban:
ok. ping me on IRC if you need any help. I can help with the reviews...but I don't have the motivation and time  to work on new feature

mthuurne:
Here is a draft for the release notes:

- Python 3.6 or higher is required
- There is now a user manual which can be built with Sphinx or read online on Read The Docs; this is a work in progress and the online version will be updated between releases
- Added support for Python language features:
  * Type annotations of function parameters and return value are used when the docstring does not document a type
[...]

PR #297, opened 2020-11-27
Add a Publish pydoctor to PyPI workflow

tristanlatr:
automatically release pydoctor to pypi on tag push as mentioned in #290 .

You just need to configure PyPI to accept the `__token__` user and store the token as github secret: ${{ secrets.PYPI_TOKEN }} .

Cheers,


.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,38 @@
...
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@master
+
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
```
adiroiban, line 17: can we go with 3.8 or 3.9 ?

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,38 @@
...
+  push:
+    tags: 
+      - '*'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@master
```
adiroiban, line 12: can we pin a specific checkout version? ... maybe in the future master will introduce backward incompatible changes

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,38 @@
...
+          python -c "import os; [print(f'{k}={v}') for k, v in os.environ.items()]"
+
+      - name: Install deps
+        run: |
+          python -m pip install --upgrade pip setuptools wheel
+          python -m pip install -r requirements.txt
+
+      - name: Build pydoctor
+        run: python3 setup.py --quiet build check sdist bdist_wheel
+
```
adiroiban, line 33: After build, I like to have a Run `ls -al dist/` as a simple way to debug what was created.

adiroiban (approved):
It looks good and it's a good start.

Rather than having it as a separate workflow that runs only on tag, I prefer to have it as a job on the main workflow that is executed on each PR.
Is just that the last publish step is skipped if we don't have a tag.

Like we have for ldapor there https://github.com/twisted/ldaptor/blob/master/.github/workflows/main.yml#L102

With running this workflow only on tags, there is a big risk of having someting broken inside the workflow and then we will find out about this error only when the tag / release is created.
Which is a bit too late.

So, if @mthuurne is happy, I am ok to merge it as it is, but I would prefer to have the Publish task executed for each PR.

Thanks again for this PR.

tristanlatr:
> I would prefer to have the Publish task executed for each PR

I can make that happen

tristanlatr:
There is an issue to use the "needs" config in the yaml file. It seems that all jobs needs to be in the same file (we cannot cross reference job names) if we want to use them as dependencies to other jobs. 

So  should I move everything in a main.yaml like `ldaptor` do ?

adiroiban:
> So should I move everything in a main.yaml like `ldaptor` do ?

You can  only depend on jobs from the same workflow.
So yes, if you want to run the release code on each PR and block the release if the tests fail, you need to move it to main.yaml

adiroiban:
Maybe we can have publish only inside `unit.yaml` ... and maybe have something shorter than `Main CI workflow for pydoctor` ... maybe just `CI` ?

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,43 @@
+name: Publish pydoctor to PyPI on tags
```
adiroiban, line 1: I think that this file should be removed.

.github/workflows/unit.yaml
```diff
@@ -45,3 +45,39 @@ jobs:
...
+        run: |
+          python setup.py --quiet build check sdist bdist_wheel
+          ls -alh ./dist/
+
+      - name: Publish pydoctor to PyPI on tags
+        if: startsWith(github.ref, 'refs/tags')
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_TOKEN }}
```
adiroiban, line 83: I have created a PYPI token from my username with the scope of pydoctor.

.github/workflows/pypi.yaml
tristanlatr: Sorry that was an error

.github/workflows/unit.yaml
```diff
@@ -1,11 +1,11 @@
-name: Unit tests
+name: CI
 
 on:
   push:
     branches: [ master ]
```
adiroiban, line 5: I think that we also need to add the tags here... as otherwise this workflow will not be triggered  for tags
```
    tags:
      - v*
```

tristanlatr: Ok, I think works
```
tags:
      - '*'
```

.github/workflows/unit.yaml
```diff
@@ -79,5 +79,5 @@ jobs:
         if: startsWith(github.ref, 'refs/tags')
         uses: pypa/gh-action-pypi-publish@master
         with:
-          user: __token__
+          user: adiroiban
```
adiroiban, line 84: I think that you should have `__token__` here.
I have created an API token.
I have not saved my password :)

.github/workflows/unit.yaml
```diff
@@ -1,11 +1,13 @@
...
 
 on:
   push:
     branches: [ master ]
+    tags: 
+      - '*'
   pull_request:
     branches: [ master ]
-
+
```
mthuurne, line 10: I think the reason `git` (the command line version, anyway) marks excess whitespace in red is to avoid situations like this, where there is no real change but the line is recorded as having been changed.

mthuurne (approved):
I'm still relatively new to GitHub Actions, but when I read the config it seems to make sense.

.github/workflows/unit.yaml
adiroiban, line 10:
```suggestion

```

.github/workflows/unit.yaml
```diff
@@ -1,11 +1,13 @@
-name: Unit tests
+name: CI
```
adiroiban, line 1:
```suggestion
name: Unit tests and release
```

.github/workflows/unit.yaml
```diff
@@ -1,11 +1,12 @@
...
 
 on:
   push:
     branches: [ master ]
+    tags: 
+      - '*'
   pull_request:
     branches: [ master ]
-
 jobs:
```
adiroiban, line 10:
```suggestion

jobs:
```

.github/workflows/unit.yaml
```diff
@@ -45,3 +47,39 @@ jobs:
       continue-on-error: true
       run: |
         tox -e codecov
+
+  publish:
```
adiroiban, line 51:
```suggestion
  release:
```

adiroiban:
I have updated the release documentation.

mthuurne:
Thanks! This was one of the release blockers.

PR #299, opened 2020-11-29
HTML templates

tristanlatr:
Hello,

This PR tries to solve #3, it introduces a new argument: `--template-dir`.
The users can overwrite any template files contained in the `pydoctor/templates`. 

Custom place holders have been created: `header.html`, `subheader.html` and a CSS sheet: `extra.css`. Overwriting those won't create any warnings on upgrades since they are empty by default. Actual template files have a version number ( independent from pydoctor's ). Overwriting those is supported but might trigger warnings for missing information if the default template version has increase. (So Updating pydoctor doesn't mean all your custom templates will be outdated)

**This feature will break (but not crash) the Twisted custom build. Twisted should use the template system to insert custom HTML and JS into the pages.**

New errors or warnings: 

- Pydoctor is politely aborting if a template is too new. 
- Pydoctor is warning if a template version cannot be read or the version is outdated.
With the current implementation it's hard to distinguish a placeholder template (that is not supposed to have a version) and a normal HTML template with invalid version tag...

Structural changes: 

- All pages inherits from [`Page`](https://pydoctor--299.org.readthedocs.build/en/299/api/pydoctor.templatewriter.pages.Page.html) to avoid duplicating the rendering of : `Nav`, header.html, footer.html and pageHeader.html.
- All elements based on a Template file inherits from [`TemplateElement`](https://pydoctor--299.org.readthedocs.build/en/299/api/pydoctor.templatewriter.TemplateElement.html) that is created by passing a ITemplateLoader to the constructor. 
- The [`TemplateLookup`](https://pydoctor--299.org.readthedocs.build/en/299/api/pydoctor.templatewriter.TemplateLookup.html) object is core of the templates handling system logic. 
- The [`Template`](https://pydoctor--299.org.readthedocs.build/en/299/api/pydoctor.templatewriter.Template.html) object represent any file in `pydoctor/templates` or custom templates files. They are identified by their filenames. 
The Template object is abstract class.
Template.fromfile is the factory to the concrete templates. For now concrete templates are _HtmlTemplate or _StaticTemplate. 
- The template writer inherits from [`IWriter`](https://pydoctor--299.org.readthedocs.build/en/299/api/pydoctor.templatewriter.IWriter.html) that define the 3 methods. 

Style changes:
-  add a new CSS class: `navlinks`: List of links to add in the nav header. 
-  Now the project home link is rendering the same way on all pages nicely.
- I've also fixed a few issue regarding HTML generation: #121 and made the HTML readable on smaller screens.


mthuurne:
These are just some first impressions, I'll do a more thorough review later.

Thanks for picking this up. I haven't thought about customization in detail yet, but this is certainly a big improvement over monkeypatching.

We should define and document a backwards compatibility strategy. Even if we can't actually solve that problem, having a way to warn "your templates were written for pydoctor _oldversion_; these things need updating:" would be much better than crashing or silently producing wrong output when using outdated templates.

I think it's best if we merge this after the next release, both because of timing and because of compatibility. I have several ideas on presentation that I want to try out after releasing and some of those might break templates; releasing template customization and then breaking most templates in the release after will not make users happy.

> To do so, I used a singleton for the TemplateFileLookup object.

Over the years I've become convinced that singleton is an anti-pattern. It often starts out simple, but it will cause problems later which by then are hard to solve. I should probably do a blog post on it some day. Is it possible to do this without a singleton, via `System.options` for example?

> I've also written a simple header for pydoctor and twisted, as a demo

This is very useful in evaluating the changes. But by the time we're ready to merge, the Twisted templates should be moved to Twisted's repo.

In a quick look over the code, I saw some parts use `str` for file paths, some use `Path` and some use `FilePath`. I think `Path` offers the nicest interface, so I'd prefer to use that as the primary type for passing file paths.

In situations where we're dealing with paths coming from the source being analyzed and we want to be careful, `FilePath` might be a better choice. But I don't think that applies to templates, since those are selected by someone who already has enough privileges to construct command lines.

tristanlatr:
> I think it's best if we merge this after the next release

I agree

> having a way to warn "your templates were written for pydoctor oldversion; these things need updating:" would be much better 

Should we store the pydoctor version in every template i.e. `<meta name="generator" content="pydoctor 20.12.0" />` and have a script punch it ?

>  Is it possible to do this without a singleton, via System.options for example?

Probably! But I think this involves using another way than the static method `templatefile()` to find the templates. 
I can write something that will use System.options to create a the System's TemplateFileLookup. But it means that the twisted build have to be changed too.

tristanlatr:
> I think Path offers the nicest interface, so I'd prefer to use that as the primary type for passing file paths.

Twisted templates XMLString warns if we pass something other than a FilePath, and it's also the interface used in pydoctor

mthuurne:
Maybe the "generated by" message should be in `footer.html`, so people can customize that by for example adding their project's logo and navigation links.

tristanlatr:
There two things here:
- header.html, pageHeader.html and footer.html are place holders for the users to overwrite. They are currently empty by default. This is why users would be able to add custom html and css in those files and upgrade to newer version of pydoctor without risk of breaking stuff.
- if the users want to customize the actual templates for the twisted templating system, they can. BUT it's where we should be careful about the compatibility of the templates. The twisted tenplating system throws error if they are unfilled spots. So if we add more spots to fill, we need to make them compatible with template that don't have the correct tags (fillable spots). This is where the meta genetator tags can maybe help, comparing the version, warning, and removing the tags that ate unimplemented. So I think the tags should appear in all templates (except the place holders) . Or maybe we should think of another manner to do the templatinf system.

tristanlatr:
It's  ready for first round of reviews, I've updated the main comment with new changes

mthuurne:
I think CI is failing because GitHub it is using the changes merged in #314, but this PR does not contain those changes yet.

tristanlatr:
Ive merged the changes from master after this PR was merged... so may be it's just a question if time

tristanlatr:
Tell me if you like the CSS adjustments I've made in the custom template (`docs/pydoctor_templates/pydoctor/header.html`) to move to the default apidocs.css or to disappear ;)

pydoctor/templatewriter/writer.py
```diff
@@ -95,3 +116,21 @@ def _writeDocsForOne(self, ob, fobj):
         self.written_pages += 1
         ob.system.progress('html', self.written_pages, self.total_pages, 'pages written')
         flattenToFile(fobj, page)
+
+    def _checkTemplatesV(self) -> None:
```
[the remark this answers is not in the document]
mthuurne: It seems we have two strategies for handling template directories:
- this method will scan the directory for templates early
- pages load templates on demand

On-demand loading has some downsides:
- broken templates will only be reported when HTML writing starts, which can be quite some time after startup on large code bases
- we seem to be doing a lot of repeated loading of the same templates

I think it would be useful to introduce something like a `Template` and `TemplateCollection` class, where the latter reflects the contents of one specific templates directory. `TemplateFileLookup` could then be renamed to `TemplateLookup` and perform the lookup in one or more`TemplateCollection` objects.

A `TemplateCollection` would parse all its templates on creation, so any errors are reported as soon as possible. And it would keep the `Template` objects, so we don't have to parse the same files multiple times.

pydoctor/test/test_templatewriter.py
```diff
@@ -119,6 +120,53 @@ def test_multipleInheritanceNewClass(className: str) -> None:
     assert "methodA" in html
     assert "methodB" in html
 
+def test_templatefile_lookup() -> None:
+    
+    lookup = TemplateFileLookup()
+
+    here:Path = Path(__file__).parent
```
mthuurne, line 127:
```suggestion
    here = Path(__file__).parent
```

mthuurne (changes requested):
Overall, I think you're on the right track, but perhaps staying a bit too close to the current approach and therefore not able to solve some of the problems that the templating system has.

One such problem is that we do a lot of repeated work. While `XMLFile` caches the loaded template, that caching is not used if we construct a new `XMLFile` object for every output document. The new version check adds another repeated loading and parsing of the same template files.

Another problem is that the code customization is complex and fragile. I don't think this problem is solved with the new `IWriter` class. Maybe we don't need to solve this issue at all though: Twisted is the only project that I'm aware of that uses this mechanism and as discussed in #315, the customizations that Twisted does should be moved into pydoctor or replaced by new pydoctor features.

docs/source/pydoctor_templates/twisted/header.html
```diff
@@ -0,0 +1,32 @@
+<div>
+    <style>
+        #banner {
+            background: #444;
+            background-image: -webkit-linear-gradient(bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, .3));
+            background-image: -moz-linear-gradient(bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, .3));
+            background-image: -ms-linear-gradient(bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, .3));
+            background-image: -o-linear-gradient(bottom, rgba(0, 0, 0, 0), rgba(0, 0, 0, .3));
```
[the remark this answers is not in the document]
tristanlatr: > not your mess to clean up

Yep

I'll delete it all together

docs/source/pydoctor_templates/pydoctor/nav.html
```diff
@@ -0,0 +1,40 @@
...
+                Module Index
+            </a> </li>
+
+            <li>  <a href="classIndex.html">
+                Class Index
+            </a> </li>
+
+            <li>  <a href="nameIndex.html">
+                Name Index
+            </a> </li>
```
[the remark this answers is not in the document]
tristanlatr: Should those links always be there by default ? right now it's only in our custom template. 

Shorter version it will be :)

docs/source/pydoctor_templates/pydoctor/header.html
```diff
@@ -0,0 +1,24 @@
...
+            bottom: 0;
+            left: 0;
+            padding: 5px;
+            opacity: 0.7;
+            margin-bottom: 40px;
+            background-color: #efefef;
+            border-radius: 1px;
+        }
+
+    </style>
```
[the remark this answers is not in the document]
tristanlatr: It is to include it in one if the custom templates. There is no empty spot for CSS template right now, so we have to include it there. 
It's only to show you, should I integrate it to the default CSS ?

docs/source/usage.rst
```diff
@@ -264,23 +282,14 @@ This is also used by the Twisted project to handle deprecation.
...
-~~~~~~~~~~~~~~~~~~~~~~~~
+Custom writer
+~~~~~~~~~~~~~
 
 You can subclass the :py:class:`pydoctor:pydoctor.templatewriter.writer.TemplateWriter` and pass your custom class dotted name with the following argument::
 
-  --html-writer=mylib._pydoctor.CustomTemplateWriter
 
-.. note:: Not fully documented, prone to break
+  --writer-class=mylib._pydoctor.CustomTemplateWriter
```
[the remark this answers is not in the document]
tristanlatr: My point behind the `--writer-class=` rename is to prepare for a plaintext writer, then a RST or any kind of writers.

Note that the old option is still there in the code.

pydoctor/driver.py
```diff
@@ -382,16 +392,29 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
             from pydoctor import templatewriter
             if options.htmlwriter:
                 writerclass = findClassFromDottedName(
-                    options.htmlwriter, '--html-writer',
-                    templatewriter.TemplateWriter)
+                    options.htmlwriter, '--writer-class', IWriter) # type: ignore
```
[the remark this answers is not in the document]
tristanlatr: It was something about instantiation of abstract classes

pydoctor/templates/attribute-child.html
```diff
@@ -1,4 +1,5 @@
 <div xmlns:t="http://twistedmatrix.com/ns/twisted.web.template/0.1" class="function">
+  <meta name="template" content="pydoctor-default" major_version="1" minor_version="0" />
```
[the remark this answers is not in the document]
tristanlatr: If we add new template tag in the twisted templates, they won't be filled (missing information) if the custom template is outdated. 

> Do we need a major and minor version 

Probably not.

> `<meta name="pydoctor-template-version" content="1" />`

This will work too

pydoctor/templates/common.html
```diff
@@ -2,22 +2,19 @@
 <html xmlns:t="http://twistedmatrix.com/ns/twisted.web.template/0.1" t:render="all">
   <head>
     <title><t:slot name="title">Something</t:slot> : API documentation</title>
-
+    <meta name="generator" content="pydoctor" version="20.12.0" />
```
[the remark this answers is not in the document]
tristanlatr: I'll remove alltogether the pydoctor version for now

pydoctor/templatewriter/__init__.py
```diff
@@ -1,6 +1,6 @@
 """Render pydoctor data as HTML."""
 
-DOCTYPE = b'''\
+DOCTYPE:bytes = b'''\
```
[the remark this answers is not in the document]
tristanlatr: It's for mypy

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -41,17 +42,78 @@ def get(self, ob, summary=False):
...
+
+    @renderer
+    def nav(self, request, tag):
+        return Nav(self.system)
+
+    @renderer
+    def header(self, request, tag):
+        template = self.system.templatefile_lookup.get_templatefilepath('header.html')
+        if template.getsize()>0:
+            return html2stan(template.open('r').read().decode())
```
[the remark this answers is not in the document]
tristanlatr: Good question... 
I thought XMLFile was only for renderring stuff with twisted template but I guess we could use it to dump normal HTML as a stan tree.

pydoctor/templatewriter/writer.py
```diff
@@ -95,3 +116,21 @@ def _writeDocsForOne(self, ob, fobj):
...
+        """
+        Issue warnings when custom templates are outdated. 
+        """
```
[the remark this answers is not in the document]
tristanlatr: Yes indeed I did not saw this option at first.
I think we should just politely crash pydoctor if a template that are too new is encountered. 
Because the chances are if the version was bumped it's because a new slot was added, and if so, it will crash at the rendering time.
It should be clear in the docs that the users should not increase the template versions themselves.

pydoctor/model.py
```diff
@@ -568,6 +568,10 @@ def __init__(self, options: Optional[Values] = None):
         self.buildtime = datetime.datetime.now()
         self.intersphinx = SphinxInventory(logger=self.msg)
 
+        from pydoctor.templatewriter.util import TemplateFileLookup
+        self.templatefile_lookup = TemplateFileLookup()
+        """System's L{TemplateFileLookup} object"""
```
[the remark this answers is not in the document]
tristanlatr: Moved it to the writer class

pydoctor/templatewriter/pages/__init__.py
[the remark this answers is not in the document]
tristanlatr: I realized that using `html2stan` makes sense because it adds the surrounding div that avoid an error when we don't in the template it self.

docs/source/usage.rst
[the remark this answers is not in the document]
tristanlatr: I've removed the`--writer-class` synonym since the RST output will probably come with a `--make-rst` option that will select another writer class.

tristanlatr (commented):
So I've been working on this for a few days and it's ready for another review (right after I fix the tests lol). The implementation is much more serious, but there is obviously room for improvment: For instance i'm still using bs4 to parse the HTML version, I guess we could retrieve the value from the STAN tree but I don't even know what STAN means and since the meta tag can be nested, I'de just prefer use a library to parse the tree. 

- Pydoctor is politely crashing if a template is too new (with a RuntimeError)
- The XML files are not re-created all the time: they are stored as the 'renderable' property of the Template object. 
- It does not compare default version to itself anymore
- The Template object is abstract and can be later subclassed to introduce RST output
- Template.fromfile is the factory to the concrete templates
- Pydoctor is only warning if a template version cannot be read or the version is outdated. 
  With the current implementation it's hard to distinguish a placeholder template (that is not supposed to have a version) and a normal HTML template with invalid version tag...
- The TemplateLookup do not do a lookup in several collection of Template object but rather save the Template in a dict for o(1) access with the filename.  
- I've also fixed a few issue regarding HTML generation: #121 and made the HTML readable on smaller screens. 

Look forward to read your feedbacks.

mthuurne:
I ran some quick tests; I'll do a code review later.

> For instance i'm still using bs4 to parse the HTML version, I guess we could retrieve the value from the STAN tree but I don't even know what STAN means and since the meta tag can be nested, I'de just prefer use a library to parse the tree.

I would prefer if we could avoid that dependency, especially since we already have the ability to parse HTML in both the standard library and in Twisted's template mechanism. I'll look into how that could be done.

I think technically Stan is the syntax `twisted.web.template` uses for generating an HTML tree that can be flattened to HTML text. The tree doesn't have a typical DOM API; it's mostly nested lists. But it can probably be introspected, especially if what we're looking for is simple.

> * Pydoctor is politely crashing if a template is too new (with a RuntimeError)

That's certainly better than an incomprehensible crash, but ideally it would log an error message and exit with exit code 1, without showing a backtrace.

> * The XML files are not re-created all the time: they are stored as the 'renderable' property of the Template object.

Nice. Did you measure whether this makes a significant difference in performance?

pydoctor is much quicker when it doesn't write output, but it's not clear to me yet whether that was caused by the repeated template loading, by the template mechanism being slow in general, by processing being done on demand that should perhaps be done earlier, or by some inefficiency we haven't discovered yet.

> * It does not compare default version to itself anymore

Good.

> * The Template object is abstract and can be later subclassed to introduce RST output

That would require more than template changes, wouldn't it? Since `epydoc2stan` can only output HTML and does essential formatting as well.

I think the "native" data for Sphinx is a document tree using types from `docutils.nodes`, is that correct? If so, it might make sense to use that as the internal representation inside pydoctor and at the end convert the tree to HTML in pydoctor itself or pass it to Sphinx for rendering as part of a larger document.

> * Template.fromfile is the factory to the concrete templates
> * Pydoctor is only warning if a template version cannot be read or the version is outdated.
>   With the current implementation it's hard to distinguish a placeholder template (that is not supposed to have a version) and a normal HTML template with invalid version tag...
> * The TemplateLookup do not do a lookup in several collection of Template object but rather save the Template in a dict for o(1) access with the filename.

I can't really comment on this yet without having read the code.

> * I've also fixed a few issue regarding HTML generation: #121 and made the HTML readable on smaller screens.

Having the type next to the parameter name is an improvement, I think. Instead of parentheses, maybe we could use a colon, like Python itself. So `url: str` instead of `url (str)`. I think that would be easier to read, as well as a bit more compact.

Perhaps we could move from a 3-column layout to a 2-column one, by making "Parameters", "Returns" and "Raises" mini headings. I'll try to explain it in ASCII art.

Current:
```
Parameters  |  x (int)   | x-coordinate
            |  y (int)   | y-coordinate
Returns     |  int       | distance to origin when traveling over an integer grid
```
Alternative:
```
Parameters
  x: int   | x-coordinate
  y: int   | y-coordinate
Returns
  int      | distance to origin when traveling over an integer grid
```
This would allow us to make the name + type column wider if necessary, which would reduce the number of times the type would have to be line wrapped.

> Look forward to read your feedbacks.

I like the Modules, Classes and Names buttons, but is it possible to baseline-align (vertically) their label text with the "Project API Documentation" header?

Now that we have the Modules, Classes and Names buttons, maybe the front page could have the root package as the first option? It is the option I pick most often, since the root package often contains a docstring describing the project as a whole.

Alternatively, maybe we don't need a front page at all and we could immediately redirect to the root package, unless we have more than one root (is that even properly supported?). But that should probably be something to discuss in a separate PR.

mthuurne:
Oh, I forgot to add a comment about the failing test: it seems something was changed in an incompatible way, breaking Twisted's customizations. Since we use Twisted as an important part of our system test, it would be really nice if we'd have a workaround for this.

mthuurne:
Another CSS tweak we could consider is putting the parameter name and its type in separate flexboxes, to force line wrapping to happen between name and type before the type starts wrapping.

tristanlatr:
> Did you measure whether this makes a significant difference in performance?

No not really, but the cpython apidocs still takes about 5mn to build on my Mac. 

> That would require more than template changes, wouldn't it? Since epydoc2stan can only output HTML and does essential formatting as well.

Yes  you are right, we would need to reorganize epydoc2stan to be able to output some kind of data structure that will include all typing informations and other goodies added by epydoc2stan in a form that understandable (docutils.node is a great idea). 
But it will also require new templates for RST files. 
 
> Perhaps we could move from a 3-column layout to a 2-column

Yes I think it makes sens too

> Alternatively, maybe we don't need a front page at all and we could immediately redirect to the root package, unless we have more than one root

Yes totally, like epydoc style, where the front page is the docs of the root folder + the module index on a same page. 

> is that even properly supported?

Yes it looks like so, and it even might be useful for some. 

But the auto redirect would be great for single root systems. 

> It seems something was changed in an incompatible way, breaking Twisted's customizations

Yes it will break Twisted customizations at least. I can mark the old function as deprecated and keep them arround: this will avoid Twisted build to crash to it will still break it. The customizations won't show up.

mthuurne:
The mypy error can be fixed by adding `if not TYPE_CHECKING:` around the second import, similar to how it's handled in `pydoctor/__init__.py`.

tristanlatr:
I don't get it... 
I'm trying to replace bs4 dependency with Twisted microdom module but I think it's buggy somehow... 

```
>>> from twisted.web.microdom import parseXMLString
>>> parseXMLString("""<table xmlns:t="http://twistedmatrix.com/ns/twisted.web.template/0.1" class="children sortable">
  <meta name="pydoctor-template-version" content="1" />
  <t:attr name="id" t:render="id" />
  <tr t:render="rows">
    <t:attr name="class" t:render="class_" />
    <td t:render="kind">
      Kind
    </td>
    <td t:render="name">
      Name
    </td>
    <td t:render="summaryDoc">
      Summary
    </td>
  </tr>
</table>""")
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/site-packages/twisted/web/microdom.py", line 1102, in parseXMLString
    return parseString(st, caseInsensitive=0, preserveCase=1)
  File "/usr/local/lib/python3.9/site-packages/twisted/web/microdom.py", line 1085, in parseString
    return parse(BytesIO(st.encode('UTF-16')), *args, **kw)
  File "/usr/local/lib/python3.9/site-packages/twisted/web/microdom.py", line 1052, in parse
    mdp.dataReceived(readable.getvalue())
  File "/usr/local/lib/python3.9/site-packages/twisted/web/sux.py", line 179, in dataReceived
    newState = doFn(byte)
  File "/usr/local/lib/python3.9/site-packages/twisted/web/sux.py", line 486, in do_afterslash
    self.gotTagStart(self.tagName, self.tagAttributes)
  File "/usr/local/lib/python3.9/site-packages/twisted/web/microdom.py", line 917, in gotTagStart
    for k, v in attributes.items():
RuntimeError: dictionary changed size during iteration
```

mthuurne:
> I'm trying to replace bs4 dependency with Twisted microdom module but I think it's buggy somehow...

I can reproduce the problem. It does look like a bug: when applying a namespace, the `attributes` dictionary is modified while it is being iterated over. The code takes care to not delete items inside the loop, but it does add items inside the loop, which is not allowed either.

mthuurne:
`xml.dom.minidom` from the standard library might be an alternative. It has a `getElementsByTagName()` method which can be used to quickly find the `<meta>` tag.

tristanlatr:
It does not work either, I get another error...

```
xml.parsers.expat.ExpatError: junk after document element: line 4, column 4
````

I think the parsers don't like the tags like `<tr t:render="rows">`

**Comment by mthuurne** [comment]
Strange; it worked for me when I tried the example you pasted earlier.

Maybe it differs per Python version? (3.8.6 on Linux here) Or maybe it matters that I have `lxml` installed?

Did you use `xml.dom.minidom.parseString()` or a different approach?

**Comment by tristanlatr** [comment]
It seems that the file `pydoctor/test/testcustomtemplates/faketemplate/table.html` is causing issues.. 

I have lxml installed and python 3.9 on Mac.
I'm out of idea.
This si so strange, why would it fail parsing the `<span>` tag ?

```
>>> import lxml
>>> from xml.dom.minidom import parseString
>>> parseString("""<meta name="pydoctor-template-version" content="1.1" />
<span>
    Random words
</span>""")
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python@3.9/3.9.0_5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/dom/minidom.py", line 1998, in parseString
    return expatbuilder.parseString(string)
  File "/usr/local/Cellar/python@3.9/3.9.0_5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/dom/expatbuilder.py", line 925, in parseString
    return builder.parseString(string)
  File "/usr/local/Cellar/python@3.9/3.9.0_5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/xml/dom/expatbuilder.py", line 223, in parseString
    parser.Parse(string, True)
xml.parsers.expat.ExpatError: junk after document element: line 2, column 0
```

**Comment by tristanlatr** [comment]
@mthuurne

**Comment by mthuurne** [comment]
An XML document can only have one root element, so `<meta>` followed by `<span>` is indeed incorrect. It should work though if you wrap both of them into a single root element.

**Comment by tristanlatr** [comment]
So I've made some additional changes such as
- updating the docs
- use standard lib XML parser
- adopt the same behaviour as sphinx regarding unknown fields. I.e. just display the name of the fields without warning about it. We currently do not support al the bibliographical fields even if they are defined in ``docsutils``. The reason is that we overwrite docutils processing of the fields. So either we list all the fields in the FieldHandler class or we choose to consider any field name valid and only apply special processing on special names.
The second option seems to need a lot less maintenance.

I'll hang out in the IRC so we can chat more efficiently.

File: pydoctor/templatewriter/pages/__init__.py [hunk 14]
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+        return epydoc2stan.type2stan(ob)
+
+class BaseElement(Element, abc.ABC):
+    """
+    Common base HTML element. 
+    """
+    def __init__(self, 
+        system:Optional[model.System]=None, 
+        template_lookup:Optional[TemplateLookup]=None, 
+        loader:Optional[ITemplateLoader]=None, ) -> None:
```
mthuurne, line 50:
```suggestion
        system: Optional[model.System] = None, 
        template_lookup: Optional[TemplateLookup] = None, 
        loader: Optional[ITemplateLoader] = None
        ) -> None:
```

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+        self.system = system
+        self.template_lookup = template_lookup
+        
+        if loader:
+            pass
+        # filename attribute should be set for most page classes
+        elif self.filename:
+            if self.template_lookup:
+                template = self.template_lookup.get_template(
+                    self.filename)
```
mthuurne, line 72:
```suggestion
                template = self.template_lookup.get_template(self.filename)
```

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+
+        The C{loader} property is usually got from the L{TemplateLookup} 
+        object but can also be set by argument.
+
+        @raises RuntimeError: If not enought information is provided 
+            to create the template loader. i.e. missing C{template_lookup} or C{loader} argument.
+            Could also be that the L{Template.renderable} property is not a L{ITemplateLoader} provider. 
+
+        @note: C{system} and C{template_lookup} can be none in special cases, like for L{LetterElement}. 
+        """
```
mthuurne, line 62: Can we simplify this? The `template_lookup` is used for acquire a loader, so maybe `__init__` should always take a loader as its argument and there could be a classmethod that takes a `template_lookup` argument, acquires the loader from that and then instantiates the class using the loader.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+        if loader:
+            pass
+        # filename attribute should be set for most page classes
+        elif self.filename:
+            if self.template_lookup:
+                template = self.template_lookup.get_template(
+                    self.filename)
+                if not zope.interface.verify.verifyObject(ITemplateLoader, template.renderable):
+                    raise RuntimeError(f"Cannot create HTML element {self} because template renderable "
+                                       f"property is not a ITemplateLoader provider: {type(template.renderable)}")
```
mthuurne, line 75: Generally `TypeError` would be the exception to raise if an argument doesn't implement the expected interface.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+        super().__init__(loader)
+
+    @abc.abstractproperty
+    def filename(self) -> str:
+        """
+        Associated output filename. 
+        
+        Can be empty string in special cases (like L{LetterElement}). 
+        """
+        pass
```
mthuurne, line 93: Maybe `LetterElement` shouldn't inherit from `BaseElement` if it doesn't implement the full interface. Is there any part of `BaseElement` that it does depend on?

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+        else:
+            return tags.span(self.system.projectname)
+
+class BasePage(BaseElement):
+    """
+    Base page element. 
+
+    Defines special HTML placeholders that are designed to be overriden by users: 
+    "header.html", "pageHeader.html" and "footer.html".
+    """
```
mthuurne, line 115: Since this is a new class, I'd like to have type annotations on it.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+
+    filename = 'common.html'
     ob: model.Documentable
 
-    def __init__(self, ob, docgetter=None):
+    def __init__(self, ob:model.Documentable, template_lookup:TemplateLookup, docgetter:Optional[DocGetter]=None):
+        super().__init__(ob.system, template_lookup)
         self.ob = ob
-        if docgetter is None:
+        if docgetter == None:
```
mthuurne, line 145: Why this change? `is None` is a standard Python idiom.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -78,7 +170,7 @@ def namespace(self, obj: model.Documentable) -> Sequence[Union[Tag, str]]:
         while ob:
             if ob.documentation_location is model.DocLocation.OWN_PAGE:
                 if parts:
-                    parts.append('.')
+                    parts.extend(['.', tags.wbr])
```
mthuurne, line 173: Is this necessary? Since the module names are links, those `<a>` tags are already points at which the browser can decide to wrap the string.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -137,9 +223,10 @@ def baseTables(self, request, tag):
         return ()
 
     def mainTable(self):
+        from pydoctor.templatewriter.pages.table import ChildTable
```
mthuurne, line 226: This avoids a circular import because `ChildTable` inherits from `BaseElement`, right? Maybe `BaseElement` shouldn't be in `pydoctor.templatewriter.pages` then, but in for example `pydoctor.templatewriter` or in a new module.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -419,3 +511,21 @@ def functionExtras(self, data):
                 )))
         r.extend(super().functionExtras(data))
         return r
+        
+# Type alias for HTML page: to use in type checking. 
+# It's necessary because L{BasePage} doesn't have the same C{__init__} siganture as concrete pages classes.
```
mthuurne, line 516: All the listed classes inherit from `CommonPage` and that class is also in the list, so maybe using `CommonPage` instead of `AnyDocPage` would be sufficient?

pydoctor/templatewriter/summary.py
```diff
@@ -160,9 +167,12 @@ def heading(self, request, tag):
         return tag.clear()("Class Hierarchy")
 
 
-class LetterElement(Element):
+class LetterElement(BaseElement):
+
+    filename = ''
```
mthuurne, line 172: Maybe this is another hint that `LetterElement` shouldn't actually be a `BaseElement`.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+        object but can also be set by argument.
+
+        @raises RuntimeError: If not enought information is provided 
+            to create the template loader. i.e. missing C{template_lookup} or C{loader} argument.
+            Could also be that the L{Template.renderable} property is not a L{ITemplateLoader} provider. 
+
+        @note: C{system} and C{template_lookup} can be none in special cases, like for L{LetterElement}. 
+        """
+        self.system = system
+        self.template_lookup = template_lookup
```
mthuurne, line 64: The superclass `__init__` is not called; that's probably a bug.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
         else:
-            doc = epydoc2stan.format_docstring(ob)
-            typ = epydoc2stan.type2stan(ob)
-            if typ is None:
-                return doc
+            return epydoc2stan.format_docstring(ob)
+    def get_type(self, ob: model.Documentable) -> Optional[Tag]:
+        return epydoc2stan.type2stan(ob)
+
+class BaseElement(Element, abc.ABC):
```
mthuurne, line 43: In general, `BaseSomething` is a superclass of `Something` instead of the other way around. See for example `BaseException`.

What makes this different from other elements? I guess it has something to do with templates, so maybe include that in the name instead.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,34 +31,125 @@ def signature(function: model.Function) -> str:
...
+                                    "object is passed to BaseElement's 'template_lookup' init argument.")
+        else:
+            raise RuntimeError(f"Cannot create HTML element {self} because no ITemplateLoader "
+                                "object is passed to BaseElement's 'loader' init argument and 'filename' property is not set.")
+        super().__init__(loader)
+
+    @abc.abstractproperty
+    def filename(self) -> str:
+        """
+        Associated output filename.
```
mthuurne, line 89: Is "output" correct? I'd expect pages to have an output filename, but not elements. Also it seems this is called to get the template filename.

mthuurne (changes requested):
I haven't reviewed every single file yet, but the comment counter is now over 100, so I figure there is plenty to look at.

Most remarks are minor things, like incorrect whitespace and spelling mistakes.

The type annotations can be simplified in some cases, I think. There was also some new code that was missing annotations. And disabling mandatory annotations for `astbuilder` is not an option.

There are also some interfaces that could be simplified. A few of them were already present in pydoctor and either lost their relevance or were preparations for features that were never implemented. Others are in new code and are mostly cases of being more flexible than we need to be: optional arguments that could be mandatory, different argument combinations that could be split into separate methods instead.

I already said this in private, but I want to have it on the record as well, in case other people read this later: for future PRs, please split them up as soon as the scope expands beyond one central change. A PR of this size is not easy to review, nor is it easy to get into shape for merging.

tristanlatr (commented):
Hi,

I've solve most of the issue you've raised. Thanks. 

The tests are still a bit light and not using the coolest `pytest` features like you suggested yet. I'll work a bit more on testing but the code is ready for another review round.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+    @overload
+    def __init__(self, htmloutput: str) -> None: ...
+    @overload
+    def __init__(self, htmloutput: str, template_lookup: 'TemplateLookup') -> None: ...
+
+    def prepOutputDirectory(self) -> None:
+        """
+        Called first.
+        """
+        ...
```
mthuurne, line 57: The docstring already counts as the mandatory statement for the function body, so the `...` is not needed here. Removing it would avoid it being counted as not-covered.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+    @overload
+    def __init__(self, htmloutput: str, template_lookup: 'TemplateLookup') -> None: ...
+
+    def prepOutputDirectory(self) -> None:
+        """
+        Called first.
+        """
+        ... 
+
+    def writeModuleIndex(self, system:'System') -> None:
```
mthuurne, line 59:
```suggestion
    def writeModuleIndex(self, system: System) -> None: 
```

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+        """
+        ... 
+
+    def writeModuleIndex(self, system:'System') -> None: 
+        """
+        Called second.
+        """
+        ...
+
+    def writeIndividualFiles(self, obs:Iterable['Documentable']) -> None:
```
mthuurne, line 65:
```suggestion
    def writeIndividualFiles(self, obs: Iterable[Documentable]) -> None:
```

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+def parse_dom(text: str) -> minidom.Document:
+    """
+    Create a L{minidom} representaton of the XML string. 
+    """
+    try:
+        dom = minidom.parseString(text)
+    except Exception as e:
+        raise ValueError(f"Can't parse XML from text '{text}'. XML documents can have only one root. ") from e
+    else:
+        return dom
```
mthuurne, line 36: If we don't need to do anything with the DOM object, the `return` statement can just be inside the `try:` clause.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+                version_str = meta.getAttribute("content")
+                if version_str:
+                    try:
+                        version = int(version_str)
+                    except ValueError:
+                        warnings.warn(f"Could not read '{template_name}' template version: "
+                                "the 'content' attribute must be an integer")
+                else:
+                    warnings.warn(f"Could not read '{template_name}' template version: "
+                        f"the 'content' attribute is empty")
```
mthuurne, line 215: Since `int('')` raises `ValueError`, I think we don't actually need to test for empty strings separately.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+        # it's most probably a placeholder template. 
+        version = -1
+        for meta in dom.getElementsByTagName("meta"):
+            if meta.getAttribute("name") != "pydoctor-template-version":
+                continue
+            if meta.hasAttribute("content"):
+                version_str = meta.getAttribute("content")
+                if version_str:
+                    try:
+                        version = int(version_str)
```
mthuurne, line 205: We could speed things up by returning immediately if we have found a valid version.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+
+    @staticmethod
+    def _extract_version(dom: minidom.Document, template_name: str) -> int:
+        # If no meta pydoctor-template-version tag found, 
+        # it's most probably a placeholder template. 
+        version = -1
+        for meta in dom.getElementsByTagName("meta"):
+            if meta.getAttribute("name") != "pydoctor-template-version":
+                continue
+            if meta.hasAttribute("content"):
```
mthuurne, line 201: We could handle the absence first:
```py
if not meta.hasAttribute("content"):
    warnings.warn(...)
    continue
```
I think that would help readability, since the check and the warning message are closer together then.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,297 @@
...
+    _default_template_dir = 'templates'
+
+    def __init__(self) -> None:
+        """Init L{TemplateLookup} with templates in C{pydoctor/templates}"""
+        # Dict comprehension to init templates to whats in pydoctor/templates
+        self._templates: Dict[str, Template] = { t.name:t for t in (Template.fromfile(f) for f in 
+                Path(__file__).parent.parent.joinpath(
+                    self._default_template_dir).iterdir()) if t }
+        
+        self._default_templates = copy.deepcopy(self._templates)
```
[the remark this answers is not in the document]
mthuurne: I think `Template` objects get all their data on construction and then never change again, so a shallow copy should be enough.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+
+        Warns if the custom template is designed for an older version of pydoctor. 
+
+        @raises UnsupportedTemplateVersion: If the custom template is designed for a newer version of pydoctor. 
+        """
+        
+        try:
+            default_version = self._default_templates[template.name].version
+        except KeyError:
+            warnings.warn(f"Invalid template filename '{template.name}' (will be ignored). Valid filenames are: {list(self._templates)}")
```
mthuurne, line 266: What does "will be ignored" mean exactly: that the template will be ignored or that the invalid name will be ignored? It seems ignoring the invalid name implemented, since the template object is stored in the `finally:` clause, but perhaps ignoring the template would be better.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+        if not template_lookup:
+            template_lookup = TemplateLookup()
+        return template_lookup.get_template(cls.filename) # type: ignore[arg-type]
+
+    @abc.abstractproperty
+    def filename(self) -> str:
+        """
+        Associated template filename. 
+        """
+        pass
```
mthuurne, line 353: The coverage notice can be avoided by either removing the `pass` statement or using `raise NotImplementedError`. Not sure which is better here, but I guess both work, assuming mypy knows about `abc.abstractproperty`. If not, `raise` would be better.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+            assert loader is not None
+            return loader
+        else:
+            raise TypeError(f"Cannot create loader because '{cls}.filename' property is not set.")
+
+    @classmethod
+    def template(cls, template_lookup: Optional[TemplateLookup] = None) -> Template:
+        # error: Argument 1 to "get_template" of "TemplateLookup" has incompatible type
+        # "Callable[[TemplateElement], str]"; expected "str"
+        # We can ignore mypy error because filename is an abstractproperty.
```
mthuurne, line 343: Is mypy really wrong here? You are calling `filename` as if it is a class method, but it's a property instead.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+        else:
+            raise TypeError(f"Cannot create loader because '{cls}.filename' property is not set.")
+
+    @classmethod
+    def template(cls, template_lookup: Optional[TemplateLookup] = None) -> Template:
+        # error: Argument 1 to "get_template" of "TemplateLookup" has incompatible type
+        # "Callable[[TemplateElement], str]"; expected "str"
+        # We can ignore mypy error because filename is an abstractproperty. 
+        if not template_lookup:
+            template_lookup = TemplateLookup()
```
mthuurne, line 345: Could we use a single default `TemplateLookup` object instead of constructing a new one every time? The initialization of `TemplateLookup` looks relatively expensive.

mthuurne (changes requested):
I haven't reviewed everything yet, but I figure that quick feedback is more useful to you, so here is what I have so far.

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+        if not path.is_file():
+            warnings.warn(f"Cannot create Template: {path.as_posix()} is not a file.")
+        elif path.suffix.lower() in cls.TEMPLATE_FILES_SUFFIX:
+
+            if path.suffix.lower() == '.html':
+                return _HtmlTemplate(path=path)
+            else:
+                return _StaticTemplate(path=path)
+        else:
+            warnings.warn(f"Cannot create Template: {path.as_posix()} is not a template file.")
```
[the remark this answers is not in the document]
tristanlatr: The users can drop `.docx` files inside their custom template folder, this warning takes care of this situation.

pydoctor/templatewriter/__init__.py
tristanlatr: This warning takes care of situation where a user drops a `random.html` file inside their template folder and expect the file to appear in the output folder. We make it clear thet it doesn't work like that. 

What is your demand exactly ?

pydoctor/templatewriter/__init__.py
tristanlatr: I'll go with `raise`, thanks

pydoctor/templatewriter/__init__.py
tristanlatr: It's a class attribute declared in the super class as a `abstractproterty` so we make sure no object is instanciated without a `filename` attribute

pydoctor/templatewriter/__init__.py
tristanlatr: The optional value is only used in tests

tristanlatr: I could remove it all together

pydoctor/templatewriter/__init__.py
```diff
@@ -5,6 +5,364 @@
...
+        self.path: Path = path
+        """
+        Template file path
+        """
+        
+        with path.open('r') as f:
+            self.text = f.read()
+            """
+            File text
+            """
```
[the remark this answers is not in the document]
tristanlatr: I'm not sure to see the point, but I can do it. Would this mean that the path argument is optional ?The simpler the better no ?

pydoctor/templatewriter/__init__.py
tristanlatr: The argument is not optional anymore

pydoctor/templatewriter/__init__.py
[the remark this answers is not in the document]
mthuurne: `__init__()` wouldn't need the path argument at all if it's passed text + name. The advantage of having a constructor that does as little work as possible is that you are more flexible in the way you acquire the data for it. This is useful in particular when writing tests, but it also in other code.

And if all that `__init__()` does is store its arguments, we can use `attrs` to auto-generate the `__init__()` and `__repr__()` implementations from annotated fields.

pydoctor/templatewriter/__init__.py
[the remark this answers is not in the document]
mthuurne: Ah, then I didn't understand the mechanism correctly.

So it works like this:
- `TemplateLoader.__init__()` scans the template directory and creates `Template` objects for all files
- `TemplateLoader.get_template()` looks up a name in the dictionary of all templates (2 layers: custom and default)

I thought that `TemplateLoader` would instead fill its dictionary on demand, in which case unsupported file names would never be requested.

I think that the current implementation is actually better, because with on-demand loading the error reporting is less predictable and happens later in the execution. So let's stick with the current approach and keep this warning.

I do think it's useful to mention in the warning message that the file name extension is the reason the file is rejected as a template (as opposed to its contents).

pydoctor/templatewriter/__init__.py
mthuurne: I think it's a bit confusing that the template we're not going to use is stored in `self._templates` nevertheless. It is harmless, but it would be better to not store it, I think.

tristanlatr: Yes your're right

pydoctor/templatewriter/__init__.py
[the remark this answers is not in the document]
tristanlatr: The warning will be as follow :
```
warnings.warn(f"Cannot create Template: {path.as_posix()} is not recognized as template file. "
                f"Template files must have one of the following extensions: {', '.join(cls.TEMPLATE_FILES_SUFFIX)}")
```

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -168,7 +254,8 @@ def functionBody(self, data):
     @renderer
     def all(self, request, tag):
         return tag.fillSlots(
-            title=self.title(),
+            project=self.system.projectname,
+            # title=self.title(),
```
mthuurne, line 258:
```suggestion
```

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -343,7 +432,8 @@ def baseTables(self, request, item):
         return [item.clone().fillSlots(
                           baseName=self.baseName(b),
                           baseTable=ChildTable(self.docgetter, self.ob,
-                                               sorted(attrs, key=lambda o:-o.privacyClass.value)))
+                                               sorted(attrs, key=lambda o:-o.privacyClass.value), 
+                                                ChildTable.lookup_loader(self.template_lookup)))
```
mthuurne, line 436: The template lookup could be moved outside of the loop, to avoid doing redundant work.

pydoctor/templatewriter/summary.py
```diff
@@ -28,24 +29,23 @@ def moduleSummary(modorpack, page_url):
...
 
+class ModuleIndexPage(BasePage):
 
-class ModuleIndexPage(Element):
     filename = 'moduleIndex.html'
 
-    @property
-    def loader(self):
-        return XMLFile(util.templatefilepath('summary.html'))
+    def __init__(self, system: model.System, template_lookup: TemplateLookup, ):
```
mthuurne, line 36:
```suggestion
    def __init__(self, system: model.System, template_lookup: TemplateLookup):
```

pydoctor/templatewriter/summary.py
```diff
@@ -116,19 +116,21 @@ def subclassesFrom(hostsystem, cls, anchors, page_url):
...
+class ClassIndexPage(BasePage):
+
     filename = 'classIndex.html'
 
-    @property
-    def loader(self):
-        return XMLFile(util.templatefilepath('summary.html'))
+    def __init__(self, 
+        system:model.System, 
+        template_lookup:TemplateLookup, ):
```
mthuurne, line 125:
```suggestion
    def __init__(self, system: model.System, template_lookup: TemplateLookup):
```

pydoctor/templatewriter/summary.py
```diff
@@ -328,19 +313,18 @@ def hasdocstring(ob):
...
+    
     filename = 'undoccedSummary.html'
 
-    @property
-    def loader(self):
-        return XMLFile(util.templatefilepath('summary.html'))
-
-    def __init__(self, system):
-        self.system = system
+    def __init__(self, system:model.System, template_lookup:TemplateLookup, ):
```
mthuurne, line 320:
```suggestion
    def __init__(self, system: model.System, template_lookup: TemplateLookup):
```

pydoctor/templatewriter/util.py
```diff
@@ -1,19 +1,18 @@
...
-def srclink(o: model.Documentable) -> Optional[str]:
+def srclink(o: Documentable) -> Optional[str]:
     return o.sourceHref
 
-def templatefile(filename):
-    abspath = os.path.abspath(__file__)
-    pydoctordir = os.path.dirname(os.path.dirname(abspath))
-    return os.path.join(pydoctordir, 'templates', filename)
+def templatefile(filename: str) -> str:
+    """Deprecated"""
```
mthuurne, line 13:
```suggestion
    """Deprecated: can be removed once Twisted stops using this."""
```

pydoctor/templatewriter/util.py
```diff
@@ -1,19 +1,18 @@
...
+    return TemplateLookup().get_template(filename).path.as_posix()
 
-def templatefilepath(filename):
+def templatefilepath(filename:str) -> FilePath:
+    """Deprecated"""
```
mthuurne, line 17:
```suggestion
    """Deprecated: can be removed once Twisted stops using this."""
```

pydoctor/templatewriter/writer.py
```diff
@@ -1,28 +1,35 @@
...
     fobj.write(DOCTYPE)
-    err = []
-    def e(r):
-        err.append(r.value)
-    flattenString(None, page).addCallback(fobj.write).addErrback(e)
+    err: List[Failure] = []
+    flattenString(None, page).addCallback(fobj.write).addErrback(err.append)
     if err:
-        raise err[0]
+        raise err.pop().value
```
mthuurne, line 25: This will raise the last error instead of the first. Why this change in behavior?

pydoctor/templatewriter/writer.py
```diff
@@ -31,67 +38,86 @@ def __subclasshook__(cls, subclass: Type[object]) -> bool:
...
+            self.base.joinpath('apidocs.css'))
+        shutil.copy(
+            self.template_lookup.get_template('extra.css').path,
+            self.base.joinpath('extra.css'))
+        shutil.copy(
+            self.template_lookup.get_template('bootstrap.min.css').path,
+            self.base.joinpath('bootstrap.min.css'))
+        shutil.copy(
+            self.template_lookup.get_template('pydoctor.js').path,
+            self.base.joinpath('pydoctor.js'))
```
mthuurne, line 70: Instead of having the names hardcoded here, could we use the static template names from the default template?

pydoctor/templatewriter/writer.py
```diff
@@ -31,67 +38,86 @@ def __subclasshook__(cls, subclass: Type[object]) -> bool:
...
 
-    def writeIndividualFiles(self, obs):
+    def writeIndividualFiles(self, obs:Iterable[model.Documentable]) -> None:
+        """
+        Iterate trought C{obs} and call L{_writeDocsFor} method for each L{Documentable}.
```
mthuurne, line 74:
```suggestion
        Iterate through C{obs} and call L{_writeDocsFor} method for each L{Documentable}. 
```

pydoctor/templatewriter/writer.py
```diff
@@ -31,67 +38,86 @@ def __subclasshook__(cls, subclass: Type[object]) -> bool:
...
-                path = FilePath(self.base).child(f'{ob.fullName()}.html')
-                with path.open('wb') as out:
-                    self._writeDocsForOne(ob, out)
+                with self.base.joinpath(f'{ob.fullName()}.html').open('wb') as fobj:
+                    self._writeDocsForOne(ob, fobj)
         for o in ob.contents.values():
             self._writeDocsFor(o)
 
-    def _writeDocsForOne(self, ob, fobj):
+    def _writeDocsForOne(self, ob:model.Documentable, fobj:IO[bytes]) -> None:
```
mthuurne, line 105: Thanks for adding all the new type annotations.

pydoctor/test/test_astbuilder.py
```diff
@@ -1210,16 +1210,16 @@ def type2html(obj: model.Documentable) -> str:
     assert type2html(d) == '<code>str</code>'
     e = C.contents['e']
     assert e.docstring == """fifth"""
-    assert type2html(e) == '<code>List[C]</code>'
+    assert type2html(e).replace('<wbr></wbr>', '').replace('<wbr>\n</wbr>', '') == '<code>List[C]</code>'
```
mthuurne, line 1213: Since `type2html()` is a test helper function, you could put the `<wbr>` removal there.

pydoctor/test/test_epydoc2stan.py
```diff
@@ -171,8 +177,8 @@ def f(a, b):
         @rtype: C{bool}
         """
     ''')
-    annotation_fmt = docstring2html(annotation_mod.contents['f'])
-    classic_fmt = docstring2html(classic_mod.contents['f'])
+    annotation_fmt = docstring2html(annotation_mod.contents['f']).replace('<wbr></wbr>', '').replace('<wbr>\n</wbr>', '')
```
mthuurne, line 180: Like for `type2html()`, the `<wbr>` removal could be added to `docstring2html()` instead of having to repeat it in the test cases.

pydoctor/test/test_templatewriter.py
```diff
@@ -119,6 +120,106 @@ def test_multipleInheritanceNewClass(className: str) -> None:
     assert "methodA" in html
     assert "methodB" in html
 
+def test_html_template_version() -> None:
+    lookup = TemplateLookup()
+    for _, template in lookup._templates.items():
```
mthuurne, line 125:
```suggestion
    for template in lookup._templates.values():
```

mthuurne (changes requested):
I reviewed all files now. I think we're almost there.

pydoctor/templatewriter/writer.py
tristanlatr: Did not meant to change the behaviour. 
`raise err.pop(0).value` looks better then ?

pydoctor/templatewriter/pages/__init__.py
tristanlatr: The `work` here is only call `TemplateLookup._templates.__getitem__` with `ChildTable.filename` that should be O(1).

pydoctor/templatewriter/writer.py
tristanlatr: You're welcome :)

pydoctor/templatewriter/writer.py
tristanlatr: I could do a `TemplateLookup.itertemplates()`, then test the type so all `_StaticTemplate` would be copied to build directory by default ?

pydoctor/templatewriter/__init__.py
```diff
@@ -1,10 +1,346 @@
...
+        HTML Templates should have a version identifier as follow::
+    
+            <meta name="pydoctor-template-version" content="1" />
+
+        This is always C{-1} for CSS and JS templates. 
+        """
+        raise NotImplementedError()
+
+    @abc.abstractproperty
+    def loader(self) -> Optional[ITemplateLoader]:
```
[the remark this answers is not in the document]
tristanlatr: The `TemplateElement.lookup_loader` class method take care of that. 

There is no `assert isinstance(template, _HtmlTemplate)` statement.

pydoctor/templatewriter/__init__.py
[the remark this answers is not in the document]
tristanlatr: Done, except I'm not using `attrs`

docs/source/pydoctor_templates/pydoctor/header.html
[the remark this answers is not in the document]
tristanlatr: The `extra.css` placeholder template takes care of that use case.

pydoctor/templatewriter/pages/__init__.py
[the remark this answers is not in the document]
tristanlatr: The implementation uses `XMLString`, multi-root XML are not supported in template files.

pydoctor/templatewriter/summary.py
```diff
@@ -328,19 +311,18 @@ def hasdocstring(ob):
             return True
     return False
 
-class UndocumentedSummaryPage(Element):
+class UndocumentedSummaryPage(BasePage):
+    
     filename = 'undoccedSummary.html'
```
mthuurne, line 316: This output file is generated, but nothing links there. It's fine to leave it as-is for this PR, but we should think about whether this page is useful.

It does look slightly broken at the moment:
- the public/private filter doesn't work: private names are always shown
- inherited methods without a docstring are listed, even if the method they inherit is documented

pydoctor/templatewriter/util.py
```diff
@@ -1,19 +1,22 @@
...
 
-def templatefile(filename):
-    abspath = os.path.abspath(__file__)
-    pydoctordir = os.path.dirname(os.path.dirname(abspath))
-    return os.path.join(pydoctordir, 'templates', filename)
+def templatefile(filename: str) -> None:
+    """Deprecated: can be removed once Twisted stops using this."""
+    warnings.warn("pydoctor.templatewriter.util.templatefile() "
+        "is deprecated and returns None. It will be remove in future versions. "
+        "Please use the templating system.")
```
mthuurne, line 14: I don't see this warning being printed when I generate Twisted's docs. Is it possible it is disabled by default, like most Python warnings?

mthuurne (commented):
Here are a few comments. I'll finish the review later, maybe not today though.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -30,21 +31,113 @@ def signature(function: model.Function) -> str:
...
+    def nav(self, request: IRequest, tag: Tag) -> IRenderable:
+        return Nav(self.system, Nav.lookup_loader(self.template_lookup), )
+
+    @renderer
+    def header(self, request: IRequest, tag: Tag) -> Union[IRenderable, Tag]:
+        # Tag is a IRenderable but mypy is not smart enough.
+
+        template = self.template_lookup.get_template('header.html')
+        assert template.loader is not None
+        return template.loader.load() if not template.is_empty() else Tag('transparent')
```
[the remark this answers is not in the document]
tristanlatr: Yes you're right, this is a reminiscence of older implementation.

pydoctor/templatewriter/__init__.py
```diff
@@ -1,10 +1,349 @@
...
+@runtime_checkable
+class IWriter(Protocol):
+    """
+    Interface class for pydoctor output writer. 
+    """
+
+    @overload
+    def __init__(self, htmloutput: str) -> None: ...
+    @overload
+    def __init__(self, htmloutput: str, template_lookup: 'TemplateLookup') -> None: ...
```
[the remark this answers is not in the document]
tristanlatr: I think we still need this for mypy no ?

pydoctor/templatewriter/__init__.py
```diff
@@ -1,10 +1,349 @@
...
+        """
+        Init a new element. 
+
+        @raises TypeError: If C{loader} is not a L{ITemplateLoader} provider. 
+        """
+        try:
+            verify.verifyObject(ITemplateLoader, loader)
+        except Exception as e:
+            raise TypeError(f"Cannot create HTML element {self} because template loader "
+                            f"is not a ITemplateLoader provider: {type(loader)}") from e
```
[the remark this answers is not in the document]
tristanlatr: Mypy do not always understand `zope.interface` as seen with `Tag` objects. 
So honestly I don't know.

pydoctor/templatewriter/util.py
tristanlatr: I think this functions do not get called at all anymore.

pydoctor/templatewriter/__init__.py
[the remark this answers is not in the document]
tristanlatr: I can delete the check, I think you're right. And it was untested code.

pydoctor/templatewriter/__init__.py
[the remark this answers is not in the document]
tristanlatr: mypy doesn't like it without:

```
pydoctor/driver.py:431: error: Only concrete class can be given where "Type[IWriter]" is expected  [misc]
                        options.htmlwriter, '--html-writer', IWriter) 
                                                             ^
pydoctor/driver.py:451: error: unused 'type: ignore' comment
                        writer = writerclass(options.htmloutput, # type: ignore[abstract]
                        ^
pydoctor/driver.py:451: error: Too many arguments for "IWriter"  [call-arg]
                        writer = writerclass(options.htmloutput, # type: ignore[abstract]
                                 ^
pydoctor/driver.py:451: error: Unexpected keyword argument "template_lookup" for "IWriter"  [call-arg]
                        writer = writerclass(options.htmloutput, # type: ignore[abstract]
                                 ^
pydoctor/driver.py:455: error: unused 'type: ignore' comment
                        writer = writerclass(options.htmloutput) # type: ignore[abstract]
                        ^
pydoctor/driver.py:455: error: Too many arguments for "IWriter"  [call-arg]
                        writer = writerclass(options.htmloutput) # type: ignore[abstract]
                                 ^
pydoctor/driver.py:459: error: unused 'type: ignore' comment
                    writer = writerclass(options.htmloutput) # type: ignore[abstract]
                    ^
pydoctor/driver.py:459: error: Too many arguments for "IWriter"  [call-arg]
                    writer = writerclass(options.htmloutput) # type: ignore[abstract]
```

tristanlatr (commented):
Looks ready to be merged! @mthuurne

tristanlatr:
Hey @mthuurne , do you still want to add more commits to this PR ?

mthuurne (commented):
> Hey @mthuurne , do you still want to add more commits to this PR ?

Yes, I want to remove as many differences between the different types of pages as possible. I already did a bit for the slots, but there are more slots and renderers that we could make available on all pages. Also using the same footer on every page would be nice (as we discussed in chat).

I have some time available today and tomorrow, so expect new commits soon.

pydoctor/templatewriter/__init__.py
```diff
@@ -105,14 +100,15 @@ def fromfile(cls, path: Path) -> Optional['Template']:
         """
         if path.suffix.lower() in cls.TEMPLATE_FILES_SUFFIX:
             try:
-                with path.open('r') as fobj:
+                with path.open('r', encoding='utf-8') as fobj:
                     text = fobj.read()
+            except IOError as e:
+                warnings.warn(f"Cannot create Template: {path.as_posix()}. IO error: {e}")
```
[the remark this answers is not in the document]
mthuurne: Hmm, I guess this was an old review comment that GItHub stored somewhere. I remember typing it, but not today.

tristanlatr:
Hey, 

I saw that you removed the special class for the project home button. The fact is that it was not useless. It's effectively not the same color as the bootstrap default: it was slightly blue and would indicate the users that the link would exit the current site. But that's not a big deal. 

Also for the footer I'll add some minor CSS changes, you tell me what you think

mthuurne:
> I saw that you removed the special id for the project home button. The fact is that it was not useless. It's effectively not the same color as the bootstrap default: it was slightly blue and would indicate the users that the link would exit the current site. But that's not a big deal.

You are correct. It seems I checked the hover color that Bootstrap assigns to the other links in the header, which is equal to the hover color the project link had. But since the project link is inside a `<span class="navbar-brand">`, the hover color is overridden by Bootstrap to be gray instead of blue.

Shall I override the link hover color for `navbar-brand` then? Because using an `id` will create invalid HTML if the project link is on the page more than once, which is now the case with the default template. And even if the default template wouldn't use the link twice, we'd still want to allow user templates to do that, I think.

> Also for the footer I'll add some minor CSS changes, you'll tell me what you think

I'll have a look (nothing seems to have been pushed yet).

tristanlatr:
I've done a few minor edits.

For the link color, I've used a css class, which solves the issue while keeping a different hover color for the project home link in the navbar. 

I hope that the PR can now be merged!

tristanlatr:
Hello @mthuurne,

Don't want to be pushy but this PR looks good to me. If you don't like the CSS changes please revert them. 
This PR needs to be merged to continue the work on #362 #358 and #370.

Thanks :-)

mthuurne:
There are probably still things that can be improved, but it's already a lot better than the old template code and other development work cannot easily use the new template system until it's merged. So I merged it now.

Thanks for your hard work and your patience.

Issue #302, opened 2020-12-01
Warnings that are not logged can fail the build


mthuurne:
Currently, whether warnings are displayed depends on the log verbosity: the warning is always issued internally, but if the severity isn't high enough, it isn't printed on stdout. As a result, if you pass `--warnings-as-errors`, it can happen that pydoctor exits with exit code 3 while no warnings can be seen in the logging (only a message like "40/40 modules processed 11 warnings"). I think this is not a good user experience.

The cause of the inconsistency seems to be that `System.msg()` considers a log event a warning if it has a negative threshold, while `System._warning()` prints warnings at threshold 0 but unconditionally adds them to the `warnings` dictionary.

The full set of callers of `_warning()` is:
- `System.handleDuplicate()`: "duplicate"
- `ModuleVistor.visit_ImportFrom()`: "processing import statement in odd context"
- `ModuleVistor.visit_ImportFrom()`: "relative import level too high"
- `ModuleVistor._importAll()`: "import * from unknown"
- `ModuleVistor._importAll()`: "import *"
- `ModuleVistor._importNames()`: "cannot find re-exported name"
- `ModuleVistor.visit_Import()`: "processing import statement in odd context"
- `ModuleVistor.parseFile()`: "cannot parse"

None of these are specific to documentation errors: they indicate that pydoctor is not able to make sense of the code it is analyzing, which could be caused by problems in the code or by limitations or bugs of pydoctor. So it would make sense to remove the storing of warnings from the `_warning()` method and perhaps also rename it.

adiroiban:
So.... the big problem with the logger is that  DEBUG, WARNING, INFO (etc) are just "log levels" and not types/tags/classes.

We need to find a good name for these "warnings" to make it clear that they are not just a log level, but they also have extra semantic: it is a log message based on which we expect end users to act and change their source code.

I see that in flake8 they are called "violations" - https://flake8.pycqa.org/en/latest/glossary.html#term-violation


Maybe we can rename them `check_warnings` ? 

And in system have `System.reportCheckWarning()` or `System.reportCheckError()` or `System.reportCheckViolation()`

If we refactor this, I would also like to add a code to each violation type. It can be a simple code, without warning or error class... just 3 zero padded digits.

adiroiban:
In terms of pydoctor progress  /  activity reporting I think that we should have 3 types of messages:

* error - this is a message that will cause pydoctor to fail right away and pydoctor build process can ingore it and recover from it
[...]

PR #306, opened 2020-12-01
Do not count calls to 'System._warning()' as violations

mthuurne:
See #302 for details.

### DIFF
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -508,7 +508,7 @@ class System:
     def __init__(self, options: Optional[Values] = None):
         self.allobjects: Dict[str, Documentable] = {}
         self.rootobjects: List[Documentable] = []
-        self.warnings: Dict[str, List[Tuple[str, str]]] = {}
+        self.violations: Dict[str, List[Tuple[str, str]]] = {}
         self.packages: List[str] = []
 
         if options:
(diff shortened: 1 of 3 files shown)


pydoctor/model.py
```diff
@@ -508,7 +508,7 @@ class System:
     def __init__(self, options: Optional[Values] = None):
         self.allobjects: Dict[str, Documentable] = {}
         self.rootobjects: List[Documentable] = []
-        self.warnings: Dict[str, List[Tuple[str, str]]] = {}
+        self.violations: Dict[str, List[Tuple[str, str]]] = {}
```
adiroiban, line 511: maybe add a docstring with some info about what `violations` are.

I am still not sure how to define them, and what is the difference between `violations` and the remaining `_warnings`.

I would say that a warning like `cannot parse` should end up with a build failure...and not just be ignored.

adiroiban (approved):
It looks good. Thanks!

I think we can merge it as it is now and later see how to resolve #302  and see which of the remaking warnings are `violations` or `errors`.

I would still like to have pydoctor fail on a non-clean build... and a warning like `cannot parse` should mark the build as "dirty" :)

adiroiban (approved):
Nice :)

Thanks!

mthuurne:
> I think we can merge it as it is now and later see how to resolve #302 and see which of the remaking warnings are `violations` or `errors`.

Yes, we need to re-do the logging/reporting system at some point, but this should be good enough for making `--warnings-as-errors` usable in the coming release.

PR #307, opened 2020-12-02
Enable Intersphinx cache by default

### LINKED ISSUE #300: Enable intersphinx cache by default?
We support caching intersphinx inventories, but this requires passing the `--enable-intersphinx-cache` option on the command line. I think it would make more sense to enable the cache by default and provide an option to disable it.

The default cache duration is only one day, so I don't think there is a big risk of working with stale data if we enable caching by default.

mthuurne:
Fixes #300.

### DIFF
--- a/pydoctor/driver.py
+++ b/pydoctor/driver.py
@@ -178,8 +178,8 @@ def getparser() -> OptionParser:
         '--intersphinx', action='append', dest='intersphinx',
         metavar='URL_TO_OBJECTS.INV', default=[],
         help=(
-            "Use Sphinx objects inventory to generate links to external"
-            "documetation. Can be repeated."))
+            "Use Sphinx objects inventory to generate links to external "
+            "documentation. Can be repeated."))
 
     parser.add_option(
         '--enable-intersphinx-cache',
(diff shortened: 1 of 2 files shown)


pydoctor/driver.py
```diff
@@ -251,6 +258,10 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
     if options.configfile:
         readConfigFile(options)
 
+    if options.enable_intersphinx_cache_deprecated:
+        print("The --enable-intersphinx-cache option is deprecated; "
+              "the cache is now enabled by default.", file=sys.stderr)
```
adiroiban, line 263: maybe log it as a deprecation warning, instead of a simple print... the idea that a deprecation warning should be picked by an automated test run.... but since this is CLI usage... I don't know. Maybe the stderr print will also do the job.

```suggestion
        warnings.warn("The --enable-intersphinx-cache option is deprecated; "
              "the cache is now enabled by default.", DeprecationWarning)
```

adiroiban (approved):
Looks good. Thanks!

Only a minor comment...and I am not sure that my comment is valid.

pydoctor/driver.py
mthuurne: `DeprecationWarning` is intended to inform developers that they're using deprecated API without bothering the user about it (since warnings are disabled by default). But since this is a CLI option that is deprecated, we should be alerting the user to update their scripts.

PR #308, opened 2020-12-02
DOC: Fix RST Consolidated Fields reference

tristanlatr:
This removes part of of the doc copy/pasted from epydoc that are not working in PyDoctor / Should not be used.
- ``:except:``: Was just an error, it was changed to ``:raises:`` in PyDoctor
- ``:group:``: unkown fields
- ``:keyword:``: Same as parameters
- ``:Types:``: Type hints should be used


mthuurne:
Currently pydoctor does not recognize all synonyms such as `except` that epydoc does, because Twisted uses a specific subset. But to make it easier for other projects to move from epydoc to pydoctor, we could consider adding those synonyms.

The `type` field is still useful, both for projects that aren't using type annotations and for cases where the type annotation syntax cannot accurately express the type and the author wants to manually provide a more meaningful description. For that reason, if a `type` field exists, it overrides the annotation in the output.

The `keyword` field is used to document keywords that can be passed to `**kwargs` arguments, so it has a different role than the `param` field. You're [not the only one](https://github.com/twisted/pydoctor/pull/296#discussion_r531704014) for whom the `keyword` field was unclear (and I didn't pick up on the difference either until I started working on that part of the code), so it would be worth documenting. Until that PR, it was possible to (ab)use `keyword` to document parameters, but now pydoctor issues a warning for that.

The `group` field is something that we don't support and don't have plans for supporting, so removing that is good.

adiroiban:
> For that reason, if a `type` field exists, it overrides the annotation in the output.

+1



> Until that PR, it was possible to (ab)use `keyword` to document parameters, but now pydoctor issues a warning for that.

Thanks for that work :)
 
----------

> The `group` field is something that we don't support and don't have plans for supporting, so removing that is good.

+1 keep it simple :)

tristanlatr:
Changed to only remove ``:except ValueError:`` and ``:group:``.
It's ready to me merged

mthuurne:
Regarding `except`, maybe we can add support for the synonym instead of declaring the feature unsupported.

For example the `returns` synonym for `return` is implemented like this in `pydoctor/epydoc2stan.py`:
```py
class FieldHandler:
    ...
    def handle_return(self, field: Field) -> None:
        ...
    handle_returns = handle_return
```
If a single line code change is sufficient, I'd prefer that over removing the documentation.

tristanlatr:
All done

mthuurne:
Does this fix the consolidated field support?
If so, there is no need to remove `:Exceptions:` from the consolidated fields table.
If not, then why not? ;)

In `restructuredtext.doctest` there is this test:
```
>>> parse_and_print(
... """A test function
...
... :exceptions: - `KeyError`: if the key is not found
...              - `ValueError`: if the value is bad
... """)
A test function
except "KeyError": if the key is not found
except "ValueError": if the value is bad
```
However, this only tests that epydoc parses the fields correctly, it doesn't run the data through `epydoc2stan.FieldHandler`.

tristanlatr:
> Does this fix the consolidated field support?

No

> If not, then why not? ;)

Cause I don't know where to look for this code and the PR was only about documentation at the beginning.

mthuurne:
> > Does this fix the consolidated field support?
> 
> No
> 
> > If not, then why not? ;)
> 
> Cause the PR was only about documentation at the beginning.

The mechanism works like this: epydoc parses an `:Exceptions:` block and emits epydoc `except` fields, then `FieldHandler` converts those epydoc `except` fields to pydoctor `raises` fields.

So in theory, adding the `except` synonym in `FieldHandler` will already have fixed support for the `:Exceptions:` consolidated field. However, in practice it's safer to not assume it works until it has been tested.

mthuurne:
> The mechanism works like this: epydoc parses an `:Exceptions:` block and emits epydoc `except` fields, then `FieldHandler` converts those epydoc `except` fields to pydoctor `raises` fields.
>
> So in theory, adding the `except` synonym in `FieldHandler` will already have fixed support for the `:Exceptions:` consolidated field. However, in practice it's safer to not assume it works until it has been tested.

I tested manually and indeed the consolidated `:Exceptions:` field is working now. So I put it back into the supported fields table.

PR #309, opened 2020-12-03
Replace --add-package and --add-module by positional arguments

### LINKED ISSUE #255: --add-package should fail when directory does not exists.
I have this tox env

```
[testenv:apidocs]
description = Build only the API documentation

commands =
    pydoctor \
    --add-package=bla_pydoctor \
    --project-name=pydoctor \
    --project-url=https://github.com/twisted/pydoctor/ \
    --html-viewsource-base=https://github.com/twisted/pydoctor/tree/master \
    --html-output=build/apidocs \
    --project-base-dir="{toxinidir}" \
    --docformat=epytext \
    --intersphinx=https://docs.python.org/3/objects.inv \
    --make-html
```

And it reporting that all is ok **adding module /home/adi/chevah/pydoctor/bla_pydoctor**

```
$ tox -e apidocs

adding module /home/adi/chevah/pydoctor/bla_pydoctor
writing html to build/apidocs using pydoctor.templatewriter.writer.TemplateWriter
starting ModuleIndexPage ...took 0.000845s
starting ClassIndexPage ...took 0.000718s
starting IndexPage ...took 0.001189s
starting NameIndexPage ...took 0.000789s
starting UndocumentedSummaryPage ...took 0.000718s
Generating objects inventory at build/apidocs/objects.inv
___________________________________ summary ____________________________________
  apidocs: commands succeeded
  congratulations :)
```

mthuurne:
The two command line options were already just added to the positional arguments, but the naming and documentation suggested otherwise.

The options remain available, but print a warning when used.

We now also reject non-existing paths passed using either option or as a positional argument.

Fixes #255.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,7 @@ Simple Usage
 
 You can run pydoctor on your project like this::
 
-    $ pydoctor --make-html --html-output=docs/api --add-package=src/mylib
+    $ pydoctor --make-html --html-output=docs/api src/mylib
 
 For more info, `Read The Docs <https://pydoctor.readthedocs.io/>`_.
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -82,7 +82,6 @@ if os.environ.get('READTHEDOCS', '') == 'True':
 
 _pydoctor_root = pathlib.Path(__file__).parent.parent.parent
 pydoctor_args = [
-    f'--add-package={_pydoctor_root}/pydoctor',
     '--html-output={outdir}/api',
     f'--project-base-dir={_pydoctor_root}',
     f'--html-viewsource-base=https://github.com/twisted/pydoctor/tree/{_git_reference}',
@@ -95,4 +94,5 @@ pydoctor_args = [
     '--intersphinx=https://twistedmatrix.com/documents/current/api/objects.inv',
     '--intersphinx=https://urllib3.readthedocs.io/en/latest/objects.inv',
     '--intersphinx=https://requests.readthedocs.io/en/latest/objects.inv',
+    f'{_pydoctor_root}/pydoctor',
     ]
--- a/pydoctor/driver.py
+++ b/pydoctor/driver.py
@@ -335,12 +335,10 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
                     system.msg('addModuleFromPath', 'adding module ' + path)
                     system.addModuleFromPath(prependedpackage, path)
                 system.packages.append(path)
+        else:
+            error("No source paths given.")
 
         # step 3: move the system to the desired state
-
-        if not system.packages:
-            error("The system does not contain any code, did you "
-                  "forget an --add-package?")
 
         if system.options.projectname is None:
             name = '/'.join(ro.name for ro in system.rootobjects)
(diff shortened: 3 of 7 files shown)


adiroiban (approved):
Great work. Thanks!

Is ok to just add sanity checking for the positional args... as the others will be removed.

Do we have a deprecation policy ?  Purge every 2 releases?

You have 1 release to run and observer the deprecations warnings and be the next release the code is removed.

With that policy, we might want to add standard DeprecationWarning to make it easy to detect the deprecations.

mthuurne:
> Is ok to just add sanity checking for the positional args... as the others will be removed.

The `main()` merges the three lists before the sanity checks:
```py
args = list(args) + options.modules + options.packages
```
So this means that we do perform sanity checks on all three lists, but we are sloppy in the sense that you can pass a module via `--add-package` or a package via `--add-module`.

> Do we have a deprecation policy ? Purge every 2 releases?
>
> You have 1 release to run and observer the deprecations warnings and be the next release the code is removed.

No policy yet.

The earliest possible moment for removal would be the next major release (never on bug fix releases).

I think we can just leave the options there until they become a maintenance burden. So in this case, if we ever replace `optparse` by a different parser, that would be a good time to drop the options.

> With that policy, we might want to add standard DeprecationWarning to make it easy to detect the deprecations.

Easier for us to detect or for the end user? Because a `DeprecationWarning` won't been seen by end users, as printing of warnings is disabled by default in Python. And unlike libraries, we don't have a test harness like `pytest` enabling warnings when a user runs pydoctor.

adiroiban:
Thanks for the info.

Deprecation removal only on minor or major releases.

Yes. We can keep the current command line options forever... or when we found the first bug for them :)

My expectation is that pydoctor is a dev tool, and end users are running python for tests + dev tools with warnings enabled.

mthuurne:
> Deprecation removal only on minor or major releases.

With our versioning scheme being `YYYY.MM.micro`, the difference between major and minor releases is arbitrary. I think of 2020.12.01 as a major release, even though it happens in the same year as 2020.07.01.

> Yes. We can keep the current command line options forever... or when we found the first bug for them :)

Forever is long time, but rushing to remove the options can inconvenience users while it doesn't gain us much.

> My expectation is that pydoctor is a dev tool, and end users are running python for tests + dev tools with warnings enabled.

I think a lot of developers aren't even aware that warnings are disabled by default. I had been doing Python development for years before I found out and in an IRC discussion I wasn't the only one with that experience.

adiroiban:
Yes. I think that the current print to stderr is the best thing :) and there is no point in removing  it now.

----

Not sure how to call the "normal"  non-bugfix release.

So yes. Never remove deprecation on bugfix releases...but this is common sense :+1: 
Bugfix release don't add or remove any features... they just fix a defect that is urgent enough to trigger a bugfix release :)

-----

I think that we just need a little bit more education for the deprecation process.

I think that in the past, many open source project were lacking the resources (time) to care about deprecation ... so the focus was only on making the next release better without looking in the past.

At the same time, if a library is just a dev tools and not some code running in production, I have a more released (agile) approach... and I don't dedicate much time with deprecation warnings and backward compatibility.

But I think that adding deprecation is a very good practice and signal that the library is mature/professional  and can be used for production stuff and not just as a hobby:) 

So when you want to upgrade, you can install the next version, give it a run and see if you have deprecation... before pushing in production.

The calendar versioning scheme works very nice with the deprecation process.

We can keep the deprecation for 1 year for example.... as this will make sure that if your current version don't raise any version you can upgrade directly to the +1 year version and check for deprecation...and you can safely skip all the intermediary versions.

Issue #285, opened 2020-11-24
Output samples


mthuurne:
@adiroiban [wrote](https://github.com/twisted/pydoctor/pull/282#pullrequestreview-537405656):
> My suggestion is to create a pydoctor.test.sample.epydoc or pydoctor.samples.epydoc package in which we include all the possible usage of pydoctor with epydoc
>
> In this way, the result of this PR could be observed via
> https://pydoctor--282.org.readthedocs.build/en/282/api/pydoctor.test.html

@mthuurne [wrote](https://github.com/twisted/pydoctor/pull/282#issuecomment-732943736):
> I think it would be good to have a package that displays all the supported use cases, but I think it should be part of the documentation rather than part of pydoctor's code.
>
> One reason is that we'd want to showcase syntax that is not available in all Python versions, such as position-only parameters. Having that syntax in the main source dir could cause problems with distros that pre-optimize Python modules on install.
>
> Another reason is that I wouldn't want samples to be mixed into the indexes of the actual API docs. In the package index it would be just one line, but in the class hierarchy and name index it would produce more noise.

adiroiban:
True. It should not be part of the pydoctor code.

We can build the sample as part if /api-demo/ or something like this... to not mix it with the actual pydoctor API.

This might need extra hacks to build it on read the docs.

------------

We don't need to demonstrate all python version. We can start with just latest python (3.9)

mthuurne:
> This might need extra hacks to build it on read the docs.

We already have the actual API docs on readthedocs using our Sphinx extension. Perhaps we could just invoke that extension at two points in the documentation, each with a different source dir?

> We don't need to demonstrate all python version. We can start with just latest python (3.9)

There are no features in older Python versions that aren't in 3.9, nor do I expect there to be during the 3.x series. Backwards compatibility was only a concern when we supported 2.7 as well, but that is no longer the case.

The problem I was concerned about is forward compatibility: older Python versions may not be able to parse the latest language features. If the samples are in a separate source dir, this is less of a problem because only the doc builds will require a recent Python version then.

adiroiban:
>  Perhaps we could just invoke that extension at two points in the documentation, each with a different source dir?

the current extension is not designed to allow calling pydoctor multiple times :) ... but I can see if I can hack something

tristanlatr:
@adiroiban I did this here but I revert it cause it wasn't useful for my PR.
[...]

PR #312, opened 2020-12-04
Support `@property` decorators

### LINKED ISSUE #233: Properties are considered as methods
Currently `@property` decorator is not understood correctly.
It shows as a method, but is should show as a instance variable.

The docstring should be extracted from the getter (not sure if it's already the case)

This is a bug report.

The workaround is to create the property with the `property(fset, fget)`, then it's understood.

mthuurne:
There are a lot of different ways to implement and document properties. I tried to support the most common ones.

My personal favorite syntax:
```py
    class C:
        @property
        def prop(self) -> str:
            """For sale."""
            return 'seaside'
```

And a more traditional syntax, used in Twisted:
```py
class C:
        @property
        def oldschool(self):
            """
            @return: For rent.
            @rtype: L{str}
            """
            return 'downtown'
```

There is also support for `setter` and `deleter` methods, although that is very rudimentary: these methods are renamed to `<propname>.setter` and `<propname>.deleter` respectively and listed in the documentation as separate methods. Still, that is better than how they're handled today, which is overwriting the getter (since all accessor methods have the same name).

Please run this on your projects and check whether the output looks reasonable.

Fixes #233.

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -450,35 +450,36 @@ class ModuleVistor(ast.NodeVisitor):
             is_async: bool
             ) -> None:
         # Ignore inner functions.
-        if isinstance(self.builder.current, model.Function):
+        parent = self.builder.current
+        if isinstance(parent, model.Function):
             return
 
         lineno = node.lineno
         if node.decorator_list:
             lineno = node.decorator_list[0].lineno
 
+        is_classmethod = False
+        is_staticmethod = False
+        if isinstance(parent, model.Class) and node.decorator_list:
+            for d in node.decorator_list:
+                if isinstance(d, ast.Name):
+                    if d.id == 'classmethod':
+                        is_classmethod = True
+                    elif d.id == 'staticmethod':
+                        is_staticmethod = True
+
         func = self.builder.pushFunction(node.name, lineno)
         func.is_async = is_async
         if len(node.body) > 0 and isinstance(node.body[0], ast.Expr) and isinstance(node.body[0].value, ast.Str):
             func.setDocstring(node.body[0].value)
         func.decorators = node.decorator_list
-        if isinstance(func.parent, model.Class) and node.decorator_list:
-            isclassmethod = False
-            isstaticmethod = False
-            for d in node.decorator_list:
-                if isinstance(d, ast.Name):
-                    if d.id == 'classmethod':
-                        isclassmethod = True
-                    elif d.id == 'staticmethod':
-                        isstaticmethod = True
-            if isstaticmethod:
-                if isclassmethod:
-                    func.report(f'{func.fullName()} is both classmethod '
-                                f'and staticmethod')
-                else:
-                    func.kind = 'Static Method'
-            elif isclassmethod:
-                func.kind = 'Class Method'
+        if is_staticmethod:
+            if is_classmethod:
+                func.report(f'{func.fullName()} is both classmethod and staticmethod')
+            else:
+                func.kind = 'Static Method'
+        elif is_classmethod:
+            func.kind = 'Class Method'
 
         # Position-only arguments were introduced in Python 3.8.
         posonlyargs: Sequence[ast.arg] = getattr(node.args, 'posonlyargs', ())
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -460,20 +460,42 @@ class ModuleVistor(ast.NodeVisitor):
         if node.decorator_list:
             lineno = node.decorator_list[0].lineno
 
+        docstring: Optional[ast.Str] = None
+        if len(node.body) > 0 and isinstance(node.body[0], ast.Expr) \
+                              and isinstance(node.body[0].value, ast.Str):
+            docstring = node.body[0].value
+
+        is_property = False
         is_classmethod = False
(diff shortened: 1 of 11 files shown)


tristanlatr:
Does it understand the property(fget, fset) function ?

mthuurne:
> Does it understand the property(fget, fset) function ?

It doesn't, currently. Is that syntax still being used? I assumed you only tried it because `@property` wasn't working.

adiroiban:
Example of read only - https://pydoctor--312.org.readthedocs.build/en/312/api/pydoctor.epydoc.markup.ParsedDocstring.html#has_body

pydoctor/templatewriter/pages/functionchild.py
```diff
@@ -57,7 +57,10 @@ def decorator(self, request, tag):
     @renderer
     def functionDef(self, request, tag):
         def_stmt = 'async def' if self.ob.is_async else 'def'
-        return [def_stmt, ' ', self.ob.name, signature(self.ob), ':']
+        name = self.ob.name
+        if name.endswith('.setter') or name.endswith('.deleter'):
+            name = name[:name.rindex('.')]
```
adiroiban, line 62: Minor comment:
At least for the setter, if you see that a property has no setter you can think that it's read-only.

But maybe, if a deleter is defined and has no docstring, it can't be skipped from the docs?

adiroiban (approved):
Looks good. Thanks.

I think that this is a big improvement.

I will try to work at creating the sample API building so that we can have examples as part of the docs....but this can be done in a separate PR.
It's just a note that https://github.com/twisted/pydoctor/issues/285 is important :)

mthuurne:
> Example of read only - https://pydoctor--312.org.readthedocs.build/en/312/api/pydoctor.epydoc.markup.ParsedDocstring.html#has_body

I'll fix the overly long summary by putting an empty line between the two sentences.

pydoctor/astbuilder.py
```diff
@@ -450,35 +452,81 @@ def _handleFunctionDef(self,
...
-            elif isclassmethod:
-                func.kind = 'Class Method'
+                        is_staticmethod = True
+                elif isinstance(d, ast.Attribute):
+                    if d.attr in ('setter', 'deleter'):
+                        # Rename the setter/deleter, so it doesn't replace
+                        # the property object.
+                        func_name = f'{func_name}.{d.attr}'
+
+        if is_property:
```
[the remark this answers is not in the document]
mthuurne: I thought that might require passing too many values, but it's actually not so bad if I exclude `is_classmethod` and `is_staticmethod` by applying the warnings on a returned object.

pydoctor/templatewriter/pages/functionchild.py
mthuurne: It might still be useful information to know that the property can be deleted. Also the source link for the deleter would be missing then, although in practice getter/setter/deleter often get defined close to each other, so that's not a big problem.

I don't think deleters are common: there are zero in Twisted and I can't remember ever having used one myself. I had even forgotten they existed until I re-read the language documentation when working on this PR.

I think that in the future we probably want to revisit how properties are presented and create some kind of unified display that combines information from the getter, setter and deleter in a single output box. But that's too ambitious to do right now.

mthuurne:
> I think that this is a big improvement.

Thanks! I wasn't sure this was something that I could implement in a short time, but I'm happy with how it turned out.

> I will try to work at creating the sample API building so that we can have examples as part of the docs....but this can be done in a separate PR.
> It's just a note that #285 is important :)

Agreed.

I'm hoping though that the most common ways in which people document properties are now covered, such that the output will look OK even if they haven't read the manual.

tristanlatr:
> It doesn't, currently. Is that syntax still being used? I assumed you only tried it because @property wasn't working.

I did used that syntax in the past. There are times where fget and fset avoid the usage of a tweak to call the abstract inherited setter for a propertie.
```
super(self.__class__, self.__class__).x.__set__(self, value)
```
https://bugs.python.org/issue14965

I don't understand why properties are marked are "Properties" but not as "instance variable", it's not the same thing ?

mthuurne:
> > It doesn't, currently. Is that syntax still being used? I assumed you only tried it because `@property` wasn't working.
> 
> I did used that syntax in the past. There are times where fget and fset avoid the usage of a tweak to call the abstract inherited setter for a propertie.
> 
> ```
> super(self.__class__, self.__class__).x.__set__(self, value)
> ```
> 
> https://bugs.python.org/issue14965

I'm not against supporting the fget/fset syntax, but for me it has a low priority.

> I don't understand why properties are marked are "Properties" but not as "instance variable", it's not the same thing ?

That's a good question: is a property a different interface or is it an implementation detail?

There are aspects of properties that don't apply to plain variables:
- properties can be read-only
- if you set and then get the value, the result might not be the exact same value
- the setter can raise an exception if it rejects the value

In particular the raising of exceptions is relevant for documentation, since the user might want to document which exception types are raised in which situations.

Therefore I think that having properties as a separate category is worthwhile.

Note that properties do use the same color coding as instance variables, so they are visually close to them.

adiroiban:
I like that properties are not documented as "instance variables".

The property decorator is just syntactic sugaring for calling instance methods.

PR #316, opened 2020-12-04
Support custom property decorators

### LINKED ISSUE #238: Inconsistent output of function decorators
This is a simplified version of the test case of #234:
```py
from twisted.python.deprecate import deprecated
from incremental import Version
from twisted.python import deprecate
import incremental

@deprecated(Version('Twisted', 15, 0, 0), 'Baz')
def foo():
    'text in foo docstring'

@deprecate.deprecated(incremental.Version('Twisted', 16, 0, 0))
def bar():
    'text in bar docstring'
```
In the HTML output, the `@deprecate.deprecated` decorator is shown, while the `@deprecated` decorator is not. The factor that makes the difference is the dotted name.

While I didn't test class decorators in the same way yet, it's possible the issue applies there as well.

Do we want to show decorators always, sometimes or never? I guess it depends on whether a particular decoration is considered part of the API or an implementation detail. Is there a mostly reliable way to guess that? And possibly a way for the user to override if we mis-guess.

mthuurne:
Twisted has `@deprecatedProperty`, Django has `@cached_property` and I also found [`@async_property`](https://pypi.org/project/async-property/) when wondering whether `async def` functions could be properties.

So there are many decorators that act similar to properties. I figured it would be useful to support them all, simply by treating all decorations with names ending in "property" as property definitions.

Besides that, there is a commit to make sure that if `@deprecatedProperty` sets deprecation info on a property (this requires a change in Twisted's customizations), the deprecation notice will actually end  up in the output.

Fixes #238.

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -6,7 +6,9 @@ from functools import partial
 from inspect import Parameter, Signature
 from itertools import chain
 from pathlib import Path
-from typing import Any, Dict, Iterable, Iterator, Mapping, Optional, Sequence, Tuple, Union
+from typing import (
+    Any, Dict, Iterable, Iterator, List, Mapping, Optional, Sequence, Tuple, Union
+)
 
 import astor
 from pydoctor import epydoc2stan, model
@@ -24,7 +26,7 @@ else:
     _parse = ast.parse
 
 
-def node2dottedname(node):
+def node2dottedname(node: Optional[ast.expr]) -> Optional[List[str]]:
     parts = []
     while isinstance(node, ast.Attribute):
         parts.append(node.attr)
@@ -41,7 +43,7 @@ def node2dottedname(node):
     return parts
 
 
-def node2fullname(expr, ctx):
+def node2fullname(expr: Optional[ast.expr], ctx: model.Documentable) -> Optional[str]:
     dottedname = node2dottedname(expr)
     if dottedname is None:
         return None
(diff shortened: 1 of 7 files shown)


adiroiban (approved):
Looks good. Thanks.

I hope we can consolidate deprecation code into `regret` and have a single deprecation helper library.

Below is a property constructor that I am using in my project.

```python
def WritableString(name, default=None):
    """
    String configuration which can be written.

    Uses a helper to let pydoctor know that then used this will look like a property.

    @property
    """
    def getter(self):
        result = self._proxy.getString(self._section_name, name)
        if not result and default is not None:
            return default
        return result

    def setter(self, value):
        return self._proxy.setString(self._section_name, name, value)

    return property(getter, setter)


class SomeConfigurateOptions:
   """
   This is the view of a .INI conf file that has `test_config = something`
   """

   test_config = WritableString('test_config')

   def __init__(self, path):
       self._proxy = IniFile(path)
```

mthuurne:
> I hope we can consolidate deprecation code into `regret` and have a single deprecation helper library.

I agree, but `regret` is a long-term plan, while supporting `@deprecatedProperty` is a short-term solution.

> Below is a property constructor that I am using in my project.

Is this a feature request? Because this will not be recognized as a property by the current code.

adiroiban:
Thanks for this. Happy to see the custom property as it is now... as it is a big improvement.

> Is this a feature request? Because this will not be recognized as a property by the current code.

Nope... it was just an idea of how we can tag something as a property in a generic way.

For my projects, I use free text markup and only read the source code :+1: 

Even with Twisted, most of the time, I read the docstring and code... note sure why. Maybe the documentation is not good enough.

For for python stdlib / pyopenssl or pycryptography I do read the API docs.

PR #320, opened 2020-12-05
Various small improvements before the release

mthuurne:
There is no overarching theme here; just check the individual commits for details.

See [the dot color change in action.](https://pydoctor--320.org.readthedocs.build/en/320/api/pydoctor.epydoc.markup.epytext.Element.html)

Issue #86, opened 2015-05-13
Process the docstrings summary from docutils nodesm not source


glyph:
compare https://twisted.github.io/tubes/docs/tubes.fan.html and https://twisted.github.io/tubes/docs/tubes.fan.In.html

mwhudson:
Oh yeah, I bet the summary string is seen as just "A fan.In presents a single fount that delivers the inputs from multiple drains::"

I guess just s/::$/:/ on the summary would fix this, seems a touch gross though.

glyph:
It sounds like the "right" answer would be to extract the summary post-parsing – you're making it sound like it's extracted as source text?

mwhudson:
Ah yes, that would definitely make more sense (currently it is indeed done on the source).

tristanlatr:
When PR #386 is merged, we can start working on generating the summary HTML from the `docutils` document and fix this this issue.

tristanlatr:
So ``pydoctor``, two years ago, was processing the summary from the docutils node, at least in the case of the reStructuredText: https://github.com/twisted/pydoctor/commit/253f9a90b8dd9ad629e60d0f7ca88e043a1aff75#diff-9b0b2f7d4a8b784bf04769b8343cf25857cdead8111fda83f61fea1f121a6b90L258 . 

It would not be too hard to bring back this to life, and add tests to it.

PR #322, opened 2020-12-06
Support 'type' field in inline docstrings

mthuurne:
I noticed something odd on the [documentation page I linked in #320](https://pydoctor--320.org.readthedocs.build/en/320/api/pydoctor.epydoc.markup.epytext.Element.html): the `@type` tags in the summary. These are the result of a known issue though (#86). However, the type documentation was also missing from from the blocks in which `tag`, `children` and `attribs` are documented in detail and that could not be explained by the way the summary is parsed, since those blocks are parsed from the full docstring.

It turns out that the handling of the `type` field ignored any entry without a name. This makes sense for functions, where we cannot use the type if we cannot match it to a parameter, so there we report a warning and then discard the type.

But when an attribute is documented by an inline docstring, no name is required, since an inline docstring by definition only applies to a single documentable. So discarding the type was incorrect there.

I don't know whether this is a regression as a result of the parameter annotation work or whether it never worked, but I don't feel like doing code archeology right now. I think it would be good to fix it before release in either case.

[Output after the fix.](https://pydoctor--322.org.readthedocs.build/en/322/api/pydoctor.epydoc.markup.epytext.Element.html)

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -199,7 +199,7 @@ class Field:
     arg: Optional[str]
     source: model.Documentable
     lineno: int
-    body: Tag
+    body: ParsedDocstring
 
     @classmethod
     def from_epydoc(cls, field: EpydocField, source: model.Documentable) -> 'Field':
@@ -208,8 +208,11 @@ class Field:
             arg=field.arg(),
             source=source,
             lineno=field.lineno,
-            body=field.body().to_stan(_EpydocLinker(source))
+            body=field.body()
             )
+
+    def format(self) -> Tag:
+        return self.body.to_stan(_EpydocLinker(self.source))
 
     def report(self, message: str) -> None:
         self.source.report(message, lineno_offset=self.lineno, section='docstring')
@@ -226,7 +229,7 @@ def format_field_list(singular: str, plural: str, fields: Sequence[Field]) -> It
         else:
             row = tags.tr()
             row(tags.td())
-        row(tags.td(colspan="2")(field.body))
+        row(tags.td(colspan="2")(field.format()))
         yield row
 
 
@@ -270,7 +273,7 @@ class FieldHandler:
             field.report('Unexpected argument in %s field' % (field.tag,))
         if not self.return_desc:
             self.return_desc = FieldDesc(kind='return')
-        self.return_desc.body = field.body
+        self.return_desc.body = field.format()
     handle_returns = handle_return
 
     def handle_returntype(self, field: Field) -> None:
@@ -278,7 +281,7 @@ class FieldHandler:
             field.report('Unexpected argument in %s field' % (field.tag,))
         if not self.return_desc:
             self.return_desc = FieldDesc(kind='return')
-        self.return_desc.type = field.body
+        self.return_desc.type = field.format()
     handle_rtype = handle_returntype
 
     def _handle_param_name(self, field: Field) -> Optional[str]:
@@ -313,7 +316,7 @@ class FieldHandler:
         field.report('Documented parameter "%s" does not exist' % (name,))
 
     def add_info(self, desc_list: List[FieldDesc], name: Optional[str], field: Field) -> None:
-        desc_list.append(FieldDesc(kind=field.tag, name=name, body=field.body))
+        desc_list.append(FieldDesc(kind=field.tag, name=name, body=field.format()))
 
     def handle_type(self, field: Field) -> None:
         if isinstance(self.obj, model.Function):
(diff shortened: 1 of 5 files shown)


mthuurne:
Something weird happens when generating Twisted's docs:
```
src/twisted/conch/ssh/channel.py:170: Field in inline docstring should not include a name
src/twisted/internet/protocol.py:577: Field in inline docstring should not include a name
src/twisted/web/http.py:1365: Field in inline docstring should not include a name
src/twisted/web/http.py:1368: Field in inline docstring should not include a name
src/twisted/web/http.py:1255: Field in inline docstring should not include a name
src/twisted/web/http.py:1258: Field in inline docstring should not include a name
src/twisted/web/http.py:1262: Field in inline docstring should not include a name
src/twisted/web/http.py:1265: Field in inline docstring should not include a name
src/twisted/web/http.py:1268: Field in inline docstring should not include a name
src/twisted/web/http.py:1271: Field in inline docstring should not include a name
src/twisted/web/http.py:1274: Field in inline docstring should not include a name
src/twisted/web/http.py:1278: Field in inline docstring should not include a name
src/twisted/web/http.py:1282: Field in inline docstring should not include a name
src/twisted/web/http.py:1288: Field in inline docstring should not include a name
src/twisted/web/http.py:1440: Field in inline docstring should not include a name
src/twisted/web/http.py:1347: Field in inline docstring should not include a name
src/twisted/web/http.py:1348: Field in inline docstring should not include a name
src/twisted/web/http.py:1403: Field in inline docstring should not include a name
```
These lines do not contain inline docstrings though: they contain method docstrings. Which means that pydoctor is applying method docstrings to documentables of type `Attribute` for some reason (type `Function` would be expected).

My guess is that this is not a regression from this PR, but instead a problem in the processing of inline docstrings that has gone undetected until now, because it didn't lead to warnings or incorrect output.

Please do review the code changes, but we shouldn't merge this until we know exactly what is going on.

mthuurne:
The problem in `conch/ssh/channel.py` seems to be caused by pydoctor misunderstanding this code in `conch/scripts/cftp.py`:
```py
class SSHSession(channel.SSHChannel):
    ...
    def _cbSubsystem(self, result):
        self.client = filetransfer.FileTransferClient()
        self.client.makeConnection(self)
        self.dataReceived = self.client.dataReceived
        ...
```
`dataReceived` is a method in `SSHChannel`, but `SSHSession` overrides it by assigning a bound method, to delegate calls to a `FileTransferClient` instance.

mthuurne:
The problem in `internet/protocol.py` seems to be caused by pydoctor misunderstanding this code in `web/_newclient.py`:
```py
class HTTP11ClientProtocol(Protocol):
    ...
    def connectionLost(self, reason):
        """
        The underlying transport went away.  If appropriate, notify the parser
        object.
        """

    connectionLost = makeStatefulDispatcher("connectionLost", connectionLost)
```

So the statement that wraps the `connectionLost()` method makes pydoctor think that it's dealing with a class variable rather than a method from there on. This should be pretty easy to avoid.

Like the first reported problem, `HTTP11ClientProtocol.connectionLost()` was already mis-classified as an instance variable [in Twisted 20.3.0's documentation](https://twistedmatrix.com/documents/20.3.0/api/twisted.web._newclient.HTTP11ClientProtocol.html).

mthuurne:
I'm convinced now that the warnings logged when processing Twisted are caused by pre-existing problems where an assignment that shadows or replaces a method leads to the documentable type changing from `Function` to `Attribute`. That problem will be fixed in #323.

So when reviewing this PR, it's fine to ignore those warnings.

pydoctor/epydoc2stan.py
```diff
@@ -208,9 +208,12 @@ def from_epydoc(cls, field: EpydocField, source: model.Documentable) -> 'Field':
             arg=field.arg(),
             source=source,
             lineno=field.lineno,
-            body=field.body().to_stan(_EpydocLinker(source))
+            body=field.body()
             )
 
+    def format(self) -> Tag:
```
adiroiban, line 214: I think that a docstring explaining the purpose of `format` would be nice :)

adiroiban (approved):
Great work. Thanks!

Here this feature is referred as "inline docstring" ,but in the epytext docs I see this referred as " variable docstrings" - http://epydoc.sourceforge.net/manual-docstring.html

mthuurne:
> Here this feature is referred as "inline docstring" ,but in the epytext docs I see this referred as " variable docstrings"

It's an inline docstring that is used to document a variable, so I think both are correct. But "variable docstring" is more specific, so I'll change it.

pydoctor/epydoc2stan.py
mthuurne: I'll add one.

PR #323, opened 2020-12-06
Avoid methods being classified as variables

mthuurne:
Fixes for the incorrect classifications that cause warnings in Twisted in PR #322. See that PR for code examples that pydoctor handled incorrectly.

When an assignment wraps a method, that assignment would create an `Attribute` object in the code model. Then when documenting that object, the docstring was inherited from a method and pydoctor would warn that fields like `@type name: str` don't need a parameter name when attached to a variable (`Attribute`).

The solution I picked is that before creating an `Attribute`, pydoctor will check whether the name already exists in the class itself or through inheritance. If the name exists and it does not belong to an `Attribute` (because it's a `Function` instead, for example), the assignment will be ignored. I think this solution is good enough for now and it solves all the false positive warnings in Twisted.

One thing to revisit later is whether `Class.allbases()` uses the correct method resolution order. But the current implementation seems to be good enough for Twisted at least.

Another thing we could consider is whether it is useful to be able to place a docstring after a wrapped method. I don't see an immediate use for it, but there might be. For now, I picked the safer option of ignoring the assignment completely, so if a docstring follows it, that will be ignored as well.

### DIFF
--- a/pydoctor/test/test_astbuilder.py
+++ b/pydoctor/test/test_astbuilder.py
@@ -552,41 +552,61 @@ def test_methoddecorator(systemcls: Type[model.System], capsys: CapSys) -> None:
 def test_assignment_to_method_in_class(systemcls: Type[model.System]) -> None:
     """An assignment to a method in a class body does not change the type
     of the documentable.
+
+    If the name we assign to exists and it does not belong to an Attribute
+    (it's a Function instead, in this test case), the assignment will be
+    ignored.
     """
     mod = fromText('''
     class Base:
-        def base_method(): ...
+        def base_method():
+            """Base method docstring."""
 
     class Sub(Base):
         base_method = wrap_method(base_method)
+        """Overriding the docstring is not supported."""
 
-        def sub_method(): ...
+        def sub_method():
+            """Sub method docstring."""
         sub_method = wrap_method(sub_method)
+        """Overriding the docstring is not supported."""
     ''', systemcls=systemcls)
     assert isinstance(mod.contents['Base'].contents['base_method'], model.Function)
     assert mod.contents['Sub'].contents.get('base_method') is None
-    assert isinstance(mod.contents['Sub'].contents['sub_method'], model.Function)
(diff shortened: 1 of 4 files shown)


pydoctor/test/test_astbuilder.py
```diff
@@ -548,6 +548,47 @@ def method_both():
...
+        def base_method(): ...
+
+    class Sub(Base):
+        base_method = wrap_method(base_method)
+
+        def sub_method(): ...
+        sub_method = wrap_method(sub_method)
+    ''', systemcls=systemcls)
+    assert isinstance(mod.contents['Base'].contents['base_method'], model.Function)
+    assert mod.contents['Sub'].contents.get('base_method') is None
```
adiroiban, line 567: I don't understand this assertion.... I was expecting that since base_method is redefined in `Sub` it will be part of `Sub`s content.

Also, I was expecting to see an assertion on the docstring content... to check which docstring is used.

adiroiban (approved):
Looks good. Thanks! Only a minor comment.

pydoctor/test/test_astbuilder.py
mthuurne: We ignore the assignment that does the redefinition because it would change the type of `base_method` from `Function` to `Attribute`. That's why `base_method` doesn't exist in the `Sub` class itself.

If there were a `Function` object on the subclass, checking the docstring would be useful. But currently, the only thing we could check is the docstring on the base class, which is not very useful.

adiroiban: Maybe also update the docstring with the info that an assigment will just ignore the overwritten method.
I got this. And I think that for now is ok... and later if someone complains we have this documented.

In term of `sub_method` ... I guess that it will have the original docstring and not the docstring from wrap_method...but I think that it is still worth checking it.

Thanks!

adiroiban (approved):
Thanks. Great docs :) This can be merged

PR #325, opened 2020-12-09
Avoid crashes where relative_to() raises ValueError

mthuurne:
Twisted's pydoctor tests failed on macOS when computing a relative path. The problem seems to be caused by symbolic links in the source path leading to a mismatch between the absolute source path and the absolute project base directory, since both are converted to absolute paths in different ways. So while macOS had the right circumstances to trigger this bug in the default test setup, the bug itself can occur on all platforms.

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -17,8 +17,10 @@ jobs:
         python-version: [3.6, 3.7, 3.8, 3.9, pypy3]
         os: [ubuntu-20.04]
         include:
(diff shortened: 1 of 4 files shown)


adiroiban (approved):
Great work. Thanks!

Maybe update the version in the same PR and after merge in master, tag the release?

mthuurne:
I would like to also like to add the missing package docstring source link (module docstrings are linked, package docstrings are not) and then do a bugfix release.

PR #326, opened 2020-12-09
Bring back package source link

mthuurne:
Modules had a source link, but it was missing on packages, since I removed the call that stored the URL when cleaning up the code a few weeks ago.

Instead of storing the URL, we now fetch it from the `__init__` module; the package and the module should really be the same thing, but they're currently implemented as separate Documentables.

I also added a unit test.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -65,6 +65,7 @@ pydoctor 20.12.1 (unreleased)
 
 * Reject source directories outside the project base directory (if given), instead of crashing
 * Fixed bug where source directories containing symbolic links could appear to be outside of the project base directory, leading to a crash
+* Bring back source link on package pages
 
 pydoctor 20.12.0
 ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -65,6 +65,7 @@ pydoctor 20.12.1 (unreleased)
 
 * Reject source directories outside the project base directory (if given), instead of crashing
 * Fixed bug where source directories containing symbolic links could appear to be outside of the project base directory, leading to a crash
+* Bring back source link on package pages
```
adiroiban, line 68: Very minor comment.

When we have full sentences, I prefer to have a full stop at the end. It's also a hint that the sentence was not accidentally left unfinished.

Maybe move this into a separate NEWS.rst /  Changelog.rst / ReleaseNotes.rst file

adiroiban (approved):
All good. Thanks!

README.rst
mthuurne: I thought it was useful to have the release notes immediately visible on the GitHub page. But I underestimated how many changes we'd be making and the release notes became quite bulky, so I guess moving it to a separate document would make sense. We could do both full stops and separate file for the next major release.

PR #328, opened 2020-12-11
Add helper to insert the git source reference.

adiroiban:
This moves some code outside of conf.py as I saw the need to copy/paste it when adding pydoctor extension to twisted/twisted 

Let me know  what do you think?

Thanks!

### DIFF
--- a/pydoctor/sphinx_ext/__init__.py
+++ b/pydoctor/sphinx_ext/__init__.py
@@ -1,3 +1,58 @@
 """
 Public and private extensions for Sphinx.
 """
+import os
+import subprocess
+from pprint import pprint
+
+
+def get_git_reference(main_branch: str = 'master', debug: bool = False, ) -> str:
+    """
+    Return the reference for current git checkout to be used when linking
+    the source files.
+
+    @param main_branch: Name of the main branch to be used to link the latest
+        build on Read The Docs.
+
+    @param debug: Pass C{True} if you want to print environment variables
+        and other useful info.
+
+    @return: Tag name, default branch name, or fallback to SHA.
+    """
+    reference_name = subprocess.run(
+        ["git", "rev-parse", "--abbrev-ref", "HEAD"],
+        text=True,
+        encoding="utf8",
+        capture_output=True,
+        check=True,
+    ).stdout.strip()
+
+    result = reference_name
+
+    if result == "HEAD":
+        # It looks like the branch has no name.
+        # Fallback to commit ID.
+        result = subprocess.run(
+            ["git", "rev-parse", "HEAD"],
+            text=True,
+            encoding="utf8",
+            capture_output=True,
+            check=True,
+            ).stdout.strip()
+
+    if os.environ.get("READTHEDOCS", "") == "True":
+        rtd_version = os.environ.get("READTHEDOCS_VERSION", "")
+        if rtd_version == 'latest':
+            # This is the RTD build for the main branch.
+            result = main_branch
+        elif "." in rtd_version:
+            # It looks like we have a tag build.
+            result = rtd_version
+
+    if debug:
+        print("== Environment dump for {} {} ===".format(
+            reference_name, result))
+        pprint(dict(os.environ))
+        print("======")
+
+    return result
--- a/pydoctor/sphinx_ext/build_apidocs.py
+++ b/pydoctor/sphinx_ext/build_apidocs.py
@@ -9,8 +9,12 @@ Inside the Sphinx conf.py file you need to define the following configuration op
   - C{pydoctor_args} - an iterable with all the pydoctor command line arguments used to trigger the build.
                      - (private usage) a mapping with values as iterables of pydoctor command line arguments.
 
+  - C{pydoctor_git_reference} - The branch name or SHA reference for current build.
+  - C{pydoctor_debug} - C{True} if you want to see extra debug message for this extension.
+
 The following format placeholders are resolved for C{pydoctor_args} at runtime:
   - C{{outdir}} - the Sphinx output dir
+  - C{{git_reference}} - the Git reference that can be used for source code links.
 
 You must call pydoctor with C{--quiet} argument
 as otherwise any extra output is converted into Sphinx warnings.
@@ -26,7 +30,7 @@ from sphinx.util import logging
 
 from pydoctor import __version__
 from pydoctor.driver import main
-
+from pydoctor.sphinx_ext import get_git_reference
 
 logger = logging.getLogger(__name__)
 
@@ -41,6 +45,10 @@ def on_build_finished(app: Sphinx, exception: Exception) -> None:
 
     placeholders = {
         'outdir': app.outdir,
+        'git_reference': get_git_reference(
+            main_branch=config.pydoctor_main_branch,
+            debug=config.pydoctor_debug,
+            ),
         }
 
     runs = config.pydoctor_args
@@ -84,6 +92,8 @@ def setup(app: Sphinx) ->  Mapping[str, Any]:
     """
     app.connect('build-finished', on_build_finished)
     app.add_config_value("pydoctor_args", None, "env")
+    app.add_config_value("pydoctor_main_branch", "main", "env")
+    app.add_config_value("pydoctor_debug", False, "env")
 
     return {
         'version': str(__version__),
--- a/tox.ini
+++ b/tox.ini
@@ -144,5 +144,12 @@ description = Build the full documentation
 
 extras = docs
 
+setenv =
+    # Set this to `True` to run similar to Read the docs.
+    READTHEDOCS=False
+    READTHEDOCS_LANGUAGE=en
+    # Possible version: 1.2.3 stable latest
+    READTHEDOCS_VERSION=1.0.0.dev0
+
 commands =
     sphinx-build -aE -b html {toxinidir}/docs/source {toxinidir}/build/docs
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -66,15 +66,11 @@ html_theme = "sphinx_rtd_theme"
 # so a file named "default.css" will overwrite the builtin "default.css".
 html_static_path = []
 
-# Try to find URL fragment for the GitHub source page based on current
-# branch or tag.
-_git_reference = _get_git_reference(main_branch='master', debug=True)
-
 _pydoctor_root = pathlib.Path(__file__).parent.parent.parent
 _common_args = [
     '--quiet',
     '--make-html',
-    f'--html-viewsource-base=https://github.com/twisted/pydoctor/tree/{_git_reference}',
+    '--html-viewsource-base=https://github.com/twisted/pydoctor/tree/{git_reference}',
 
     '--project-url=https://github.com/twisted/pydoctor/',
 
@@ -99,3 +95,5 @@ pydoctor_args = {
         f'{_pydoctor_root}/docs/epytext_demo',
         ] + _common_args,
     }
+pydoctor_main_branch = 'master'
+pydoctor_debug = True
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -14,9 +14,9 @@
 # import sys
 # sys.path.insert(0, os.path.abspath('.'))
 
-import os
-import subprocess
 import pathlib
+
+from pydoctor.sphinx_ext import get_git_reference as _get_git_reference
 
 
 # -- Project information -----------------------------------------------------
@@ -68,17 +68,7 @@ html_static_path = []
 
 # Try to find URL fragment for the GitHub source page based on current
 # branch or tag.
-_git_reference = subprocess.getoutput('git rev-parse --abbrev-ref HEAD')
-if _git_reference == 'HEAD':
-    # It looks like the branch has no name.
-    # Fallback to commit ID.
-    _git_reference = subprocess.getoutput('git rev-parse HEAD')
-
-if os.environ.get('READTHEDOCS', '') == 'True':
-    rtd_version = os.environ.get('READTHEDOCS_VERSION', '')
-    if '.' in rtd_version:
-        # It looks like we have a tag build.
-        _git_reference = rtd_version
+_git_reference = _get_git_reference(main_branch='master', debug=True)
 
 _pydoctor_root = pathlib.Path(__file__).parent.parent.parent
 _common_args = [


mthuurne:
Sounds useful, but I'd prefer to have `git` support as an option instead of making it part of the default flow. For example, only run `get_git_reference()` if `{git_reference}` actually occurs in the passed args.

pydoctor/sphinx_ext/__init__.py
```diff
@@ -1,3 +1,58 @@
...
+            ).stdout.strip()
+
+    if os.environ.get("READTHEDOCS", "") == "True":
+        rtd_version = os.environ.get("READTHEDOCS_VERSION", "")
+        if rtd_version == 'latest':
+            # This is the RTD build for the main branch.
+            result = main_branch
+        elif "." in rtd_version:
+            # It looks like we have a tag build.
+            result = rtd_version
```
mthuurne, line 50: This code is specific to RTD, which supports other source control systems as well, so I think it would make sense to move it out of `get_git_reference()`.

pydoctor/sphinx_ext/__init__.py
```diff
@@ -1,3 +1,58 @@
 """
 Public and private extensions for Sphinx.
 """
+import os
+import subprocess
+from pprint import pprint
+
+
+def get_git_reference(main_branch: str = 'master', debug: bool = False, ) -> str:
```
mthuurne, line 9:
```suggestion
def get_git_reference(main_branch: str = 'master', debug: bool = False) -> str:
```
Note that if the RTD code is moved out (see below), we don't need `main_branch` as an argument anymore.

pydoctor/sphinx_ext/__init__.py
mthuurne: Should we even map `latest` to a branch name? The reason we fetch the revision number is to get a non-moving target to link to and overriding that with a branch name means we're back to linking to a moving target.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -9,8 +9,12 @@
   - C{pydoctor_args} - an iterable with all the pydoctor command line arguments used to trigger the build.
                      - (private usage) a mapping with values as iterables of pydoctor command line arguments.
 
+  - C{pydoctor_main_branch} - The name of the main branch name. Used on RTD for latest build.
+  - C{pydoctor_debug} - C{True} if you want to see extra debug message for this extension.
+
 The following format placeholders are resolved for C{pydoctor_args} at runtime:
   - C{{outdir}} - the Sphinx output dir
+  - C{{git_reference}} - the Git reference that can be used for source code links.
```
mthuurne, line 17: Maybe call it `source_revision` or something similar? Since it doesn't have to originate from `git` when using RTD.

tox.ini
```diff
@@ -144,5 +144,12 @@ description = Build the full documentation
 
 extras = docs
 
+setenv =
+    # Set this to `True` to run similar to Read the docs.
+    READTHEDOCS=False
+    READTHEDOCS_LANGUAGE=en
+    # Possible version: 1.2.3 stable latest
```
mthuurne, line 151: We don't actually support `stable` at the moment, I think.

Does RTD literally pass `stable` in `READTHEDOCS_VERSION` or does it pass the release tag instead?

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -9,8 +9,12 @@
   - C{pydoctor_args} - an iterable with all the pydoctor command line arguments used to trigger the build.
                      - (private usage) a mapping with values as iterables of pydoctor command line arguments.
 
+  - C{pydoctor_main_branch} - The name of the main branch name. Used on RTD for latest build.
```
mthuurne, line 12:
```suggestion
  - C{pydoctor_main_branch} - The name of the main branch name. Used when building the C{latest} version on ReadTheDocs.
```

mthuurne (changes requested):
This introduces two ways of getting the source reference: from Git or from RTD env vars. I'd prefer to see those two sources as separated as possible:
- such that projects that don't use Git can still use the RTD features
- to keep the code simpler by separation of concerns

Getting the source revision is a problem a lot of projects are going to encounter, so offering an implementation for that is very useful. In fact, we could consider to go one step further and put the Git support in pydoctor itself rather than in the Sphinx extension. That way, it would also be available to projects that don't use Sphinx.

pydoctor/sphinx_ext/__init__.py
adiroiban: I will look at moving the code outside of git.

Regarding 'latest' and main branch... it was an attempt :) I agree that it's best to link to SHA revision for latest :)

tox.ini
adiroiban: It literally pass "stable".

My plan is to also support stable and any other tag as part of #330

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: This will be removed... and we will not link to main branch name... it should simplify the config.

pydoctor/sphinx_ext/__init__.py
adiroiban: Yes. I plan to completely remove the idea of "main_branch" and just link to SHA

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -57,6 +53,21 @@ def on_build_finished(app: Sphinx, exception: Exception) -> None:
...
         runs = {'main': runs}
 
+    # Defer resolving the git reference for only when asked by
+    # end users.
+    is_git_needed = False
+    for key, value in runs.items():
+        for arg in value:
+            if '{git_reference}' in arg:
+                is_git_needed = True
+                break
```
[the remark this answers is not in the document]
adiroiban: I will apply this if you want. Is just a matter of style.

I prefer to avoid one liners, as they are harder to read. You write once, but the read the code many times... and oftern you read it when you are tired of trying to hunt down a bug.

Some might argue that functional programming is better ... I argue that easy to read (even for a junior) programming is better :)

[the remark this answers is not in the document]
mthuurne: I put it on one line since it was short enough, but it's fine to spread it over multiple lines if you think that makes it more readable.

The important part is `any()`, since that expresses what we're actually trying to compute: is there any argument containing `{git_reference}`?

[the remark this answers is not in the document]
adiroiban: I went with your suggestion as it's not that hard to read..and maybe even easier to read :) Thanks!

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: I went with source_reference

adiroiban:
Thanks for a great review :) I think that the code looks better after your comments.

Please give it another look.

Thanks!

pydoctor/sphinx_ext/__init__.py
```diff
@@ -1,3 +1,51 @@
...
 Public and private extensions for Sphinx.
 """
+import os
+import subprocess
+
+
+def get_source_reference() -> str:
+    """
+    Return the reference for current build to be used when linking
+    the source files.
```
mthuurne, line 11: And what do we do when we can't get a source reference?

Since this function is only called when `{source_reference}` was specified by the user, I think the correct response would be to fail the build if we can't get a reference. But it would be good to do that with a short and clear message rather than a stack trace. So I think this function should raise an exception and the caller should catch and report it.

mthuurne, line 11:
```suggestion
    Return a reference to the particular source version used in this build.
    This can be used in the URL passed with C{--html-viewsource-base}, to link to the
    version of the source code that corresponds with the generated documentation.
```

mthuurne (changes requested):
Thanks for cleaning up the code; it's easier to understand now.

I think the handling of errors from `git rev-parse` should be documented and implemented before merging. It doesn't have to be fancy: just logging an error and failing the build is fine.

Looking at the implementation, this feature doesn't seem to have any particular interaction with Sphinx, so it may be better to support this placeholder in pydoctor itself. That would make it available to more projects.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -33,26 +38,26 @@
 
 from pydoctor import __version__
 from pydoctor.driver import main, parse_args
-
+from pydoctor.sphinx_ext import get_source_reference
 
 logger = logging.getLogger(__name__)
 
+# Shared state between init and finish.
+_placeholders: Dict[str, str] = {}
```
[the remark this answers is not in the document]
adiroiban: That was my first try... but I think it's a hack.

I think is better not to touch the Sphinx App or Config ...and mypy will also complain :) 

Even if the main deal is moved outside of sphinx, the placeholder injection thing will still be used in the sphinx extension.

adiroiban:
Thanks for your great review :) I will update the code and push for another review.

pydoctor/sphinx_ext/build_apidocs.py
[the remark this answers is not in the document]
mthuurne: > That was my first try... but I think it's a hack.

Using global state is also a bit of a hack, since that assumes that there will be at most one build operating at the same time. I don't see a likely scenario in which that is not the case, but there is also no guarantee that there won't be multiple builds in the future.

> I think is better not to touch the Sphinx App or Config ...and mypy will also complain :)

Not touching `app` might be a good idea, since that is data that belongs to Sphinx. But the `config` object already contains attributes that are named by us such as `pydoctor_args`, so I don't think there is a problem adding attributes there.

To make mypy accept it, `Config.__setitem__()` can be used:
```py
    config['placeholders'] = placeholders
```

> Even if the main deal is moved outside of sphinx, the placeholder injection thing will still be used in the sphinx extension.

True, we still need it to inject `outdir`. However, `source_reference` is the only placeholder we currently have that is heavy to compute, so without that, it would also be an option to just reconstruct the placeholders in `on_build_finished()`. But it's possible we'll get placeholders in the future that are non-trivial to compute and don't belong in pydoctor itself, so storing the placeholders is probably the better solution.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -86,6 +91,16 @@ def on_config_inited(app: Sphinx, config: Config) -> None:
...
         runs = {'main': runs}
 
+    # Defer resolving the git reference for only when asked by
+    # end users.
+    is_source_reference_needed = any(
+        '{source_reference}' in arg
+        for value in runs.values() for arg in value
+        )
+
+    if is_source_reference_needed:
```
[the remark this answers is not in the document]
adiroiban: I prefer to keep the variable as it servers as documentation for that logical expression.

pydoctor/sphinx_ext/__init__.py
adiroiban: For example, when git is not found Sphinx build fails with this error.

```
Handler <function on_config_inited at 0x7fe5f7f13280> for event 'config-inited' threw an exception (exception: Failed to get git reference. [Errno 2] No such file or directory: 'git')

```

adiroiban:
Thanks for the great review. I think this is ready for another round of review.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -38,26 +39,27 @@
...
-_placeholders: Dict[str, str] = {}
-
 
 def on_build_finished(app: Sphinx, exception: Exception) -> None:
     """
     Called when Sphinx build is done.
     """
+
+    # Share placeholders between init and finish.
+    placeholders = cast(Mapping[str, str], app.config['placeholders'])
```
mthuurne, line 53: Since `__getitem__()` returns `Any`, you don't need a cast: you can assign to an annotated local variable instead:
```suggestion
    placeholders: Mapping[str, str] = app.config['placeholders']
```
This is a "softer" correction in the sense that `cast` overrides everything mypy knows about a type, while this annotation adds information without overriding what mypy already knows.

In this case it probably doesn't matter much, since I don't think `__getitem__()` will ever be able to return anything other than `Any`, but in cases where a value is `Any` because the library used doesn't have annotations yet, the variant without a cast is a bit safer.

pydoctor/sphinx_ext/__init__.py
mthuurne: Hmm, that's nowhere near as bad as a backtrace, but still slightly ugly.

In any case, I see you're moving the code from the extension into pydoctor itself, so the error reporting will have to be implemented in pydoctor as well, as we can't rely on Sphinx doing it anymore.

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: Thanks. Makes sense. I am still learning mypy. Many thanks for your support. Much appreciated :)

But if  try you suggestion I get 

```
pydoctor/sphinx_ext/build_apidocs.py:55: error: Incompatible types in
assignment (expression has type "str", variable has type "Mapping[str, str]") 
[assignment]
        placeholders: Mapping[str, str] = app.config['placeholders']
                                          ^
Found 1 error in 1 file (checked 43 source files)
```

mthuurne: Hmm, for some reason `sphinx.config.Config.__getitem__()` is annotated to return `str`, while `__setitem__()` accepts `Any`. I don't know if this is intentional or a mistake in their annotations.

You can work around it by using `__getattr__()` instead, which does return `Any`:
```py
    placeholders: Mapping[str, str] = app.config.placeholders
```

adiroiban:
Thanks. This should be ready for a re-review.
`$ tox -qp all -e test,mypy,pyflakes,apidocs,docs` is all green on my laptop

pydoctor/util.py
```diff
@@ -0,0 +1,64 @@
...
+    It tries the extract the tag name from Read The Docs environment and
+    falls back to source code revision.
+
+    Only git VSC is supported for now.
+
+    @raises RuntimeError: If source reference couldn't be retrieved.
+    """
+    if os.environ.get("READTHEDOCS", "") == "True":
+        rtd_version = os.environ.get("READTHEDOCS_VERSION", "")
+        if rtd_version not in ['latest', 'stable']:
```
mthuurne, line 23: I generally prefer tuples over lists if mutability is not required. Maybe there could be runtime optimizations done for tuples that aren't done for lists, but I don't know if that is actually the case for CPython or pypy.

pydoctor/test/test_util.py
```diff
@@ -0,0 +1,48 @@
...
+    assert result == 'V_1_2'
+
+
+def test_get_source_reference_RTD_latest(monkeypatch: MonkeyPatch) -> None:
+    """
+    When the RTD environment version is `latest` it returns the git
+    branch name or SHA.
+    """
+    monkeypatch.setenv("READTHEDOCS", "True")
+    monkeypatch.setenv("READTHEDOCS_VERSION", "latest")
```
mthuurne, line 29: We also treat `stable` in a special way, so maybe the test could be parametrized.

pydoctor/test/test_util.py
```diff
@@ -0,0 +1,48 @@
...
+    assert result != "latest"
+
+
+def test_get_source_reference_git_fail(monkeypatch: MonkeyPatch) -> None:
+    """
+    When getting the git reference fails, it raised RunntimeError.
+    """
+    monkeypatch.setenv("READTHEDOCS", "False")
+    # Try to move to a dir that has not git parents.
+    monkeypatch.chdir('/')
```
mthuurne, line 45: Can we be sure of that? If the test runs in some kind of chroot or container, the root dir might have a `.git` subdir.

I would prefer to intercept the attempt to run the `git` command; see comment above. Then we can just raise an exception from the interception point.

mthuurne (changes requested):
It's getting closer, but I don't think it's quite ready to merge yet. The main issues are:
- the unit tests shouldn't be running `git`: either we should intercept the command or we should make this into a system test; see inline comments for details
- having the placeholder substitution for `{source_reference}` inside pydoctor itself would make the feature available also to projects that don't use Sphinx

The second point is not necessary for releasing Twisted, but would make pydoctor more useful in general. Since some of my projects would benefit from it, I'm willing to finish that part of the implementation if you're not motivated to.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -85,13 +92,25 @@ def on_config_inited(app: Sphinx, config: Config) -> None:
     runs = config.pydoctor_args
 
     if not isinstance(runs, Mapping):
-        # We have a single pydoctor call
+        # We have a single pydoctor call.
         runs = {'main': runs}
 
+
```
adiroiban, line 98: Should be removed. My bad.

```suggestion
```

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -85,13 +92,25 @@ def on_config_inited(app: Sphinx, config: Config) -> None:
...
     pydoctor_url_path = config.pydoctor_url_path
     if not isinstance(runs, Mapping):
+    # We have a single pydoctor inventory mapping.
```
adiroiban, line 101: My bad.

```suggestion
        # We have a single pydoctor inventory mapping.
```

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -85,13 +92,25 @@ def on_config_inited(app: Sphinx, config: Config) -> None:
...
 
+    # Defer resolving the git reference for only when asked by
+    # end users.
+    is_source_reference_needed = any(
+        '{source_reference}' in arg
+        for value in runs.values() for arg in value
+        )
+
+    if is_source_reference_needed:
+        placeholders['source_reference'] = get_source_reference()
```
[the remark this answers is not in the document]
adiroiban: I will do that here.

pydoctor/test/test_util.py
```diff
@@ -0,0 +1,48 @@
...
+    branch name or SHA.
+    """
+    monkeypatch.setenv("READTHEDOCS", "True")
+    monkeypatch.setenv("READTHEDOCS_VERSION", "latest")
+
+    result = get_source_reference()
+
+    # This is a system integration test and is hard to assert the actual
+    # value here.
+    # On local build you get the branch name, on CI you might get the SHA.
```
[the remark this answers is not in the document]
adiroiban: I prefer to run system/integration test whenever possible, as that is "the real deal".

What we can do, is skip the test if git command is not installed, and not running in a repo.

for me, mock / patch is the enemy and I try to work as little as possible with my enemies :p

I will give it a try.

mthuurne:
@adiroiban: Are you planning to finish this PR, or would it be better if someone else takes over?

adiroiban:
Hi.  Sorry for the delay. It would be great for someone else to take over this PR.

I will not have time to work on this in the next months :(

glyph:
@adiroiban I see a lot of conflicts, no time for you to work on it, and a bunch of unaddressed feedback. Is it time to close this, and reopen it as an issue that describes the problem rather than letting the code sit around unaddressed?

adiroiban:
ok. closing this.

PR #329, opened 2020-12-11
Have all contributing info in a single file.

adiroiban:
This is just a small update.

I think is best to have all contributing related info in  a single file..and not split across multiple files.

Moved design notes in the root file


mthuurne:
I think it makes sense to have it all in a single file, but I'd prefer that file under `docs`.

I know for example GNU has several all-caps files in the top dir, but that convention made more sense when people were reading docs with `less` instead of a web browser.

adiroiban:
> I know for example GNU has several all-caps files in the top dir, but that convention made more sense when people were reading docs with less instead of a web browser.

I still read a lot of documentation using a text editor and not the web browser :)

Also, having the Contributing file in a known root directly will have it implicitly linked as part of PR GUI ... but the link is super small...so we might want to make it explicit and then we can have it anywhere. 

image

I will move it into docs and remove the CONTRIBUTING.rst file.

CONTRIBUTING.rst
```diff
@@ -1,29 +1 @@
...
-
-First commit the version update to master and wait for tests to pass.
-
-You can push a tab by creating a `GitHub Release <https://github.com/twisted/pydoctor/releases/new>`
-
-You can also create a tag on local branch then push it::
-
-        git tag 1.2.3
-        git push --tags
+See `<docs/source/contrib.rst>`_
```
mthuurne, line 1: Does this work with the PR GUI? If not, maybe put in a RTD URL instead.

(Not that I ever noticed it in the PR GUI until you pointed it out, but if it's there, it would be good to have a working link.)

mthuurne (approved):
Looks good.

One question about how the minimal remaining top-level file will interact with the GitHub PR GUI, see inline comment.

CONTRIBUTING.rst
adiroiban: This works with GitHub browser https://github.com/twisted/pydoctor/blob/single_contribution_file/CONTRIBUTING.rst

PR #330, opened 2020-12-11
Have own Sphinx object inventory available at Sphinx build time.

adiroiban:
Changes
=======

Create the Sphinx object inventory for each pydoctor target before all the plugins are loaded.
In this way, we have the object before the intersphinx loads.

Intersphinx plugin is designed to load the inventory from the source directory...but we want it from build.

Intersphinx is designed to use  absolute URL as it is used for external sites... but here we have an interrnal site link.

How to test
===========

```
rm -rf build/docs
tox -e doc
firefox build/docs/usage.html
```
Or on Read The Docs 
https://pydoctor--330.org.readthedocs.build/en/330/usage.html#building-pydoctor-together-with-sphinx-html-build

Check the link at `pydoctor.sphinx_ext.build_apidocs`  - It should be an absolute path to the site... but not an absolute URL.


adiroiban:
Maybe by not creating  that many HTML files, we can speed up the process enough to allow calling  pydoctor twice.

The option option is to build pydoctor at the start in a separate directory, and then move it to the final directory at the end... but it will need parsing t he arguments to get the target dir... and renaming it to .TMP and during the build ... and then renaming at the end.
 ... it is possible :)


But maybe we can go with this, and see what performance we have.

tristanlatr:
Little question : `pydoctor_intersphinx_mapping`. What is the use case for the users to use this variable? If I understand correctly it only to use pydoctor's own intersphinx in our own documentation. So users wont need that right ? The variable should be private no ?

adiroiban:
@tristanlatr I tried to document that variable in pydoctor/sphinx_ext/build_apidocs.py  :) ...looks like the docs is not good enough. That is ok and this is why we have the review :)

pydoctor_intersphinx_mapping is needed as it does the same thing as intersphinx_mapping but the has a completely different semantics. For example, relative paths are relative to the build, and not to the source dir.
Also, they support the RTD specific placeholder.

adiroiban:
To answer you question.

The variable allow to use your own apidoc inventory inside Sphinx.

This is not just private pydoctor usage.

If your project had both apidoc generated  by pydoctor and narrative docs by sphinx this allows to reference your API inside Sphinx.

tristanlatr:
> If your project had both apidoc generated by pydoctor and narrative docs by sphinx this allows to reference your API inside Sphinx.

Thanks a lot!

adiroiban:
In theory, I can also hack the `intersphinx_mapping` to have a single conflig like this.... but I prefer not to modify the exiting `intersphinx_mapping` values...but only add new values based on pydoctor config

```
# Configure intersphinx magic
intersphinx_mapping = {
    'twisted': ('https://twistedmatrix.com/documents/current/api/', None),
   'pydoctor': ('/en/{rtd_version}/api/', '{outdir)/api/objects.inv'),
}
```

> Thanks a lot!

I think that the action based on your comments is to update the narrative docs.

I wanted to delay writing the narrative docs, only after the initial review ... but I did it now.

Let me know if it's easier to understand what the extension does and how to use it.

Thanks

tristanlatr:
I've read the documentation and yes It's clear!

Here are some thoughts : 

- Docs for linking from sphinx to pydoctor is spread across the sections "Linking from Sphinx to your pydoctor API docs" and "Building pydoctor together with Sphinx HTML build or on Read The Docs". This is because you first explain 

- "Building pydoctor together with Sphinx HTML build or on Read The Docs": it is a too long title if you ask me. Why not have a subsection called "Using with Read The Docs". 

- The `usage.rst` file is getting large, we should consider breaking it up into multiple files. 

- Overall I feel that we should restructure this part of the docs as follow : 

1. Sphinx integration
    1. Linking from pydoctor to external API docs
    2. Using Sphinx and pydoctor
        1. Building pydoctor together with Sphinx build
        2. Linking from Sphinx to your own pydoctor API docs
        3. Using with Read The Docs

- `pydoctor_intersphinx_mapping` should be presented as the default (best) way to link from sphinx to your own pydoctor apidocs right from the beginning, no?

- You show usage both with and without `pydoctor_intersphinx_mapping` in the `conf.py`. Is it necessary ? Who will actually trigger the inventory creation manually ?

- There is a typo, first line of the `conf.py` exemple : `extensions.append("pydoctor.sphinx_ext.build_apidocs)` -> `extensions.append("pydoctor.sphinx_ext.build_apidocs")`

- It's written `'/absolute/path/to/your/project/package1'`, I don't think the paths need to be absolute. (I know it's not related to this PR but I just saw this)

adiroiban:
Thanks for the review.
I will try to see how I can best reorganize usage.rst and fix the typos.

--------

If you don't build pydoctor in Read The Docs, you don't really need `pydoctor_intersphinx_mapping`
and you don't really need the whole pydoctor extension for Sphinx :)


so `pydoctor_intersphinx_mapping` is needed only if you build using the extension and only if you need the link the API from Sphinx.

--------

> You show usage both with and without pydoctor_intersphinx_mapping in the conf.py. Is it necessary ? Who will actually trigger the inventory creation manually ?

In which conf.py you see with and without `pydoctor_intersphinx_mapping` ?

------------

docs/source/conf.py
```diff
@@ -51,7 +50,6 @@
 # Configure intersphinx magic
 intersphinx_mapping = {
     'twisted': ('https://twistedmatrix.com/documents/current/api/', None),
-    'pydoctor': ('https://pydoctor.readthedocs.io/en/latest/api/', None),
```
adiroiban, line 54: Note that this URL was wrong...as when building on Read The docs for the "stable" version, you will still have links to "latest"

With this PR, this should be fixed. 

Once we have a new release, we should check that we have separate API links for

See `You can subclass the pydoctor.zopeinterface.ZopeInterfaceSystem and pass your custom class `

stable vs  latest

* https://pydoctor.readthedocs.io/en/latest/usage.html#use-a-custom-system-class
* https://pydoctor.readthedocs.io/en/stable/usage.html#use-a-custom-system-class

Right now the API links is always latest.

tristanlatr:
> In which conf.py you see with and without pydoctor_intersphinx_mapping ?

From line 178 to 186 (usage.rst) it's explained who to use `intersphinx_mapping` with our own project, then , the `pydoctor_intersphinx_mapping` variable is introduced.

adiroiban:
> From line 178 to 186 (usage.rst) it's explained who to use `intersphinx_mapping` with our own project, then , the `pydoctor_intersphinx_mapping` variable is introduced.

I will try to see what I can do to improve this. Thanks :)

The idea is that first explains how to use just Sphinx and second is about using Sphinx + pydoctor-sphinx-extension ... but I will see

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -65,7 +77,8 @@ def _run_pydoctor(name: str, arguments: Iterable[str], placeholders: Mapping[str
     for argument in arguments:
         args.append(argument.format(**placeholders))
 
-    logger.info(f"Building '{name}' pydoctor API docs as:")
+    build_type = next(iter(arguments))
```
mthuurne, line 80: I'm not a fan of this:
- this permanently eats the first argument if `arguments` is a generator
- the first argument having any special meaning is not part of this method's documented interface

The first concern can be resolved by changing the type annotation from `Iterable[str]` to `Sequence[str]`, which will disqualify generators. After this change you can just look up `arguments[0]`.

However, as the full list of arguments is logged anyway, do we really need to report the first argument separately?

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -76,14 +89,55 @@ def _run_pydoctor(name: str, arguments: Iterable[str], placeholders: Mapping[str
...
+    for key, value in config.pydoctor_intersphinx_mapping.items():
+        url, target_path = value
+
+        url = url.format(**{'rtd_version': rtd_version})
+        inventory_path = output_dir / target_path
+
+        intersphinx_mapping[key] = (url, str(inventory_path))
+
+    if not config.pydoctor_args:
+        raise ConfigError("Missing 'pydoctor_args'.")
```
mthuurne, line 112: It would be cleaner to perform input validation before making any state changes.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -76,14 +89,55 @@ def _run_pydoctor(name: str, arguments: Iterable[str], placeholders: Mapping[str
...
     Called by Sphinx when the extension is initialized.
 
     @return: The extension version and runtime options.
     """
-    app.connect('build-finished', on_build_finished)
     app.add_config_value("pydoctor_args", None, "env")
+    app.add_config_value("pydoctor_intersphinx_mapping", {}, "env")
+
+    # Make sure we have a lower priority than intersphinx extension.
+    app.connect('config-inited', update_intersphinx_mapping, priority=799)
```
mthuurne, line 138: Is the intersphinx extension at priority 800? If so, maybe use 790 or something, just in case there is ever an extension that wants to go between `build_apidocs` and intersphinx.

mthuurne (commented):
Why do we need to passes? Of course the inventory must be created before intersphinx runs, but can't we generate the HTML at the same time?

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: True. That was stupid thing to do :) Thanks!

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: True. This was updated.

But the build was existing anyway.

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: True. Thanks!

adiroiban (commented):
Thanks for the excellent review. 

I have made an initial update on the docs and UI/UX.

pydoctor is now only called once.

let me know if you are happy with the new design and documentation.

adiroiban:
There is this mypy error:

```
pydoctor/sphinx_ext/build_apidocs.py:46: error: Cannot determine type of 'config'  [has-type]
        runs = app.config.pydoctor_args
```

CONTRIBUTING.rst
```diff
@@ -8,6 +8,7 @@ If you like the project and think you could help with making it better, there ar
 - Triage old issues that needs a refresh
 - Implement existing issues (there are quite some of them, choose whatever you like)
 - Help with improving the documentation (We still have work to do!)
+- Help with improving the CLI documentation at :py:func:`pydoctor.driver.getparser`
```
adiroiban, line 11: btw... this can be removed, as I have introduced another intershphinx link in usage.rst

adiroiban:
This is still WIP. The `tox -e docs` works on my system but fails on read the docs.

mthuurne:
The mypy error can be fixed by requiring Sphinx 3.4.0 or later, since the code that confused mypy was fixed in 3.4.0. See sphinx-doc/sphinx#8398 for details.

pydoctor/sphinx_ext/build_apidocs.py
mthuurne: Yeah, it might not matter here, it's just a general good habit.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -50,40 +88,70 @@ def on_build_finished(app: Sphinx, exception: Exception) -> None:
         runs = {'main': runs}
 
     for key, value in runs.items():
-        _run_pydoctor(key, value, placeholders)
+        arguments = _get_arguments(value, placeholders)
+
+        options, _ = parse_args(arguments)
+        output_path = pathlib.Path(options.htmloutput)
+        temp_path = pathlib.Path(options.htmloutput + '_pydoctor_temp')
```
mthuurne, line 95: Same remark about trailing slash.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -50,40 +88,70 @@ def on_build_finished(app: Sphinx, exception: Exception) -> None:
...
+        output_path.rename(temp_path)
+
+def _run_pydoctor(name: str, arguments: Sequence[str]) -> None:
     """
     Call pydoctor with arguments.
 
     @param name: A human-readable description of this pydoctor build.
-    @param arguments: Iterable of arguments used to call pydoctor.
-    @param placeholders: Values that will be interpolated with the arguments using L{str.format()}.
+    @param arguments: Sequence of arguments used to call pydoctor.
```
mthuurne, line 114:
```suggestion
    @param arguments: Command line arguments used to call pydoctor.
```
Now that pydoctor presents the type annotation, we no longer need to include the type in the text itself ;)

docs/source/usage.rst
```diff
@@ -195,48 +205,53 @@ Possible links are::
...
+    pydoctor_url_fragment = '/en/{rtd_version}/api/'
+
+You can pass almost any argument to `pydoctor_args`
+in the same way you call `pydoctor` from the command line.
 
-You can pass any argument, in the same way you call `pydoctor` from the
-command line.
+You should not pass the `--make-html`, `--make-intersphinx` or `--quiet`
+arguments as this will break the build.
+The extension will add them automatically.
```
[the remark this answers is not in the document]
mthuurne: For my understanding: `--make-html` and `--quiet` are auto-added, while `--make-intersphinx` is not auto-added, but having that option enabled would break the output because no HTML would be produced by pydoctor, correct?

mthuurne (approved):
I found a few issues. After addressing those, I think it's good to merge.

CONTRIBUTING.rst
adiroiban:
```suggestion
```

docs/source/conf.py
```diff
@@ -108,3 +103,5 @@
         f'{_pydoctor_root}/docs/epytext_demo',
         ] + _common_args,
     }
+
+pydoctor_url_fragment = '/en/{rtd_version}/api/'
```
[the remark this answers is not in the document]
adiroiban: ok. I will use path
```suggestion
pydoctor_url_path = '/en/{rtd_version}/api/'
```

docs/source/usage.rst
```diff
@@ -167,14 +167,24 @@ pydoctor's HTML generator will also generate a Sphinx objects inventory that can
...
+You can use the `--make-intersphinx` option to only generate the object inventory file.
+
+You will then reference this file inside the Sphinx `intersphinx_mapping`.
+
+Note that relative paths are relative to the Sphinx source directory.
+You might need to exit the source and reference the build directory::
 
     intersphinx_mapping = {
-        'twisted': ('https://twistedmatrix.com/documents/current/api/', None),
+        'twisted': ('https://twistedmatrix.com/documents/current/api/', '../../build/apidocs/objects.inv'),
```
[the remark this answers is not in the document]
adiroiban: Yes. This is docs for the first case ... but it's about using pydoctor and sphinx without the extra pydoctor-extensison.

for pydoctor + Sphinx you no longer need a local path. The extension takes are of automatically detecting and injecting that path.

docs/source/usage.rst
```diff
@@ -195,48 +205,53 @@ Possible links are::
   :py:attr:`twisted:twisted.protocols.amp.BinaryBoxProtocol.boxReceiver`
 
 
-Building pydoctor together with Sphinx HTML build
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Building pydoctor together with Sphinx HTML build or on Read The Docs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
adiroiban, line 209:
```suggestion
Building pydoctor together with Sphinx HTML build
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

docs/source/usage.rst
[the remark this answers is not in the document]
adiroiban: Well. Adding those arguments will not break the build .. at least for now.

So if you add them, nothing would happen.

But to leave some space for future changes, I prefer to document it in this way.

docs/source/usage.rst
```diff
@@ -195,48 +205,53 @@ Possible links are::
...
         '--html-viewsource-base=https://github.com/ORG/REPO/tree/default',
         '--html-output={outdir}/api',
-        '--project-base-dir=/absolute/path/to/your/project',
-        '/absolute/path/to/your/project/package1'
+        '--project-base-dir=path/to/source/code',
+        'path/to/source/code/package1'
         ]
 
-The `{outdir}` will replaced with the Sphinx build dir.
+    pydoctor_url_fragment = '/en/{rtd_version}/api/'
```
adiroiban, line 235:
```suggestion
    pydoctor_url_path = '/en/{rtd_version}/api/'
```

docs/source/usage.rst
```diff
@@ -195,48 +205,53 @@ Possible links are::
...
 
-You can pass any argument, in the same way you call `pydoctor` from the
-command line.
+You should not pass the `--make-html`, `--make-intersphinx` or `--quiet`
+arguments as this will break the build.
+The extension will add them automatically.
 
-The `--quiet` argument is recommend, as any output produced by pydoctor is
-converted into Sphinx warnings.
+The `pydoctor_url_fragment` is an URL fragment,
```
adiroiban, line 244:
```suggestion
The `pydoctor_url_path` is an URL path,
```

docs/source/usage.rst
```diff
@@ -195,48 +205,53 @@ Possible links are::
...
+relative to your public API documentation site.
+`{rtd_version}` will be replaced with the Read The Docs version.
```
adiroiban, line 246:
```suggestion
`{rtd_version}` will be replaced with the Read The Docs version (`stable` , `latest`, tag name).
```

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -6,26 +6,34 @@
...
 Inside the Sphinx conf.py file you need to define the following configuration options:
 
-  - C{pydoctor_args} - an iterable with all the pydoctor command line arguments used to trigger the build.
-                     - (private usage) a mapping with values as iterables of pydoctor command line arguments.
+  - C{pydoctor_url_fragment} defined the URL path to the API documentation
+    You can use C{{rtd_version}} to have the URL automatically updated
+    based on Read The Docs build.
+
+  - C{pydoctor_args} - an sequence with all the pydoctor command line arguments used to trigger the build.
+                       Don't include the C{--make-html} argument, as it will be added automatically.
```
adiroiban, line 14:
```suggestion
  - C{pydoctor_args} - Sequence with all the pydoctor command line arguments used to trigger the build.
```

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -50,40 +88,70 @@ def on_build_finished(app: Sphinx, exception: Exception) -> None:
...
+
 def setup(app: Sphinx) ->  Mapping[str, Any]:
     """
     Called by Sphinx when the extension is initialized.
 
     @return: The extension version and runtime options.
     """
-    app.connect('build-finished', on_build_finished)
     app.add_config_value("pydoctor_args", None, "env")
+    app.add_config_value("pydoctor_url_fragment", "", "env")
```
adiroiban, line 149:
```suggestion
    app.add_config_value("pydoctor_url_path", "", "env")
```
```suggestion
    app.add_config_value("pydoctor_url_fragment", "", "env")
```

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -50,40 +88,70 @@ def on_build_finished(app: Sphinx, exception: Exception) -> None:
...
+
+        options, _ = parse_args(arguments)
+        output_path = pathlib.Path(options.htmloutput)
+        temp_path = pathlib.Path(options.htmloutput + '_pydoctor_temp')
 
+        # Update intersphinx_mapping.
+        pydoctor_url_fragment = config.pydoctor_url_fragment
+        if pydoctor_url_fragment:
+            intersphinx_mapping = config.intersphinx_mapping
+            url = pydoctor_url_fragment.format(**{'rtd_version': rtd_version})
```
adiroiban, line 101:
```suggestion
            url = pydoctor_url_path.format(**{'rtd_version': rtd_version})
```

adiroiban, line 101:
```suggestion
        pydoctor_url_fragment = config.pydoctor_url_fragment
        if pydoctor_url_fragment:
            intersphinx_mapping = config.intersphinx_mapping
            url = pydoctor_url_fragment.format(**{'rtd_version': rtd_version})
```
```suggestion
        pydoctor_url_path = config.pydoctor_url_path
        if pydoctor_url_path:
            intersphinx_mapping = config.intersphinx_mapping
            url = pydoctor_url_path.format(**{'rtd_version': rtd_version})
```

.github/workflows/static.yaml
```diff
@@ -44,8 +44,3 @@ jobs:
     - name: Run docs and check extensions
       run: |
         tox -e docs
-        # Check a known `pydoctor --help` argument.
```
adiroiban, line 47: these checks were moved to tox, to make it easier to run them on local system

docs/source/usage.rst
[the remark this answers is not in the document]
adiroiban: I feel that this is a dev tools and there is not much value in making the tool bullet-proof.
The whole extension is a hack... quite fragile..but for now it works.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -35,7 +43,37 @@ def on_build_finished(app: Sphinx, exception: Exception) -> None:
...
+    for key, value in runs.items():
+        arguments = _get_arguments(value, placeholders)
+
+        options, _ = parse_args(arguments)
+        output_path = pathlib.Path(options.htmloutput)
+        sphinx_files = pathlib.Path(options.htmloutput + '_sphinx_files')
+
+        temp_path = pathlib.Path(options.htmloutput + '_pydoctor_temp')
+        shutil.rmtree(sphinx_files, ignore_errors=True)
+        output_path.rename(sphinx_files)
```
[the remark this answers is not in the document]
adiroiban: This will not work. It just calls os.rename. I read the code and the docs and also tried it :)

pydoctor/sphinx_ext/build_apidocs.py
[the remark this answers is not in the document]
adiroiban: I have remove that info from the docstring... without adding extra logic for command line handling.

adiroiban:
Thanks @mthuurne for another round of great comments.
Please check latest changes and you can merge if you are happy :)

I have not implemented extra logic for sanitizing the command line argumets passed via the Sphinx extension.

Can we have a new release ?
I would like to update Twisted docs and also cut a release for Twisted.

pydoctor/sphinx_ext/build_apidocs.py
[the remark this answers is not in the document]
mthuurne: OK, then the docs of `Path.replace()` are wrong for saying "If _target_ points to an existing file **or directory**, it will be unconditionally replaced."

Anyway, thanks for trying.

docs/source/usage.rst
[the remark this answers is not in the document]
mthuurne: My issue is not with the code being fragile, but with the documentation suggesting that the code is more fragile than it actually is. I'd like to think that the extension will become less fragile over time. So if it can handle these arguments now, why would it be a problem to keep it that way in the future?

And if there really is no way to stay compatible, we're likely making such invasive changes that argument lists need to be revised anyway. Do you have a particular scenario in mind where this warning will pay off? Because I can't think of one.

docs/source/usage.rst
```diff
@@ -204,39 +214,44 @@ HTML files that can be used any HTTP server.
...
+    pydoctor_url_path = '/en/{rtd_version}/api/'
+
+You can pass almost any argument to `pydoctor_args`
+in the same way you call `pydoctor` from the command line.
 
-You can pass any argument, in the same way you call `pydoctor` from the
-command line.
+You should not pass the `--make-html`, `--make-intersphinx` or `--quiet`
+arguments as this can break the build.
+The extension will add them automatically.
```
adiroiban, line 242: What about?

```suggestion
You don't need to pass the `--make-html`, `--make-intersphinx` or `--quiet`
arguments.
The extension will add them automatically.
```

mthuurne: That's fine, but you'd also have to remove `--make-intersphinx` from the list then.

By the way, I checked the code and `--make-intersphinx` does nothing if `--make-html` is passed as well, so it is harmless.

mthuurne:
I'm OK with releasing, but I'm wondering if the version should be 20.12.2 or 20.13.0:
- 20.12.2 would mean we pick the minimum required changes only
- 20.13.0 would mean we just release what is in master

Already merged in master since 20.12.1 are #289, #314 and #334.
(Plus one for the GitHub Actions config that doesn't affect users.)

Releasing the removal of `--html-write-function-pages` would break Twisted, but if Twisted is updated immediately after that may not be an issue.

Do you want any other PRs in the release or just this one? It is a bit large for a bug fix, but since the extension isn't very useful without it, I guess it could qualify as one from the user's perspective. As it doesn't touch any functionality outside the extension, there is not much risk of regressions.

adiroiban:
+1 for major 20.13.0 release

Only this PR. This is a major new feature to the extension. Previous version only had support for RTD build but not for own Sphinx inventory.

As soon as this is released, I plan to update Twisted trunk to have working RTD links :)

PR #332, opened 2020-12-12
(Fixes #331) Add project version to Sphinx object inventory.

### LINKED ISSUE #331: Write project version in the Sphinx inventory object.
Right now, this is hardcoded to 2.0 ...as 7 years ago I though that this is the version of the file format.

https://github.com/twisted/pydoctor/blob/13c011dc1e984712e39032744b440c56a9b0f1fe/pydoctor/sphinx.py#L185

As a start, add a CLI argument that will alllow passing the version as a string.

adiroiban:
Fixes #331

This updated the sphinx inventory implementation to support project version.

Previously it was hardcoded to version 2.0

Changes
=======

* Add `--project-version` to specify a fixed version,
* Refactor `parse_args` as a high level function for handling the CLI options and preparing them for internal usage.

As a drive by, I have updated the tox default env list and Contributing docs, so that you can just run `tox -p all` and have all the "recommended" env executed before pushing your changes.

We tried extracting the info from project meta file, but decided to move to a separate PR

How to test
=========

`tox -e apidocs` builds the version and name from metadata.

`strings build/apidocs/objects.inv | head` to check the version in Sphinx object inventory.

`tox -e docs` to build with version and name from both metadata and custom CLI args (for epydoc sample).

On read the docs. go here https://pydoctor--332.org.readthedocs.build/en/332/usage.html#building-pydoctor-together-with-sphinx-html-build

And then go with the mouse over : This can be done by using the **pydoctor.sphinx_ext.build_apidocs** extension.

You should see the current pydoctor version.


adiroiban:
This is ready for review. Thanks! Let me know if you want me to split it.

pydoctor/driver.py
```diff
@@ -279,6 +303,48 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
...
+            """
+
+    if not options.projectversion:
+        if meta:
+            options.projectversion = meta.version
+        else:
+            options.projectversion = '0.1.0.dev0'
+
+    if not options.projectname and meta:
+        options.projectname = meta.metadata['name']
```
graingert, line 337: might be nice to load this from the metadata too:

```suggestion
        options.projectname = meta.metadata['name']
        
        if not options.projecturl and meta:
            options.projecturl = meta.metadata['Home-Page']
```

graingert:
these are some very nice tests, I'd like to see a review from @mthuurne too

pydoctor/test/test_sphinx.py
```diff
@@ -66,12 +66,20 @@ def inv_reader_nolog() -> sphinx.SphinxInventory:
 
 @pytest.fixture
 def inv_writer() -> InvWriter:
-    return InvWriter(logger=PydoctorLogger(), project_name='project_name')
+    return InvWriter(
```
adiroiban, line 69: I don't like per-test fixtures as I think that they make the code harder to read.

I like the module or package fixtures.

I would argue that they can be replaced by a simple function that is explicitly called inside each test.
In this way, you can pass extra arguments to affect the execution of the test... and you also have the function call for inspection with your IDE (if you use any) 

With the current approach, you read a test and see the assertion of 'project-name' or '1.2.0.rc1` and you will ask: from where did those  value came?

If the test don't care about the value... is ok to use the default.

```python
def get_writer(name=None, version=None, logger=None):
  if name is None:
    name = 'project_name'
  if version is None:
    version = '1.2.0.rc1'
  if  logger is None:
    logger = PydoctorLogger()
  return InvWriter(logger=logger, project_name=name, project_version=version)

# When we have multiple calls with similar arguments.
def get_writer_without_log(name=None, version=None):
  return get_writer(name=name, version=version, logger=PydoctorNoLogger())

def test_error_on_writing():
   writer = get_writer()
   # Trigger an error in which name and version don't matter.


def test_project_name_and_version_in_inventory():
    writer = get_writer(name='acme-lib', version='1.3.0')
    # have assertion based on values defined in the same tests.

```

pydoctor/driver.py
adiroiban: Thanks. I will move all the 'meta' stuff to a separate PR, to also see what other info we can extract.

pydoctor/test/test_sphinx.py
graingert: > I would argue that they can be replaced by a simple function that is explicitly called inside each test.

I tend to prefer that style

mthuurne (commented):
I agree that it's best to split this into a fairly simple PR that deals with the version stuff and a more complex one for the metadata processing.

Once we start looking at metadata, we might want to reconsider a lot of pydoctor's interface. For example `--project-base-dir` currently only affects the computation of source URLs, but gets a different role with this PR.

mthuurne:
It looks like `pep517`, when we ask for the project's metadata, will build a `dist-info` directory in a temporary location and then use `importlib.metadata` to read the metadata from there. That seems like a very indirect way of fetching the metadata. When I test it on pydoctor itself, I even see `pip` getting called twice somewhere in the process.

I'm not comfortable with launching such an elaborate processing as a side effect of getting metadata, especially if we want to fetch the metadata also when the user didn't explicitly request it, such as when `--project-base-dir` is passed. I'd much prefer to extract metadata by parsing files instead of executing a build.

adiroiban:
> I'm not comfortable with launching such an elaborate processing as a side effect of getting metadata, especially if we want to fetch the metadata also when the user didn't explicitly request it, such as when `--project-base-dir` is passed. I'd much prefer to extract metadata by parsing files instead of executing a build.

I understand your pain :)

I went with pep517 with the assumption that is standard and is convenient.
We can use pep517 only when --base-meta-dir (or renamed to --meta-dir) is passed from the command line.

I am not aware of any other library that can extract metadata just py parsing a file.
I went with pep517 in order to avoid reinventing the wheel.

You can check the first commit, in which pep517 is not used.

----

I will address your review comments and will move the meta extraction in a separate PR.

mthuurne:
> I went with pep517 with the assumption that is standard and is convenient.
> We can use pep517 only when --base-meta-dir (or renamed to --meta-dir) is passed from the command line.
> 
> I am not aware of any other library that can extract metadata just py parsing a file.
> I went with pep517 in order to avoid reinventing the wheel.

Metadata seems to be one of the last parts of Python packaging getting a proper redesign, so `pep517` might reflect the current state of things. But it doesn't fit in pydoctor's approach: we put a lot of effort into static analysis of source code, so it would be strange to adopt execution for extracting metadata.

@graingert pointed me to [PEP 621](https://www.python.org/dev/peps/pep-0621/) on IRC, which puts metadata in a tool-neutral place in `pyproject.toml`. This is a relatively new document, but is there anything stopping us from adopting it right now?

For broader compatibility, we could fall back on tool-specific sections of `pyproject.toml`, such as `[tool.poetry]`.

For Twisted and other projects using Incremental or otherwise storing the version in the source code, we could get the project name and version from `__version__` in the top-level `__init__.py`.

@graingert also mentioned the option of querying an installed package using `importlib_metadata`. When running pydoctor from `tox`, the package we're documenting is often installed already, and if not it's easy to request that. So that could be a way to get access to the same information `pep517` provides without starting a build from pydoctor. There is a risk though that the installed version is not the version we're documenting.

mthuurne:
As discussed on IRC, the plan is to add `--project-version` now and add metadata processing later.

The default project version will be empty string, which makes `intersphinx` produce a link without a version in it. Note that the `Version:` line must be present or the inventory will be rejected. But the value can be empty.

pydoctor/sphinx.py
```diff
@@ -182,10 +182,9 @@ class SphinxInventoryWriter:
     Sphinx inventory handler.
     """
 
-    version = (2, 0)
-
-    def __init__(self, logger: Callable[..., None], project_name: str):
+    def __init__(self, logger: Callable[..., None], project_name: str, project_version: str):
         self.project_name = project_name
+        self._project_version = project_version
```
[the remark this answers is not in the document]
adiroiban: well.. I make new things private by default. Nobody should query the project version via `SphinxInventoryWriter`

I would say that project_name should also be private.

 so for consistency. I will with a private project name.

pydoctor/test/test_commandline.py
```diff
@@ -116,6 +116,48 @@ def test_cli_warnings_on_error() -> None:
...
+
+    assert options.projectversion == '3.4.0.dev0'
+
+
+def test_project_version_string() -> None:
+    """
+    --project-version can be passed as a simple string.
+    """
+    options, args = driver.parse_args(['--project-version', '1.2.3.rc1'])
+    assert options.projectversion == '1.2.3.rc1'
```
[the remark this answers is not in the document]
adiroiban: should e moved.

pydoctor/test/test_commandline.py
```diff
@@ -130,6 +172,23 @@ def test_main_project_name_guess(capsys: CapSys) -> None:
...
+    setup_cfg = tmp_path / 'setup.cfg'
+
+    setup_cfg.write_text("""
+[metadata]
+name = acme-lib
+""")
+
+    options, args = driver.parse_args(['--project-meta-dir', str(tmp_path)])
+
+    assert options.projectname == 'acme-lib'
```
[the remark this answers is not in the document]
adiroiban: code no longer used

pydoctor/driver.py
```diff
@@ -279,6 +303,48 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
...
+
+
+def main(args: Sequence[str] = sys.argv[1:]) -> int:
```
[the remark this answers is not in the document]
adiroiban: we need to accept a default value as this is the main script. I left it as it is. Why you want remove the sys.argv[1:] default value?

pydoctor/driver.py
```diff
@@ -279,6 +303,48 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
...
+            meta = pep517_meta_load(options.projectmetadirectory)
+        except (RuntimeError, FileNotFoundError, NotADirectoryError):
+            """
+            PEP517 raises  when metadata file is not found:
+            * py3.8 - RuntimeError
+            * py3.6 - FileNotFoundError Unix
+            * py3.6 - NotADirectoryError Windows
+
+            It's ok if project has no meta data.
+            """
```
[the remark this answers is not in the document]
adiroiban: code no longer used

pydoctor/driver.py
adiroiban: meta code removed from this PR.

adiroiban:
This is ready for a re-review. See PR description for how to check.

pydoctor/driver.py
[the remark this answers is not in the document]
mthuurne: `pydoctor/__main__.py` calls this using:
```py
sys.exit(main(sys.argv[1:]))
```
So while we certainly need `sys.argv[1:]`, we don't need it as a default.

pydoctor/test/test_commandline.py
```diff
@@ -205,3 +221,26 @@ def test_main_source_outside_basedir(capsys: CapSys) -> None:
             'pydoctor/test/testpackages/basic/'
             ])
     assert "Source path lies outside base directory:" in capsys.readouterr().err
+
+
+def test_make_intersphix(tmp_path: Path) -> None:
+    """
+    --make-interphinx will only produce the Sphinx inventory object.
```
mthuurne, line 228: This happens only when `--make-html` is not passed. I don't think we need a separate test for that, but it might be good to mention here in the docstring.

mthuurne, line 228:
```suggestion
    --make-intersphinx will only produce the Sphinx inventory object.
```

pydoctor/test/test_commandline.py
```diff
@@ -205,3 +221,26 @@ def test_main_source_outside_basedir(capsys: CapSys) -> None:
...
+
+    This is also an integration test for Sphinx inventory writer.
+    """
+    inventory = tmp_path / 'objects.inv'
+    exit_code = driver.main(args=[
+        '--project-base-dir=.',
+        '--make-intersphinx',
+        '--project-name=acme-lib',
+        '--project-version=1.2.0',
```
mthuurne, line 237: Since the system test already checks using "1.2.0", maybe use something more complex here, like "20.12.0-dev123". Of course that doesn't actually make a difference with the current implementation, since we just pass the string around, but if we ever switch back to parsing versions, it could be useful.

tox.ini
```diff
@@ -125,7 +125,6 @@ description = Build only the API documentation
 
 commands =
     pydoctor \
-    --project-name=pydoctor \
```
mthuurne, line 128: Should this option be removed now that the metadata change has been split off?

pydoctor/test/test_sphinx.py
```diff
@@ -66,12 +66,20 @@ def inv_reader_nolog() -> sphinx.SphinxInventory:
...
+        logger=PydoctorLogger(),
+        project_name='project_name',
+        project_version='1.2.0.rc1',
```
mthuurne, line 72: "1.2.0.rc1" is almost, but not quite, PEP 440 compliant (due to the dot before "rc"). I think it would be best to use a fully compliant version, but if you really want to test that the version is not processed in any way, it should be more explicit in its non-compliance.

docs/test.py
```diff
@@ -40,7 +41,31 @@ def test_rtd_pydoctor_multiple_call():
...
+    with open(BASE_DIR / 'objects.inv', 'rb') as stream:
+        data = stream.read()
+        assert data.startswith(
+            b'# Sphinx inventory version 2\n'
+            b'# Project: pydoctor\n'
+            b'# Version: ' + __version__.encode() + b'\n'
+            ), data
+
+    # The demo/showcase inventor for which we have a fixed version.
+    with open(BASE_DIR / 'docformat' / 'epytext' / 'demo' / 'objects.inv', 'rb') as stream:
```
mthuurne: This could be put in a separate test case, so we'll have more information when the first checks fails.

pydoctor/driver.py
```diff
@@ -92,7 +92,16 @@ def getparser() -> OptionParser:
         help=("A dotted name of the class to use to make a system."))
     parser.add_option(
         '--project-name', dest='projectname',
-        help=("The project name, appears in the html."))
+        help=("The project name, shown at the top of each HTML page."))
+    parser.add_option(
+        '--project-version',
+        dest='projectversion',
+        default='',
+        metavar='VER',
```
mthuurne, line 100:
```suggestion
        metavar='VERSION',
```
Our other metavars aren't very short either, so I think we can afford a few extra characters here.

pydoctor/driver.py
```diff
@@ -92,7 +92,16 @@ def getparser() -> OptionParser:
...
+        help=(
+            "The version of the project for which the API is generated. "
```
mthuurne, line 102:
```suggestion
            "The version of the project for which the API docs are generated. "
```

mthuurne (approved):
Looks good, some details still need refinement but nothing shocking.

Do you still want to replace the per-test fixture with a test helper to make the checked values more explicit? That was a remark from the first review that hasn't been resolved yet.

tox.ini
adiroiban: True! Thanks.

pydoctor/driver.py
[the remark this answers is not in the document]
adiroiban: @mthuurne 

This is not about pydoctor/__main__.py 
It's about the bin/pydoctor or (Scritps/pydoctor.bat on Win) call which is

```
#!/home/dev/pydoctor/.tox/test/bin/python
# -*- coding: utf-8 -*-
import re
import sys
from pydoctor.driver import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

adiroiban:
> Do you still want to replace the per-test fixture with a test helper to make the checked values more explicit? That was a remark from the first review that hasn't been resolved yet.

Thanks. I forgot about that. I did a refactoring...but a bit different in order not to touch the private logger in the assertion stage.

pydoctor/test/test_sphinx.py
adiroiban: Please check the current implementation... it's no longer a fixture, but is implemented a bit different so that you don't have to touch the private logger attribute.

adiroiban:
@mthuurne please do a quick re-review and merge if you are happy.

I will re-run the static check jobs a few time to make sure we no longer have flaky failures.
I will investigate more in a separate PR. Issue created.

Thanks!

adiroiban:
I have manually re-run the static  code checks more than 3 times in a row and they are green. I will try to see what is going on there in #339

docs/test.py
adiroiban: Was moved to separate test.

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -93,9 +93,10 @@ def on_config_inited(app: Sphinx, config: Config) -> None:
         output_path = pathlib.Path(options.htmloutput)
         temp_path = output_path.with_suffix('.pydoctor_temp')
 
-        # Update intersphinx_mapping.
         pydoctor_url_path = config.pydoctor_url_path
-        if pydoctor_url_path:
+        if pydoctor_url_path and key == "main":
+            # Update intersphinx_mapping only for the main API doc.
+            # FIXME:https://github.com/twisted/pydoctor/issues/339
```
mthuurne, line 99: While it's good to have a workaround, is there any reason to not fix this properly now?

If I understand it correctly, the root of the problem is that we have two projects (`main` and `epydoc_demo`) for which API docs are generated, but only one `pydoctor_url_path`. While only updating the mapping for `main` fixes the broken links in our documentation, it means we aren't able to link to anything inside the epydoc demo.

A possible solution would be to treat `pydoctor_url_path` the same as `pydoctor_args` and support both a single value and a dictionary. That shouldn't require a lot of code to implement (see line 85 above).

pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -155,6 +156,6 @@ def setup(app: Sphinx) ->  Mapping[str, Any]:
 
     return {
         'version': __version__,
-        'parallel_read_safe': True,
-        'parallel_write_safe': True,
+        'parallel_read_safe': False,
+        'parallel_write_safe': False,
```
mthuurne, line 160: Is this change still needed?

mthuurne (approved):
The actual feature implementation is good now. But the PR still has some leftovers from diagnosing the flaky test results (#339) that should be cleaned up before merging.

Do you think it's better to fix #339 in this PR or in a separate PR? A separate PR would be cleaner, but I don't want to create a lot of extra work.

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: I tried to reduce the scope of this PR  and at the same time, not to block this PR while #339 is not yet fixed.
So I prefer to merge it as it is :)

docs/test.py
```diff
@@ -40,7 +43,44 @@ def test_rtd_pydoctor_multiple_call():
 
 def test_rtd_extension_inventory():
     """
-    The Sphinx inventory is available during normalsphinx-build.
+    The Sphinx inventory is available during normals sphinx-build.
+    """
+    with open(BASE_DIR / 'usage.html', 'r') as stream:
+        data = stream.read()
+        assert 'href="/en/latest/api/pydoctor.sphinx_ext.build_apidocs.html"' in data, data
```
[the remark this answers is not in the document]
adiroiban: if you run `pytest -s` with your approach from  https://github.com/twisted/pydoctor/commit/fbf280adcd2b499b9f4101627206e8f11885986f you end up with an ugly output.

I think that having the full page listed only on error is the "right thing" (tm) :) and this is how the testing frameworks are designed to work.

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: is not needed yet ... I will revert.  Will be needed when we will introduce the `{source_reference}` placeholder. But I can add it there. Thanks!

docs/test.py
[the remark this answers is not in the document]
mthuurne: > if you run `pytest -s` with your approach from [fbf280a](https://github.com/twisted/pydoctor/commit/fbf280adcd2b499b9f4101627206e8f11885986f) you end up with an ugly output.

Whether it's ugly or useful depends on what you're doing: there could be a reason you want to see the raw data also when a test does not fail, for example if you suspect the test is not strict enough.

But these are details; it's fine to leave this as it is now and I'll revert fbf280a then before merging master into #338.

.github/workflows/static.yaml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   static_checks:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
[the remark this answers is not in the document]
adiroiban: This is not needed. I will revert it. Note that we might want to use a pinned release, so that we decide when to upgrade and don't let Github to decide for ourselves :)

pydoctor/sphinx_ext/build_apidocs.py
adiroiban: Issue fixed :) Thanks for your push :p

adiroiban:
@mthuurne thanks for the review. I am going to merge this.
I have updated the PR description as it was updated.

Issue #315, opened 2020-12-04
Integrate Twisted's customizations


mthuurne:
Twisted has a [source file](https://github.com/twisted/twisted/blob/trunk/src/twisted/python/_pydoctor.py) in which it customizes some aspects of pydoctor:

- workaround for doc URL lookups from `zope.interface`
- formatting of deprecation notices
- make test packages invisible, except `twisted.test.proto_helpers`

I'm not sure if the workaround is still necessary, but if it is, it would be useful for all projects that want to link to `zope.interface` documentation, which means it would be better to move this code into pydoctor.

The decorators used to mark deprecations in Twisted are public API, so a project that uses Twisted could use those as well. In that case, it would be an advantage if the functionality for formatting them was inside pydoctor. In the future, we might want to support other deprecation mechanisms such as [regret](https://pypi.org/project/regret/), which would be easier to do in a consistent fashion if all the code is inside pydoctor.

Having the ability to exclude certain packages from the documentation sounds like a useful feature, so we could consider to facilitate that through command line options instead of customization code.

Put together, I think all of the code customizations that Twisted currently does would actually fit better inside pydoctor itself. Integrating them would have multiple advantages:
- for users: it makes pydoctor more capable out of the box
- for pydoctor developers: we can change the design of pydoctor without worrying about breaking Twisted's docs
- for Twisted developers: better chances of this code being kept up-to-date

The template customizations would remain in Twisted. We are working on breaking up the templates into smaller ones, to make customizations more fine-grained, see #299 for details.

mthuurne:
Note that the code for deprecation notices is currently spread over both Twisted and pydoctor: Twisted extracts them, but pydoctor inserts them into the output and filters the deprecation decorators out of the output.

tristanlatr:
> Having the ability to exclude certain packages from the documentation sounds like a useful feature, so we could consider to facilitate that through command line options instead of customization code.

May be `:public:`, `:private:` epydoc/rst tags ?

mthuurne:
> > Having the ability to exclude certain packages from the documentation sounds like a useful feature, so we could consider to facilitate that through command line options instead of customization code.
> 
> May be `:public:`, `:private:` epydoc/rst tags ?

Good idea. Especially for complex cases, where you first want to exclude something (`twisted.test`) and then re-include a subset (`twisted.test.proto_helpers`), tagging the packages might be more elegant than command line options.

And we already had a tag idea for marking private `__init__` methods.

[...]

PR #334, opened 2020-12-15
Define version in metadata and extract it using importlib.metadata

mthuurne:
Until now, pydoctor used Incremental to define the version in code and extract it from there during the build to insert it into the
metadata. This commit reverses that approach, by defining the version in metadata and having the code extract it from there.

Static metadata is easier to use by tools such as pydoctor itself. In the future we'll probably want to adopt PEP 621 for that purpose; this is a stepping stone.


adiroiban (approved):
Looks good. Happy to see that "incremental" is removed :)

pydoctor/__init__.py
```diff
@@ -4,10 +4,17 @@
...
+# On older versions, a compatibility package can be installed from PyPI.
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    if not TYPE_CHECKING:
+        import importlib_metadata
 
-from ._version import __version__
 
+__version__ = importlib_metadata.version('pydoctor')
```
graingert, line 18: changing this from an Incremental instance to a str is a breaking change - I think you might as well forgo the `__version__ =` entirely and require callers `importlib.metadata.version("pydoctor")` if they ever forget what version they installed

mthuurne: That's a fair point, but pydoctor doesn't offer a stable API yet, so there shouldn't be many callers. Even Twisted's customizations could be phased out; see #315.

mthuurne:
Note that we do have to update Twisted's unit tests to not access `__version__`, or at least not expect there to be an Incremental `Version` object there. But that doesn't need to happen until we do a new major release, which might have other breaking changes as well.

PR #335, opened 2020-12-15
Add type annotations to code imported from epydoc

mthuurne:
Some parts of the epydoc code already had annotations, but now all of it has.


pydoctor/epydoc/markup/epytext.py
```diff
@@ -1230,35 +1240,35 @@ def descr(self):
...
     @param errors: A list where any errors generated during parsing
         will be stored.
-    @type errors: C{list} of L{ParseError}
-    @rtype: L{ParsedDocstring}
     """
     tree = parse(docstring, errors)
     if tree is None:
         return ParsedEpytextDocstring(None, ())
 
+    tree_children = cast(List[Element], tree.children)
```
adiroiban, line 1256: are all these casts needed just because `parse` can return None ?

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1230,35 +1240,35 @@ def descr(self):
...
-        field_list = tree.children.pop()
+        field_list = tree_children.pop()
+        field_children = cast(List[Element], field_list.children)
 
-        for field in field_list.children:
+        for field in field_children:
             # Get the tag
-            tag = field.children[0].children[0].lower()
-            del field.children[0]
+            tag = cast(str, cast(Element, field.children.pop(0)).children[0]).lower()
```
adiroiban, line 1266: I guess that `tree_children` cast is needed because `parse` may return None .
But we we need the cast here and in other places. Shouldn't Element.children already have List[Elemenet] anotation?

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -155,28 +157,32 @@ class _EpydocReader(StandaloneReader):
...
 
-    def report(self, error):
-        try: is_fatal = int(error['level']) > 2
-        except: is_fatal = True
-        try: linenum = int(error['line'])
-        except: linenum = None
+    def report(self, error: docutils.nodes.system_message) -> None:
+        try:
+            is_fatal = int(error['level']) > 2
+        except:
```
adiroiban, line 180: Maybe 
```suggestion
        except Exception:
```

adiroiban (approved):
All good. Thanks! Only minor comments.
This can be merged.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -249,20 +245,23 @@ def __repr__(self):
 ## Structuring (Top Level)
 ##################################################
 
-def parse(text, errors = None):
+@overload
+def parse(text: str) -> Element: ...
+
+@overload
+def parse(text: str, errors: List[ParseError]) -> Optional[Element]: ...
```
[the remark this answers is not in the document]
mthuurne: It is at least a sign that the API isn't super simple. Sometimes there might be good reasons for that though.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -382,23 +387,33 @@ def _pop_completed_blocks(token, stack, indent_stack):
             stack.pop()
             indent_stack.pop()
 
-def _add_para(doc, para_token, stack, indent_stack, errors):
+def _add_para(
+        para_token: 'Token',
+        stack: List[Element],
+        indent_stack: List[Optional[int]],
```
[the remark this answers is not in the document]
mthuurne: > I guess that here None is used as `end of game` marker.
> In my code, I do something like `end_marker = object()` and then use end_marker instead of None.
> In this way, I find the code easier to read.

In this case, `None` is used for "indentation unknown", while -1 is used as a sentinel:
```py
    # Maintain two parallel stacks: one contains DOM elements, and
    # gives the ancestors of the current block.  The other contains
    # indentation values, and gives the indentation of the
    # corresponding DOM elements.  An indentation of "None" reflects
    # an unknown indentation.  However, the indentation must be
    # greater than, or greater than or equal to, the indentation of
    # the prior element (depending on what type of DOM element it
    # corresponds to).  No 2 consecutive indent_stack values will be
    # ever be "None."  Use initial dummy elements in the stack, so we
    # don't have to worry about bounds checking.
    stack = [cast(Element, None), doc]
    indent_stack = [-1, None]
```

> I guess that with typing, this adds extra complexity as you will first need to define a type.
> Or can you have `end_marker: EndMarkerT = object()` ?

Type names have to be defined as Python objects; you can't just introduce a name without defining it first.

Maybe [NewType](https://docs.python.org/3/library/typing.html?highlight=typing#typing.NewType) could be used for typed sentinels? I have never tried that though.

pydoctor/epydoc/markup/epytext.py
mthuurne: No, `tree.children` has type `List[Union[str, Element]]`, but the root element will never be given children of type `str`.

Maybe at some point we should create separate node subclasses for different kinds of elements, but for now I'm happy just annotating epydoc's original design without changing it much.

pydoctor/epydoc/markup/epytext.py
mthuurne: No, some elements have `str` children as well.

The design of this code isn't very friendly to type annotations: the `Element` class is very generic, but that means that the code using it contains a lot of assumptions about what kind of children different elements can contain. And those assumptions are coded in a way that mypy needs a lot of help to understand.

For now, I prefer to annotate the current design. But in the future, we could consider a few different directions for this code:
- instead of building an intermediate representation in a tree, route parse events to the code that generates Stan objects
- have separate `Element` subclasses for different elements, that put restrictions on the number and types of child nodes
- build an intermediate representation using `docutils` objects and let `docutils` convert that to HTML

pydoctor/epydoc/markup/restructuredtext.py
mthuurne: Indeed we shouldn't be catching everything.

After reading a bit of the `docutils` code, I don't think `level` could be undefined or something other than an integer, so I'll just remove this overly cautious piece of code.

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
adiroiban: Thanks for the explanation.

Thanks for the link. I think it will work and makes sense :)

```python
Python 3.8.5 (default, Jul 28 2020, 12:59:40) 
[GCC 9.3.0] on linux
>>> EndMarkerT = NewType('EndMarkerT', object)
>>> end_marker = EndMarkerT(object())
```

-------

What? NewType introduced in 3.5.**2** ?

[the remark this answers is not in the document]
mthuurne: > What? NewType introduced in 3.5.**2** ?

There are a few more like that, for example `NoReturn`, which was new in 3.5.4 and 3.6.2. I guess that with type checking being a somewhat experimental feature at that time, they favored quick releasing over sticking to a "bug fixes only" policy.

PR #336, opened 2020-12-17
Fixes, cleanups and type annotations for zope.interface support

mthuurne:
After I added annotations, mypy flagged several cases where pydoctor made invalid assumptions about types. These could lead to crashes when analyzing invalid code. Since projects don't tend to keep invalid code around, this is unlikely to occur often in practice, but of course it shouldn't happen at all. In one case, the method (`allImplementations` property) was broken but it was never called, so it couldn't lead to problems.

### DIFF
--- a/pydoctor/test/test_astbuilder.py
+++ b/pydoctor/test/test_astbuilder.py
@@ -500,8 +500,7 @@ def test_classdecorator(systemcls: Type[model.System]) -> None:
         pass
     ''', modname='mod', systemcls=systemcls)
     C = mod.contents['C']
-    assert C.decorators == [(('cd', 'mod.cd', mod.contents['cd']), None)], \
-      C.decorators
+    assert C.decorators == [('mod.cd', None)]
 
 
 @systemcls_param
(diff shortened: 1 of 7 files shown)


mthuurne:
CI seems to be broken because of actions/setup-python#171.

pydoctor/test/test_astbuilder.py
```diff
@@ -500,8 +500,7 @@ class C:
         pass
     ''', modname='mod', systemcls=systemcls)
     C = mod.contents['C']
-    assert C.decorators == [(('cd', 'mod.cd', mod.contents['cd']), None)], \
-      C.decorators
+    assert C.decorators == [('mod.cd', None)]
```
adiroiban, line 503: It looks like this case was also using the custom assert failure message.

It's ok to remove it, and we can added if if needed... not sure if it's still needed.

```suggestion
    assert C.decorators == [('mod.cd', None)], C.decorators
```

adiroiban (approved):
Wonderful work. Thanks!
This can be merged.

pydoctor/test/test_astbuilder.py
mthuurne: I think that in general, pytest's reporting on asserts is good enough. Only in special cases, like where we had to ignore the length limit, do we need a custom message.

PR #337, opened 2020-12-18
Use new syntax to select pypy version

mthuurne:
As suggested [here](https://github.com/actions/setup-python/issues/171#issuecomment-747917974), this should work around the regression in the `setup-python` task, as well as being more explicit about which version of pypy we want to run on.


adiroiban (approved):
Thanks!

PR #338, opened 2020-12-24
Prepare for 21.2.0 release

mthuurne:


README.rst
```diff
@@ -60,96 +60,99 @@ You can select a different format using the ``--docformat`` option.
 What's New?
 ~~~~~~~~~~~
 
-pydoctor dev
-^^^^^^^^^^^^
+pydoctor 20.13.0
+^^^^^^^^^^^^^^^^
 
 * Removed the ``--html-write-function-pages`` option. As a replacement, you can use the generated Intersphinx inventory (objects.inv) for deep-linking your documentation.
+* The ``build_apidocs`` Sphinx extension builds the API docs early, such that the rest of the documentation can link to it via Intersphinx.
```
adiroiban, line 67:
```suggestion
* The ``build_apidocs`` Sphinx extension builds the API docs before Sphinx docs, such that the rest of the documentation can link to it via Intersphinx.
* The ``build_apidocs`` Sphinx extension now has the `pydoctor_url_path` that will automatically update the `intersphinx_mapping` variable so that it uses the latest API inventory.
```

adiroiban (approved):
Looks good. Thanks.
Only minor comments.

Writing documentation is hard :)

mthuurne:
When the test failes, the generated HTML looks like this:
```html
<p>This can be done by using the <code class="xref py py-mod docutils literal notranslate"><span class="pre">pydoctor.sphinx_ext.build_apidocs</span></code> extension.</p>
```
So it seems that the test is correct in failing, since the output that was produced does not contain the link.

adiroiban (approved):
I would like to request the inclusion of #328 and I guess that this will be 21.1.0 

Thanks!

mthuurne:
While I would have liked to have #299 and #328 included as well, there is more than enough merged already to make a release worthwhile. Let's try to do another release in March with those PRs included.

PR #340, opened 2020-12-29
Use a mapping for pydoctor_url_path for multi project builds.

adiroiban:
This fixed the bug in which Sphinx will overwrite 2 inventory files if they have the same URL.

Just used a mapping just like we have for the pydoctor_args.

We don't have tests for this.

Pydoctor own system / docs tests are using the mapping code, but we don't have tests for the non-mapping configuration.

I plan to use the non-mapping configuration in Twisted.

### DIFF
--- a/pydoctor/sphinx_ext/build_apidocs.py
+++ b/pydoctor/sphinx_ext/build_apidocs.py
@@ -9,6 +9,8 @@ Inside the Sphinx conf.py file you need to define the following configuration op
   - C{pydoctor_url_path} - defined the URL path to the API documentation
                            You can use C{{rtd_version}} to have the URL automatically updated
                            based on Read The Docs build.
+                        - (private usage) a mapping with values URL path definition.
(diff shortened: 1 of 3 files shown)


pydoctor/sphinx_ext/build_apidocs.py
```diff
@@ -9,6 +9,8 @@
   - C{pydoctor_url_path} - defined the URL path to the API documentation
                            You can use C{{rtd_version}} to have the URL automatically updated
                            based on Read The Docs build.
+                        - (private usage) a mapping with values URL path definition.
+                          Make sure each definition will produce a unique URL.
```
adiroiban, line 13:
```suggestion
                         - (private usage) a mapping with values URL path definition.
                           Make sure each definition will produce a unique URL.
```

adiroiban:
This should be ready for a review. Thanks!

mthuurne (approved):
Looks good. Thanks!

PR #344, opened 2021-01-14
Fix recognition of '@abc.abstractproperty'

mthuurne:
Property decorators containing one or more dots were not recognized by the original implementation of custom properties support. This PR fixes that. It also adds [documentation of the supported properties syntax](https://pydoctor--344.org.readthedocs.build/en/344/codedoc.html#properties).

### DIFF
--- a/docs/source/codedoc.rst
+++ b/docs/source/codedoc.rst
@@ -110,7 +110,7 @@ Further reading:
 Properties
 ----------
 
-A method with a decoration ending in ``property`` will be included in the generated API documentation as an attribute rather than a method::
+A method with a decoration ending in ``property`` or ``Property`` will be included in the generated API documentation as an attribute rather than a method::
 
     class Knight:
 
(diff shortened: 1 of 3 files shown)


docs/source/codedoc.rst
```diff
@@ -106,6 +106,29 @@ Further reading:
 - `Python Standard Library: typing -- Support for type hints <https://docs.python.org/3/library/typing.html>`_
 - `PEP 483 -- The Theory of Type Hints <https://www.python.org/dev/peps/pep-0483/>`_
 
+
+Properties
+----------
+
+A method with a decoration ending in ``property`` will be included in the generated API documentation as an attribute rather than a method::
```
adiroiban, line 113: It is clear from the examples, buy maybe we can be more catholic  than the Pope...  you can ignore it or do it as a drive by for another PR.

```suggestion
A method with a decoration ending in ``property`` (case-insensitive) will be included in the generated API documentation as an attribute rather than a method::
```

adiroiban (approved):
Great work.All good. Thanks!

docs/source/codedoc.rst
mthuurne: The code doesn't do a full case-insensitive compare; it only accepts `property` and `Property`; not `proPERty` for example.

I tried to keep the docs simple, but maybe I oversimplified here. I'll add `Property` to the docs.

adiroiban: The docs update is excellent. Thanks!

PR #345, opened 2021-01-15
Support for RST admonitions

tristanlatr:
Beautiful right ?

Also Add a demo site: https://pydoctor--345.org.readthedocs.build/en/345/docformat/restructuredtext/restructuredtext_demo.html

Capture d’écran, le 2021-01-17 à 13 21 26


pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -478,7 +477,89 @@ def visit_doctest_block(self, node: Node) -> None:
...
+    def depart_important(self, node: Node) -> None:
+        self.depart_admonition(node)
+
+    def visit_tip(self, node: Node) -> None:
+        self.visit_admonition(node, 'tip')
+
+    def depart_tip(self, node: Node) -> None:
+        self.depart_admonition(node)
+
+    # To be integrated later
```
adiroiban, line 548: I think it's better to remove these code...and add it later :)

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -498,3 +579,9 @@ def run(self) -> List[Node]:
         return [ node ]
 
 directives.register_directive('python', PythonCodeDirective)
+
+# https://docutils.sourceforge.io/docs/user/tools.html
+
+# https://docutils.sourceforge.io/docs/dev/hacking.html
+
+# https://docutils.sourceforge.io/docs/howto/rst-directives.html
```
adiroiban, line 587: I don't understand these comments.

docs/source/docformat/restructuredtext/index.rst
```diff
@@ -96,7 +96,7 @@ Here is a list of the supported ReST directives by package of origin:
 
 - `docutils`: ``.. include::``, ``.. contents::``, ``.. image::``, ``.. figure::``, ``.. unicode::``, ``.. raw::``, ``.. math::``, ``.. role::``, ``.. table::``, etc.
 - `epydoc`: None
-- `Sphinx`: None
+- `Sphinx`: ``.. warning::``, ``.. note::`` and other admonitions.
```
adiroiban, line 99: I thought that `admonitions` are standard docutils features and not something specific to Sphinx

https://docutils.sourceforge.io/docs/ref/rst/directives.html#specific-admonitions

adiroiban (commented):
Looks good.
I think that this needs a bit of cleanup.

Maybe also document all the supported standard admonition types.

It would be nice to have a RST pydoctor demo API similar to https://pydoctor--345.org.readthedocs.build/en/345/docformat/epytext/demo/epytext_demo.html

so that it would be easy to see this in action.

Thanks!

docs/source/docformat/restructuredtext/index.rst
tristanlatr: Yeah, but .. note:: etc are actually defined in sphinx only. 
But the concept of the admonition is from docutils, so I guess you're right.

pydoctor/epydoc/markup/restructuredtext.py
tristanlatr: Just links explaining how docutils directives works

mthuurne:
> Beautiful right ?
> 
> Capture d’écran, le 2021-01-15 à 11 51 27

Looks very nice.

Is there a meaningful difference between "tip" and "hint" or between "warning" and "danger"? If not, maybe the same colors can be used. That would reduce the number of color combinations we have to take into account when we use colors for other things.

Is it possible to left-align the body text with the label? ("Note:" etc.)

tristanlatr:
> Is it possible to left-align the body text with the label? ("Note:" etc.)

For sure I guess It's only a matter of CSS, but I'm not sure to understand what you mean... Like fields ? With a table ? 

For the colors it's entirely up to us to decide. (I based my CSS on this one https://pythonhosted.org/sphinxjp.themes.basicstrap/sample.html). 
We could decide the following :
- note: gray
- warning, danger, caution, error: yellow
- tip, hint, important: blue

tristanlatr:
I think we're good for now.

adiroiban:
Excellent work. 

Happy to see the RST demo. We might talk forever about minor improvements in the demo, but I think that it's a great start and can be merged ...and later have separate PR to fix minor issues.

For reference, maybe add it to PR description

https://pydoctor--345.org.readthedocs.build/en/345/docformat/restructuredtext/restructuredtext_demo.html

I think that this is a great start. I would prefer to see the demo in a separate PR so that we can merge it without waiting for admonition support.

It looks like there is a rendering error in the purple admonition... looks like raw HTML is not supported by pydoctor.

Thanks for creating the demo!

tristanlatr:
> It looks like there is a rendering error in the purple admonition... looks like raw HTML is not supported by pydoctor.

Everything looks good I don't know what you'r talking about. Here is the source code of the purple admonition patr:

```
.. raw:: html
    <style>
        .rst-admonition-purple {
            background-color: plum ! important;
        }
        .rst-admonition-purple p.rst-admonition-title{
            color: purple ! important;
        }
    </style>
.. admonition:: Purple
   This needs additionnal CSS for the new "rst-admonition-purple" class. 
   Include additional CSS by customizing the ``apidocs.css`` temlate file or by defining a raw block::
    .. raw:: html
    <style>
        .rst-admonition-purple {
            background-color: plum ! important;
        }
        .rst-admonition-purple p.rst-admonition-title{
            color: purple ! important;
        }
    </style>
   .. note:: The ``! important`` is required to overrride ``apidocs.css``. 
```

adiroiban:
True. My bad. Thanks!

mthuurne:
What is the status now? Is it ready to merge? Does @adiroiban want to do another review pass? Should I review it?

adiroiban:
This is huge, but it looks good and it's a start.

It would be nice to merge with master and update the demo to include the inherited class... but this is not specific to RST.
So it's not critical...but it would be nice to keep the features in the demo in sync.

I don't know why we have different format for the project link / first line

https://pydoctor--345.org.readthedocs.build/en/345/api/index.html

https://pydoctor--345.org.readthedocs.build/en/345/docformat/restructuredtext/restructuredtext_demo.html

I didn't had time to check the details of the RST documentation, but this looks like a big improvement and it can be merged.

Thanks!

mthuurne:
> I don't know why we have different format for the project link / first line
> 
> https://pydoctor--345.org.readthedocs.build/en/345/api/index.html
> 
> https://pydoctor--345.org.readthedocs.build/en/345/docformat/restructuredtext/restructuredtext_demo.html

There is something weird about the way the `project` template slot is filled:
- sometimes it is not filled, resulting in a larger font but a link going to the API landing page instead of the URL passed via `--project-url`
- sometimes it is filled, but this creates a second `<a>` node which then lacks the `navbar-brand` class that makes the font larger

So neither way of filling the slot is correct. I don't know if this is a bug in Twisted's template system or we're using it incorrectly, resulting in undefined behavior.

tristanlatr:
I already solved the project home link issue here: https://github.com/twisted/pydoctor/pull/299

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -478,7 +477,80 @@ def visit_doctest_block(self, node: Node) -> None:
...
+    # Other ressources on how to extend docutils: 
+    # https://docutils.sourceforge.io/docs/user/tools.html
+    # https://docutils.sourceforge.io/docs/dev/hacking.html
+    # https://docutils.sourceforge.io/docs/howto/rst-directives.html
+    # docutils apidocs: 
+    # http://code.nabla.net/doc/docutils/api/docutils.html#package-structure
+
+    # this part of the HTMLTranslator is based on sphinx's HTMLTranslator: 
+    # https://github.com/sphinx-doc/sphinx/blob/3.x/sphinx/writers/html.py#L271
+    def _visit_admonition(self, node: Node, name: str = '') -> None:
```
mthuurne, line 491: While this is inspired by Sphinx's HTMLTranslator, we don't actually need to be compatible with that, so the default value for the `name` argument could be dropped. Then we also wouldn't need the `if` statement in the body, since nothing passes empty string for `name`.

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -60,23 +60,22 @@ def prettify(html: str) -> str:
 
 # TESTS FOR NOT IMPLEMENTTED FEATURES
```
mthuurne, line 61: Since `test_rst_directive_abnomitions()` now tests an implemented feature, it should be moved above this marker.

Alternatively, the comment could be moved to the top of the file and changed to say that "xfail" tests are for non-implemented features.

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -60,23 +60,22 @@ def prettify(html: str) -> str:
...
 
-@pytest.mark.xfail
 def test_rst_directive_abnomitions() -> None:
     html = rst2html(".. warning:: Hey")
```
mthuurne, line 64: Would it be easy to parametrize (as in `pytest.mark.parametrize`) this test so that it runs for all admonitions? I don't think it's worth adding separate test cases just to boost the coverage, but adding a parameter to an existing test case could be worth it.

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -60,23 +60,22 @@ def prettify(html: str) -> str:
 
 # TESTS FOR NOT IMPLEMENTTED FEATURES
 
-@pytest.mark.xfail
 def test_rst_directive_abnomitions() -> None:
```
mthuurne, line 63:
```suggestion
def test_rst_directive_admonitions() -> None:
```

docs/source/docformat/restructuredtext.rst
```diff
@@ -48,27 +48,27 @@ Note that *docutils* requires a space before and after the ``:`` used to mark cl
...
+        '''
         :Parameters:
             size
                 The size of the fox (in meters)
             weight : float
                 The weight of the fox (in stones)
             age : int
                 The age of the fox (in years)
-        """
+        '''
```
mthuurne, line 59: PEP 257 recommends using `"""`, so I'd prefer to use that in examples as well. Was there a particular reason for this change? I couldn't find any in the commit comment of 9c2898e8.

mthuurne (commented):
I did a quick review of the code anyway and found a few things that could be improved before we merge. None of them are critical, but now is probably a better time to address them than later.

docs/source/docformat/restructuredtext.rst
tristanlatr: I honestly don't know !

mthuurne:
The reason that mypy didn't report the same errors locally is most likely that you have an older version of mypy installed in your virtualenv. There was a new release (0.800) a week ago; after updating to that I could reproduce the errors.

PR #349, opened 2021-01-20
Document treatment of escape sequences in docstrings

### LINKED ISSUE #346: "\n" in default code directive "::" is interpreted and causes a new line in output
Using RST docformat. 
Exemple, this docstring: 

```
c) Extend both, calling the parent functions under the same
    conditions::

      def visit_example(self, node):
          if foo:
              self.body.append('<div class="foo">\n')
          else: # call the parent method
              _html_base.HTMLTranslator.visit_example(self, node)
  
      def depart_example(self, node):
          if foo:
              self.body.append('</div>\n')
          else: # call the parent method
              _html_base.HTMLTranslator.depart_example(self, node)
```

Generates: 

Capture d’écran, le 2021-01-15 à 13 31 27

mthuurne:
Closes #346

[Rendered version](https://pydoctor--349.org.readthedocs.build/en/349/codedoc.html#docstrings), last part of linked section.

### DIFF
--- a/docs/source/codedoc.rst
+++ b/docs/source/codedoc.rst
@@ -41,10 +41,23 @@ For long docstrings, start with a short summary, followed by an empty line::
         but omitted from the summary table.
         """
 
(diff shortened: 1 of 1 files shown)


adiroiban (approved):
Great update. Thanks!

PR #350, opened 2021-01-21
Format 'def func' and 'class Class:' lines using syntax highlighting

mthuurne:
Currently these code-like definition lines are formatted entirely using a bold monospace font, but it looks better and more consistent if we instead format them in the same way as code blocks.

In addition to the mentioned formatting changes, this PR also adds the use of Intersphinx to link to the documentation of external base classes.

### DIFF
--- a/docs/epytext_demo/demo_epytext_module.py
+++ b/docs/epytext_demo/demo_epytext_module.py
@@ -4,6 +4,9 @@ This is a module demonstrating epydoc code documentation features.
 Most part of this documentation is using Python type hinting.
 """
 
+from abc import ABC
+
+
 def demo_fields_docstring_arguments(m, b):  # type: ignore
     """
     Fields are used to describe specific properties of a documented object.
@@ -67,7 +70,7 @@ class _PrivateClass:
         return True
 
 
-class DemoClass:
+class DemoClass(ABC):
     """
     This is the docstring of this class.
     """
--- a/docs/epytext_demo/demo_epytext_module.py
+++ b/docs/epytext_demo/demo_epytext_module.py
@@ -5,6 +5,7 @@ Most part of this documentation is using Python type hinting.
 """
 
 from abc import ABC
+from somelib import SomeInterface
 
 
 def demo_fields_docstring_arguments(m, b):  # type: ignore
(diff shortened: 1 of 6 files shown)


adiroiban:
Just something that I have observed.

At https://pydoctor--350.org.readthedocs.build/en/350/docformat/epytext/demo/moduleIndex.html 
The description is `Scope and Purpose =================` I was expecting only `Scope and Purpose` .... but maybe as a best practice ask users to always start a module docstring using a short sentence.

mthuurne:
> Just something that I have observed.
> 
> At https://pydoctor--350.org.readthedocs.build/en/350/docformat/epytext/demo/moduleIndex.html
> The description is `Scope and Purpose =================` I was expecting only `Scope and Purpose` .... but maybe as a best practice ask users to always start a module docstring using a short sentence.

Sounds like another instance of #86.

Even "Scope and Purpose" would make for a poor summary; the first line below the heading is "Sample package for demonstrating pydoctor HTML API rendering for RST based documentation.", which would make a good summary. But in other situations the heading itself might be a better summary than the first line after it.

adiroiban (approved):
Looks good. Thanks,

I was able to see Class and attribute HTML/CSS markup here https://pydoctor--350.org.readthedocs.build/en/350/docformat/epytext/demo/epytext_demo.demo_epytext_module.DemoClass.html

---------

Maybe update the demo to contain a class that is inheriting from an intersphinx class, to also check the updated class signature.

tristanlatr:
It really looks great, love the pink.

mthuurne:
> Maybe update the demo to contain a class that is inheriting from an intersphinx class, to also check the updated class signature.

Good idea. I made `DemoClass` inherit from `abc.ABC`.

adiroiban:
In the code I see 3 cases for class signature members:

* Intersphinx link
* pydoctor internal link
* unknown link

It would be nice for the demo to have all 3 use cases.

Not necessarily in the same class... but we need to add another class for internal links :)

But this already looks good to me.

mthuurne:
> It would be nice for the demo to have all 3 use cases.
> 
> Not necessarily in the same class... but we need to add another class for internal links :)

We already had a second class in `_PrivateClass`. Linking that as a base class also makes it easier to find.

PR #351, opened 2021-01-24
Improve "Show/hide Private API"

mthuurne:
This PR moves the "Show/hide Private API" button from a floating position in the bottom-left of the viewport to the right hand side of the navigation bar. Since we have a sticky navigation bar, the button will always be on the screen, regardless of scroll position. But for narrow windows/screens, it no longer overlaps the content like the floating button did.

The show/hide state is now synced with a query argument in the location bar. This way, if you bookmark the page or send a link to someone else, the show/hide state will be preserved.

Finally, a deep link to a private API item will now automatically enable "show private API" mode.

[Demo link](https://pydoctor--351.org.readthedocs.build/en/351/api/pydoctor.astbuilder.html#_infer_type) to a private method.


pydoctor/templatewriter/util.py
```diff
@@ -19,13 +19,22 @@ def templatefile(filename):
...
+
     if label is None:
         label = o.fullName()
+
+    url = o.url
+    if url.startswith(page_url + '#'):
+        # When linking to an item on the same page, omit the path.
+        # Besides shortening the HTML, this also avoids the page being reloaded
+        # if the query string is non-empty.
+        url = url[len(page_url):]
```
adiroiban, line 34: Do we have test for this ?

I am checking the demo at https://pydoctor--351.org.readthedocs.build/en/351/docformat/epytext/demo/epytext_demo.demo_epytext_module.html#demo_fields_docstring_arguments

In the source I see

```html
<p>This function can be used in conjuction with
<a href="epytext_demo.demo_epytext_module.html#demo_typing_arguments"><code>demo_typing_arguments</code>
</a> to find an arbitrary function's zeros.
</p>
```
I was expected HREF to be just `#demo_typing_arguments`

and there is hte type `conjuction` vs conju**n**ction

pydoctor/model.py
```diff
@@ -183,22 +183,35 @@ def description(self) -> str:
...
             assert parent is not None
-            return f'{quote(parent.fullName())}.html#{quote(self.name)}'
+            return parent
         else:
             assert False, location
 
+    @property
+    def url(self) -> str:
+        """Relative URL at which the documentation for this Documentable
+        can be found.
```
adiroiban, line 206: Maybe use an explicit name ... so that you are not lost between absolute and relative URL all having URL.

And you can also have 2 types of "absolute URLs" : `/some/path/form/base.html`  vs 
`https://docs.example.com/some/path/form/base.html` 

```suggestion
    def relativeURL(self) -> str:
        """URL path and possible fragment for this Documentable,
        relative to the documentation base URL.
```

adiroiban (commented):
This is great improvement. Thanks!

I minor comment about the missing test for 

```
if url.startswith(page_url + '#'):
```

I am not sure I understood that feature correctly... but an explicit test should help clarify the expected behaviour

Thanks!

pydoctor/templatewriter/util.py
mthuurne: We have test coverage ;) ...but that coverage is meaningless since the produced links aren't checked.

I did verify the output on two projects, so I'm wondering what is different about the demo that it doesn't work there. To be continued...

mthuurne: > and there is hte type `conjuction` vs conju**n**ction

Maybe it would be a useful feature to have a spell check option for docstrings?

It's not as simple as it sounds though: to avoid false positives, we'd have to exclude for example class/function/variable names, but how those are embedded in docstrings differs per documentation format. I think we need a DOM that is doc format independent, not only for spell check but also for other features like the automatic construction of the summary line.

Possible options for the DOM would include:
- docutils
- HTML subset with well-defined semantics
- something custom

A problem with using an HTML subset is that while all our documentation format modules can convert to HTML, getting them to produce the same nodes in the same situations would probably be a significant effort. For example, if one doc format generates `<code>` and another generates `<span class="code">`, that would make it hard to process the DOM.

A problem with using something other than HTML is that we might run into problems if we ever want to support Markdown as a docstring format, since Markdown allows embedding raw HTML. On the other hand, even if we do pick HTML as the DOM, the raw HTML inside Markdown is not guaranteed to stick to our subset, so we'd need to either normalize the HTML or preserve embedded HTML as-is in a node type that we ignore during processing.

Designing something custom is always an option we can fall back on, but it shouldn't be the first thing we try, in my opinion (avoid re-inventing the wheel). Even if we do end up with a custom solution, knowing what didn't work in the previously tried approaches would help create a better design.

pydoctor/model.py
mthuurne: I don't think we should be using absolute URLs at all: if we only use relative URLs, pydoctor's output can be plugged into a site at any depth, but if we use absolute URLs the user would have to specify the final location when they generate the docs.

tristanlatr:
Would it be hard to preserve the url parameters when navigating from one page to another ?

mthuurne:
> Would it be hard to preserve the url parameters when navigating from one page to another ?

I'm not sure: if there is a global event we can intercept, it might be easy, otherwise we'd have to add event handlers to every link.

Would we want to preserve the `private` parameter? Currently the initial mode of the destination page depends on whether a public or private member was deep-linked, which might be sufficient or even desired in the majority of cases. For example, if you'd click the return type link of a private method, would you want to see the private methods of the linked class displayed by default?

pydoctor/templatewriter/util.py
mthuurne: > I did verify the output on two projects, so I'm wondering what is different about the demo that it doesn't work there. To be continued...

The reason turns out to be simple: the generation of that particular link is not done via `util.taglink()` (which I updated), but using a direct `tags.a` construction instead (which was left unchanged).

However, the solution is not as simple:

- should `resolve_identifier_xref()` call `taglink()`? this would require changing the `DocstringLinker` interface, since it would have to return a tag rather than a URL
- or should the "omit the path when link target is on the same page" behavior be refactored into a separate function?

Those two options lead to different output: `taglink()` also forces the `code` CSS class and adds a `data-type` attribute.

mthuurne: The `data-type` attribute was added in #90 and is there to tell `doc2dash` whether the linked documentable is a class, method, variable etc. The only file that the pydoctor parser of `doc2dash` searches for these `data-type` attributes is `nameIndex.html`, so we could only add them for that file and reduce the output size a bit.

I also opened a ticket to discuss whether we can export the index information in a more structured way than `data-` attributes in HTML, for example using the Intersphinx inventory: hynek/doc2dash#99.

mthuurne: Using both a `code` CSS class and `<code>` tags was adding complexity without any benefits, so I'll be moving everything over to use only `<code>` tags.

mthuurne:
Cross-references and type annotations now produce anchor-only links if the object linked to is documented on the same page.

I still need to add some unit testing for this.

pydoctor/model.py
adiroiban: then maybe rename to `pathURL()` ?

pydoctor/templatewriter/util.py
adiroiban: For spell checking and in general, yes. It would be nice to have an internal DOM and then just have convertors from Epytext or RST into this pydoctor DOM... but I am not sure if it's worth the trouble right now.


-----------

I have never used doc2dash and in fact I am not using any other API docs tools. Just reading the docs from local repo or in web browser.
But I think that making pydoctor easier for Dash would be nice.

adiroiban (approved):
Looks  good. Thanks.

Yes... the code is executed in the test, but we don't have any actual assertions / tests / checks.

Feel free to add the test or leave it as it is.

This is already a big improvement.

For persistent show/hide, I guess that we can use `localStorage.setItem` via JS...but this should be done in a separate PR.

Thanks

adiroiban (approved):
Awesome tests. Thanks!

mthuurne:
> Yes... the code is executed in the test, but we don't have any actual assertions / tests / checks.

I know; that remark was intended to highlight this as an example where there is test coverage but the functionality of the covered code isn't being tested.

> Feel free to add the test or leave it as it is.

I implemented some basic tests for the different link scenarios: not found, same page, other page, external. While this only tests cross-references and not other links, it does at least cover all possible output styles.

> For persistent show/hide, I guess that we can use `localStorage.setItem` via JS...but this should be done in a separate PR.

Do you mean a persisted Boolean per page, or a single one for all pages? Because I still have some doubts whether the latter is the desired behavior. In any case, that's something we can discuss in a new PR, if someone feels a need to work on that feature.

Thanks for the review, I'll go and merge it now.

adiroiban:
I was thinking one persisted boolean per pydoctor project... so all pages. KISS :) but yes... if someone needs it, we can discuss in the PR.

pydoctor/templatewriter/util.py
mthuurne: I think it's worth a try to convert epytext to `docutils.node`: if that works, we don't need to design our own DOM. And we might get better Sphinx integration as well.

Dash itself seems to be Mac-only, but I installed Zeal under Linux and it looks like a useful application. The documentation itself is the same as in a browser, but loading and searching is very fast.

PR #354, opened 2021-02-02
Hyperlink exception types in `raises` fields

### LINKED ISSUE #324: Hyperlink exception types in `raises` fields
The exception type in a `raises` field is also a type that we can link to its documentation.

mthuurne:
Before this feature could be implemented, some refactoring on `FieldDesc` and related code had to be performed.

[Example](https://pydoctor--354.org.readthedocs.build/en/354/api/pydoctor.sphinx.html?private=1) contains two linked exception types: `ValueError` from the standard library (linked via Intersphinx) and `InvalidMaxAge` as a custom class (internal link).

Closes #324.

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -221,7 +221,6 @@ class _EpydocLinker(DocstringLinker):
 class FieldDesc:
     _UNDOCUMENTED: ClassVar[Tag] = tags.span(class_='undocumented')("Undocumented")
 
-    kind: str
     name: Optional[str] = None
     type: Optional[Tag] = None
     body: Optional[Tag] = None
@@ -309,7 +308,7 @@ class FieldHandler:
         self.notes: List[Field] = []
         self.authors: List[Field] = []
         self.sinces: List[Field] = []
-        self.unknowns: List[FieldDesc] = []
+        self.unknowns: DefaultDict[str, List[FieldDesc]] = defaultdict(list)
 
     def set_param_types_from_annotations(
             self, annotations: Mapping[str, Optional[ast.expr]]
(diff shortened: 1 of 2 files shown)


pydoctor/epydoc2stan.py
```diff
@@ -299,24 +308,25 @@ class FieldHandler:
 
     def __init__(self, obj: model.Documentable):
         self.obj = obj
+        self.linker = _EpydocLinker(self.obj)
```
adiroiban, line 311: I guess that this is public just to be consistent with the other instance variables ... from the usage, it looks like it could be made private.

adiroiban (approved):
Looks good. Thanks. Only minor comments.

pydoctor/epydoc2stan.py
mthuurne: I don't really give public vs private much thought for pydoctor, since the API is a mess anyway. But I guess we have to start somewhere with cleaning it up.

pydoctor/test/test_epydoc2stan.py
```diff
@@ -356,7 +356,21 @@ def __init__(self, p):
...
+        """
+        @raise SpanishInquisition: If something unexpected happens.
+        """
+    ''', modname='test')
+    html = docstring2html(mod.contents['f']).split('\n')
+    assert '<a href="test.SpanishInquisition.html">SpanishInquisition</a>' in html
+
+
+def test_func_raise_missing_exception_type(capsys: CapSys) -> None:
     """Raise fields must include the exception type."""
```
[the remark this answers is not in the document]
mthuurne: I want to avoid `@raise` since that's epytext syntax, while we support other documentation formats as well (RST released and @tristanlatr has an experimental branch with Google/Numpy support).

The `test_func_raise_linked` test does check `@raise` with a type, but it doesn't verify that no warning is issued in that case, so I think there is some value in having both a correct and incorrect use of the "raise" field in this test.

I will update the docstring to mention the fact that the HTML is now checked as well.

Issue #41, opened 2015-01-12
make a function's parameters actual system objects


mwhudson:
This will make things more consistent and allow linking to them.

PR #358, opened 2021-02-07
Napoleon: Google and Numpy docstring parsing

### LINKED ISSUE #123: Support google/numpy docstrings, too
Currently, `pydoctor` seems to support only javadoc-style docstrings, however, there seems to be a trend towards Google or numpy-style docstrings (supported e.g. by the [napoleon Sphinx extension](https://pypi.python.org/pypi/sphinxcontrib-napoleon). It would be cool if `pydoctor` could accept those as well.

### LINKED ISSUE #268: Param types should be clickable

tristanlatr:
This PR adds a new module: `pydoctor.napoleon` that is a fork of napoleon sphinx extension, giving the new  "google" and "numpy" docformats. Closes #123. 

I've quite improved the original Napoleon extension in such ways that the type preprocessing is wrapped into a new object (see [TypeDocstring](https://pydoctor--358.org.readthedocs.build/en/358/api/pydoctor.napoleon.docstring.TypeDocstring.html)), this is used by default for both google and numpy docstring. Additionaly, I've added support for using the same type specification construct inside reStructuredText and Epytext `type` and `rtype` fields with the option `--process-types` (see [ParsedTypeDocstring](https://pydoctor--358.org.readthedocs.build/en/358/api/pydoctor.epydoc.markup._types.ParsedTypeDocstring.html)), this fixes #268. 

Quick demo sites, adapted from sphinx's are also available. 

It has #299 and #362  and #386 merged so I'll mark it as draft until the work on the other PR is finished.

Docformat docs: https://pydoctor--358.org.readthedocs.build/en/358/docformat/index.html
pydoctor.napoleon docs: https://pydoctor--358.org.readthedocs.build/en/358/api/pydoctor.napoleon.html
Fields docs: https://pydoctor--358.org.readthedocs.build/en/358/codedoc.html#fields


tristanlatr:
I know this is a massive PR, but it’s also massively tested. I hope it can be merged soon, I’m working on this for months.

Look at the linked issue to know what are the more subtle changés of this PR.

Thanks

pydoctor/epydoc/markup/_types.py
```diff
@@ -0,0 +1,188 @@
...
+
+            elif _type is TokenType.DELIMITER: 
+                if _token == "[": open_square_braces += 1
+                elif _token == "(": open_parenthesis += 1
+
+                if open_square_braces + open_parenthesis > 0:
+                    try: last_processed_token = combined_tokens[-1]
+                    except IndexError: 
+                        # weird
+                        combined_tokens.append((_token, _type))
```
tristanlatr, line 138: Not even sure how we could test this block honestly. Maybe a case to use `assert`  again.

pydoctor/epydoc/markup/_types.py
```diff
@@ -0,0 +1,188 @@
...
+                    else:
+                        if last_processed_token[1] is TokenType.OBJ and isinstance(last_processed_token[0], Tag): 
+                            # Merge with last Tag
+                            last_processed_token[0](_token)
+                        else:
+                            # weird
+                            combined_tokens.append((_token, _type))
```
tristanlatr, line 145: This corner case seem less difficult to test, this could happen for instance if one tries to convert a type like:
```
:type: default[str]
```

Because `default` is a special keyword. As well as `optional`. I'm not 100% sure this will get this block to be tested, though. This should render like *default*[`str`] and NOT `default[str]`, I suppose. 

Also the following types could make this block to execute, because the first token will be considered as a natural language delimiter, literal, reference or unknown not an object:

```
:type: ,[str]
:type: , [str]
:type: of [str]
:type: or [str]
:type: : [str]
:type: and [str]
```

```
:type: 'hello'[str]
:type: "hello"[str]
```

```
:type: `hello`[str]
:type: `hello <https://github.com>`_[str]
```

```
:type: **hello**[str]
```

I think we should leave a warning for those types.

pydoctor/epydoc/markup/_types.py
```diff
@@ -0,0 +1,188 @@
...
+        tokens = self._convert_obj_tokens_to_stan(self._tokens, docstring_linker)
+
+        warnings: List[ParseError] = []
+
+        converters: Dict[TokenType, Callable[[Union[str, Tag]], Union[str, Tag]]] = {
+            TokenType.LITERAL:      lambda _token: tags.span(_token, class_="literal"),
+            TokenType.CONTROL:      lambda _token: tags.em(_token),
+            TokenType.REFERENCE:    lambda _token: get_parser_by_name('restructuredtext')(_token, warnings, False).to_stan(docstring_linker) if isinstance(_token, str) else _token, 
+            TokenType.UNKNOWN:      lambda _token: get_parser_by_name('restructuredtext')(_token, warnings, False).to_stan(docstring_linker) if isinstance(_token, str) else _token, 
+            TokenType.OBJ:          lambda _token: _token, # These convertions are done in _convert_obj_tokens_to_stan()
```
tristanlatr, line 170:
```suggestion
            TokenType.OBJ:          lambda _token: _token, # These convertions (OBJ and DELIMITER) are done in _convert_obj_tokens_to_stan().
```

pydoctor/epydoc/markup/_types.py
```diff
@@ -0,0 +1,188 @@
...
+            TokenType.DELIMITER:    lambda _token: _token, 
+            TokenType.ANY:          lambda _token: _token, 
+        }
+
+        for w in warnings:
+            self.warnings.append(w.descr())
```
tristanlatr, line 176: Testing warnings would be good.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1263,19 +1266,41 @@ def parse_docstring(docstring: str, errors: List[ParseError]) -> ParsedDocstring
                     cast(str, cast(Element, field.children.pop(0)).children[0])
             else:
                 arg = None
+            
+            # TODO: Use ParsedTypeDocstring for numpy and google by default and
+            # allow the other markup to use it as well with a CLI option like --process-types
```
tristanlatr, line 1271: This is done.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1263,19 +1266,41 @@ def parse_docstring(docstring: str, errors: List[ParseError]) -> ParsedDocstring
...
+def get_parser(obj: Optional[Documentable]) -> Callable[[str, List[ParseError], bool], ParsedDocstring]:
+    """
+    Get the L{parse_docstring} function. 
+    """
+    return parse_docstring
+
+def set_nodes_parent(nodes: Iterable[nodes.Node], parent: nodes.Node) -> Iterable[None]:
+    for node in nodes:
+        node.parent = parent
+        yield node
```
tristanlatr, line 1303: This function is offered by the `epydoc.docutils` module, so it should be removed from here.

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -84,20 +88,33 @@ def parse_docstring(docstring: str, errors: List[ParseError]) -> ParsedDocstring
     @param docstring: The docstring to parse
     @param errors: A list where any errors generated during parsing
         will be stored.
+    @param processtypes: Use L{ParsedTypeDocstring} to parsed 'type' fields.
     """
     writer = _DocumentPseudoWriter()
     reader = _EpydocReader(errors) # Outputs errors to the list.
+
+    # Strip Sphinx interpreted text roles for code references: :obj:`foo` -> `foo`
```
tristanlatr, line 96:
```suggestion
    # Credits: mhils - Maximilian Hils from the pdoc repository https://github.com/mitmproxy/pdoc
    # Strip Sphinx interpreted text roles for code references: :obj:`foo` -> `foo`
```

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -250,9 +273,18 @@ def _add_field(self,
...
         field_doc = self.document.copy()
-        for child in fbody: field_doc.append(child)
-        field_pdoc = ParsedRstDocstring(field_doc, ())
-        self.fields.append(Field(tagname, arg, field_pdoc, lineno - 1))
+        for child in fbody: 
+            field_doc.append(child)
+
+        # This allows restructuredtext markup to use TypeDocstring as well with a CLI option: --process-types
+        field_parsed_doc: ParsedDocstring
+        if self._processtypes and tagname in ['type', 'rtype']:
```
tristanlatr, line 281: We repeat two times this list: ``['type', 'rtype']``, maybe it would be good to have it stored somewhere as a constant ?

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -364,6 +396,16 @@ def handle_consolidated_definition_list(self, items: Iterable[nodes.Node], tagna
     def unknown_visit(self, node: nodes.Node) -> None:
         'Ignore all unknown nodes'
 
+
+class SeeAlso(BaseAdmonition):
```
tristanlatr, line 400:
```suggestion
# Do like Sphinx does for the seealso directive. 
class SeeAlso(BaseAdmonition):
```

pydoctor/epydoc2stan.py
```diff
@@ -318,7 +324,9 @@ class Field:
     """
 
     tag: str
+    """Field tag, i.e. C{:<tag>:} """
     arg: Optional[str]
+    """Field name, i.e. C{:param <name>:}"""
```
tristanlatr, line 329:
```suggestion
    """Field argument, i.e. C{:param <argument>:}"""
```

pydoctor/epydoc2stan.py
```diff
@@ -403,17 +413,31 @@ def set_param_types_from_annotations(
...
+        self._report_unexpected_argument(field)
         if not self.return_desc:
             self.return_desc = FieldDesc()
         self.return_desc.body = field.format()
     handle_returns = handle_return
 
+    # TODO support types for yields section
+    # Currently, type will be shown as it's included in napoleon processed strings
+    # but not it's not "the right place". https://github.com/sphinx-doc/sphinx/issues/8004
+    # We could add support for ':ytype:' and use it inside pydoctor.napoleon.
```
tristanlatr, line 431: I suggest we leave this as is for now.

pydoctor/epydoc2stan.py
```diff
@@ -403,17 +413,31 @@ def set_param_types_from_annotations(
...
             self.return_desc = FieldDesc()
         self.return_desc.body = field.format()
     handle_returns = handle_return
 
+    # TODO support types for yields section
+    # Currently, type will be shown as it's included in napoleon processed strings
+    # but not it's not "the right place". https://github.com/sphinx-doc/sphinx/issues/8004
+    # We could add support for ':ytype:' and use it inside pydoctor.napoleon. 
+    def handle_yield(self, field: Field) -> None:
+        self._report_unexpected_argument(field)
```
tristanlatr, line 433: We need a test for the new sections:

```
:yields: 
:warns:
```

pydoctor/napoleon/docstring.py
```diff
@@ -0,0 +1,1617 @@
+"""
```
tristanlatr, line 1: I suggest we don't take too much time to review this file as it's pretty huge and most of it's code is coming straight from sphinx's extension "napoleon".

pydoctor/napoleon/iterators.py
```diff
@@ -0,0 +1,273 @@
+"""
```
tristanlatr, line 1: Exact same (with strict type hints) as Sphinx extension's version.

pydoctor/themes/base/apidocs.css
```diff
@@ -211,10 +211,6 @@ ul ul ul ul ul ul ul {
     padding: 2px;
 }
 
-/* Section header */
-.fieldTable tr td:first-child.fieldName {
-    width: max-content;
-}
```
tristanlatr, line 217: Not sure from where this comes from, thought.

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -169,31 +168,12 @@ def test_rst_directive_adnomitions() -> None:
...
-        </div>"""
-    assert prettify(html) == prettify(expected_html)
-
-@pytest.mark.xfail
-def test_rst_directive_deprecated() -> None:
-    html = rst2html(""".. deprecated:: 0.2
-    For security reasons""")
-    expected_html="""
-        <div class="deprecated">
-        <p><span class="versionmodified deprecated">Deprecated since version 0.2: For security reasons</span></p>
```
tristanlatr, line 197: These tests have been removed, probably, because valid versions are included here: https://github.com/twisted/pydoctor/pull/395

tristanlatr (commented):
So I've done a self-review in order to make the process less painful for the reviewer. I hope this will help. I realize there are a couple of things that should be changed, but nothing major.

pydoctor/epydoc/markup/_types.py
```diff
@@ -0,0 +1,188 @@
...
+
+    @property
+    def has_body(self) -> bool:
+        return len(self._tokens)>0
+
+    def to_node(self) -> nodes.document:
+        """
+        Not implemented.
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
tristanlatr: Complementary comment: We actually need to implement a special `to_stan` method for the `ParsedTypeDocstring` class because of the way it handles the types inside braces in a way it could not be done from Docutils nodes. 

I'll add a exhaustive comment in the code.

pydoctor/epydoc/markup/_types.py
```diff
@@ -0,0 +1,188 @@
...
+
+        for _token, _type in tokens:
+
+            if _type is TokenType.OBJ:
+                new_token = docstring_linker.link_xref(_token, _token, self._lineno)
+                if open_square_braces + open_parenthesis > 0:
+                    try: last_processed_token = combined_tokens[-1]
+                    except IndexError: 
+                        # weird
+                        combined_tokens.append((_token, _type))
```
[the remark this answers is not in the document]
tristanlatr: Or maybe it could happend with a type like the following? 

```
:type: [str]
```

pydoctor/themes/base/apidocs.css
tristanlatr: After reviewing the git blame and the referenced [commit diff](https://github.com/twisted/pydoctor/commit/abb4914b5b43042764b43a45f722e3de8901ac0a#diff-12fe32cd4f1603b1cc87fcbf51992c45e28b713d8b5ce440ee228b1ebded0dcdR200), I think it's Ok to remove it. It does not change anything special, the row always takes the maximum width possible anyhow.

pydoctor/napoleon/docstring.py
tristanlatr: Nevertheless, I’ve donne some major improvements to the original version of napoleon Sphinx extension, keeping the same behavior (This is tested). I’m pretty familiar with it now. I can ensure you that, if there is a bug/improvement to apply, I’ll look at it pretty quickly.

What I’m saying is that,  it’s a big file and to do an exhaustive review of it would probably take too much time any of the current reviewers can spend. Then, this PR will stay unmerged for ever. This would be a shame, so I’de suggest we skip exhaustive review for this particular case (I did not went yolo while coding this). Unless, we risk to never offer this feature that is google and numpy docstring parsing.

Let’s note that at least one project is using my development branch with numpy parsing to have this feature (

 Tell me what you think @adiroiban @mthuurne

tristanlatr:
Not sure to understand how the warnings filters strings works. 
We need to ignore a warning in a Sphinx module. 

Can you give me a hand, @graingert  ? Thanks.

pydoctor/epydoc2stan.py
tristanlatr: Maybe it's better to have it the right way from the start. So we should introduce this ``:ytype:`` field, I think.

pydoctor/epydoc2stan.py
tristanlatr: Done.

pydoctor/astbuilder.py
```diff
@@ -953,10 +953,14 @@ def _get_all_args() -> Iterator[ast.arg]:
             yield from base_args.args
             varargs = base_args.vararg
             if varargs:
+                # Adding the asterixes
+                varargs.arg = f"*{varargs.arg}"
```
adiroiban, line 957: shouldn't the asterixes be added in the view-rendering level?

For the low-level we try to keep the data as "pure" as possible.
Then, in the view level  / view-model / template we can decide how to show them.

Open option is to create a special class dedicated to variable or keyword arguments.
The class can wrap the actual name.

```suggestion
                varargs.arg = VariableArgument(varargs.arg)
```

Then, instead of checking if the name starts with '*' you can do an isinstance check...

but I guess that for this case, it's ok to have this convention and add the extra semantic via string characters and not via a type.

tox.ini
```diff
@@ -58,9 +59,9 @@ description = Run pyflakes over the pydoctor code
 deps: pyflakes
 
 commands =
-    ; Custom pyflakes run to exlcude test files.
+    ; Custom pyflakes run to exlcude some files.
```
adiroiban, line 62: why extend the extensions?
If for some reason non-test files need special handling for pyflakes, I think that the comment  should document that.

I think that pyflakes is the minium check that you should run to help keep the code clean.

So if possible, I would like not to skip pyflakes checks.

pydoctor/epydoc2stan.py
```diff
@@ -448,7 +475,8 @@ def _handle_param_not_found(self, name: str, field: Field) -> None:
             if name in source.constructor_params:
                 # Constructor parameters can be documented on the class.
                 return
-        field.report('Documented parameter "%s" does not exist' % (name,))
+        if not name.startswith('*'):
```
adiroiban, line 478: I guess that this is why you wanted to add the asterix ... I am not happy with the result, but you are the one implementing it...so if this is what you want, it's ok :)

adiroiban (changes requested):
Many thanks for all this huge work!

I am not happy with all the red tests.

I think that at a minimum, the PR should have a green test run.

I have only checked the common AST and RST code.

I have no reviewed napoleon  / google as I have never used that.
Also, since they are new features it's ok if they have a few bugs. 
It's more important to have the feature and people can test and give feedback.
They can be fixed later

tristanlatr:
> I am not happy with all the red tests.

Me neither, I'll fix that of course. 

> I guess that this is why you wanted to add the asterix ... I am not happy with the result

What's the result you'd want ? See https://github.com/twisted/pydoctor/issues/367 for the rationale. And see https://github.com/twisted/pydoctor/blob/6852f465455b7a72ef8ea04e2bf9b5a04a82ac21/pydoctor/test/test_epydoc2stan.py#L492 for the tests. 

Creating ``VariableArgument`` and ``KeywordArgument`` classes seem like a good option. I'll look into that.

Regarding pyflakes, I skipped the demo sites because it was triggering errors for unused imports that are used by pydoctor to resolve links.

pydoctor/epydoc2stan.py
tristanlatr: Can you explain what you mean exactly ? 
I'm happy with the result on my end, but I'de like to know what's wrong in your opinion.

Thanks

tristanlatr:
I've fixed the test, it was caused by an failed merge commit with the Github UI.

setup.cfg
```diff
@@ -77,6 +76,8 @@ doctest_optionflags = ELLIPSIS IGNORE_EXCEPTION_DETAIL
 xfail_strict = true
 filterwarnings =
     error
-
+    ignore:The distutils package is deprecated and slated for removal in Python 3.12\. Use setuptools or check PEP 632 for potential alternatives:DeprecationWarning:
+    ; for python 3.6 only:
+    ignore:.* is deprecated\. Use files\(\) instead\. .*\/\/importlib-resources\.readthedocs\.io\/en\/latest\/using\.html#migrating-from-legacy for migration advice\.:DeprecationWarning:
```
tristanlatr, line 81:
```suggestion

```

setup.cfg
```diff
@@ -77,6 +76,8 @@ doctest_optionflags = ELLIPSIS IGNORE_EXCEPTION_DETAIL
 xfail_strict = true
 filterwarnings =
     error
-
+    ignore:The distutils package is deprecated and slated for removal in Python 3.12\. Use setuptools or check PEP 632 for potential alternatives:DeprecationWarning:
+    ; for python 3.6 only:
```
tristanlatr, line 80:
```suggestion

```

tox.ini
tristanlatr, line 62:
```suggestion
    ; Custom pyflakes run to exlcude some files, namely the testing packages and the demo packages. 
```

pydoctor/astbuilder.py
tristanlatr: I don't see how we could use a different type without having another massive change. I think it should be done while fixing #41, which is not tomorrow.

tristanlatr: Or maybe `VariableArgument` and `KeywordArgument` can be subclasses of `str`, in this case it could work!

tristanlatr: I'll give it a try

pydoctor/epydoc2stan.py
tristanlatr: Here, we have to keep the ``.startswith('*')`` because, now we set the name with stars in ``_handle_param_name``, and that's the name that we have here.

tristanlatr: I think I've fixed the issue you raised.

tox.ini
tristanlatr: This is only for the demo site, so I'de say it's ok. But I can add some `#noqa` comments instead, it will probably work ? 
Oh no, pyflakes does not support those I think...

tristanlatr:
Ok, it's ready to merged in my opinion. Thanks you for the review @adiroiban !

tristanlatr:
I’ll add a comment regarding why some pyflakes tests are skipped and merge this PR.
So happy this’s is  going to be merged. Finally!

adiroiban (approved):
Please add the comment about pyflakes and merge.

Thanks and sorry for the delay

setup.cfg
```diff
@@ -77,6 +76,6 @@ doctest_optionflags = ELLIPSIS IGNORE_EXCEPTION_DETAIL
 xfail_strict = true
 filterwarnings =
     error
-
+    ignore:The distutils package is deprecated and slated for removal in Python 3.12\. Use setuptools or check PEP 632 for potential alternatives:DeprecationWarning:
```
tristanlatr, line 79:
```suggestion
    ; Sphinx imports distutils, this warning gets trigerred only in the test. See https://github.com/sphinx-doc/sphinx/issues/9820
    ignore:The distutils package is deprecated and slated for removal in Python 3.12\. Use setuptools or check PEP 632 for potential alternatives:DeprecationWarning:
```

tox.ini
tristanlatr, line 62:
```suggestion
    ; Custom pyflakes run to exlcude some files. The demo packages are trigerring some unused imports warnings. But pydoctor uses those imports to resolve links. So we skip pyflakes for the demo sites. But not the test files. 
```

PR #359, opened 2021-02-12
Fix field overriding attribute docstring

mthuurne:
I noticed the problem on `twisted.web._newclient.Request`, where for `method` and `headers` the inherited docstring was used as the summary instead of the docstring set using `@ivar`.

The Twisted 20.3.0 docs don't have this problem, while docs I generated locally with pydoctor 20.7.1 do have the problem, so it looks like this is a regression introduced in pydoctor 20.7.0. As we're dealing with a regression, I think it would be good to do a bug fix release (20.12.2) after this PR passes review.

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -613,17 +613,19 @@ def format_summary(obj: model.Documentable) -> Tag:
     """Generate an shortened HTML representation of a docstring."""
 
     doc, source = get_docstring(obj)
-    if doc is None:
+
+    if (doc is None or source is not obj) and isinstance(obj, model.Attribute):
         # Attributes can be documented as fields in their parent's docstring.
-        if isinstance(obj, model.Attribute):
-            pdoc = obj.parsed_docstring
-        else:
-            pdoc = None
-        if pdoc is None:
-            return format_undocumented(obj)
+        pdoc = obj.parsed_docstring
+    else:
+        pdoc = None
+
+    if pdoc is not None:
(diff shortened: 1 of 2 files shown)


mthuurne:
Note that the long term plan is still to produce the summary from the parsed docstring instead of the first lines of docstring source. See #86.

adiroiban (approved):
Thanks for the update.
The tests are ok.

The code might work, but it's hard to follow the logic.

I was expecting `get_docstring(obj)` to return an "undocumented" placeholder when the documentation could not be found.

In this way, we don't need a separate `format_undocumented` and it can be handled as part of `format_docstring`.

maybe have `get_direct_docstring(obj)` and `get_resolved_docstring(obj)` that will return obj.parent as source if required.

But I think that this can be merged as there is not point in not merging it.

mthuurne:
The thing that makes it complex is that we can get docstrings from different sources:
- everything can have an individual docstring
- class variables can also be documented in the class docstring
- class variables and methods can inherit docstrings

Having a placeholder wouldn't necessarily make things easier, since you'd have to be careful to only apply the placeholder if all potential documentation locations are empty.

I do think this code should be simpler, but I'm not sure yet about the order in which to change things. The steps I'm thinking of right now are:
- make docutils the official DOM for docstrings
- update the epytext parser to produce a docutils DOM
- rewrite the summary generator to use the first paragraph of the parsed docstring, or the first sentence of that paragraph (so split the parse result instead of the raw docstring)

So any cleanups we make should either prepare for that behavioral change or be done after it has landed, I think.

PR #360, opened 2021-02-14
Package is a module

mthuurne:
Previously pydoctor had a `Package` object for the namespace and a `Module` object for the code from the `__init__.py` source file. However, in the Python Language Reference, it is explicitly stated that packages are modules.

This PR makes the `Package` class inherit from the `Module` class and simplifies the related code.

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -271,10 +271,13 @@ class ModuleVistor(ast.NodeVisitor):
             return
 
         modname = node.module
-        if node.level:
+        level = node.level
+        if level:
             # Relative import.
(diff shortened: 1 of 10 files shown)


adiroiban (approved):
Sorry for the late reply.

thanks for the refactoring. it looks good.

In the last week I was busy with other projects 
 
If @tristanlatr is happy with this PR, I think that it should be merged to not block other work

I will not be able to do reviews soon :(

tristanlatr:
Yes I'm happy with this changes :)

Thanks

mthuurne:
OK, I'll merge it then.

PR #369, opened 2021-03-01
Fix rendering of project URL

mthuurne:
The navigation bar rendering was inconsistent due to duplicate code that hadn't been kept in sync. I removed the duplication by introducing a base class for all pages.

This is intended for a 21.2.1 bug fix release, so we can generate Twisted's API docs. The solution for the next major release will be to merge #299.

### DIFF
--- a/pydoctor/templatewriter/summary.py
+++ b/pydoctor/templatewriter/summary.py
@@ -242,6 +242,10 @@ class IndexPage(util.Page):
     def __init__(self, system):
         self.system = system
 
+    @renderer
+    def title(self, request, tag):
+        return tag.clear()(f"API Documentation for {self.system.projectname}")
+
     # Deprecated: pydoctor's templates no longer use this, but it is kept
     #             for now to not break customized templates like Twisted's.
     @renderer
--- a/pydoctor/templatewriter/summary.py
+++ b/pydoctor/templatewriter/summary.py
@@ -242,6 +242,12 @@ class IndexPage(util.Page):
     def __init__(self, system):
         self.system = system
 
+    # Deprecated: pydoctor's templates no longer use this, but it is kept
+    #             for now to not break customized templates like Twisted's.
+    @renderer
+    def project_link(self, request, tag):
+        return self.project_tag
+
     @renderer
     def recentChanges(self, request, tag):
         return ()
--- a/pydoctor/templatewriter/summary.py
+++ b/pydoctor/templatewriter/summary.py
@@ -29,19 +29,12 @@ def _lckey(x):
     return (x.fullName().lower(), x.fullName())
 
 
-class ModuleIndexPage(Element):
+class ModuleIndexPage(util.Page):
     filename = 'moduleIndex.html'
 
     @property
     def loader(self):
         return XMLFile(util.templatefilepath('summary.html'))
-
-    def __init__(self, system):
-        self.system = system
-
-    @renderer
-    def project(self, request, tag):
-        return self.system.projectname
 
     @renderer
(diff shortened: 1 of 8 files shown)


tristanlatr:
I feel like this issue has already been taken care of in #299 and this will create conflicts with #299 (where a base page is also introduced)

mthuurne:
If the conflicts with #299 are not easy to resolve, we could merge this PR to a `21.2.x` bug-fix branch and not to `master`.

pydoctor/templatewriter/summary.py
```diff
@@ -260,17 +242,11 @@ def loader(self):
...
-        if self.system.options.projecturl:
-            return tags.a(href=self.system.options.projecturl)(
-                self.system.projectname)
-        else:
-            return self.system.projectname
-
-    @renderer
-    def project(self, request, tag):
-        return self.system.projectname
+        return self.project_tag
```
adiroiban, line 249: I would say that project_link should also be moved to util.Page base page as it should be common.


Now I see that the template no longer uses it...but I don't know if that is the right thing to do :)

pydoctor/templatewriter/util.py
```diff
@@ -17,3 +18,22 @@ def templatefile(filename):
...
+
+class Page(Element):
+
+    def __init__(self, system: model.System):
+        super().__init__()
+        self.system = system
+
+    @property
+    def project_tag(self) -> Tag:
+        system = self.system
```
adiroiban, line 31: What is the project_tag ?

And then we also have project_link somewhere in the code.

I think that at the view-model level we should only present raw data and then it's up to the template to decide how to present it... in an A tag or a button with JS actions :)

adiroiban (approved):
Looks good. Thanks! Only minor comment. Looking forward to the new HTML template code.

I have not followed the HTML template PR and I hope it will be easy to fix the conflicts.

But I hope that we can have an Model–view–viewmodel (MVVM) patter https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel 

Model will be the low level object containing the info for an object ... raw docstring , inheritance ...

view-model will process all those information and add logic to distill that info into something that should be present to the user. It should contain as few CSS class names and HTML tags as possible.

view - HTML5 template will decide how to show the information in HTML format.

pydoctor/templatewriter/util.py
mthuurne: The old situation was a mess and we can't really clean it up without breaking compatibility. Since #299 has to break compatibility anyway, I think it's best to do cleanups there and after that has been merged.

The mess was as follows:
- a slot named `project` that was used by one template
- a renderer named `project` that was used by three templates
- a renderer named `project_link` that is not used by pydoctor, but is used by Twisted

The `project_tag` property is an attempt to make a single implementation of rendering a project name that links to the project URL that can be used by all of the three listed above.

pydoctor/templatewriter/util.py
```diff
@@ -17,3 +18,22 @@ def templatefile(filename):
 
 def templatefilepath(filename):
     return FilePath(templatefile(filename))
+
+
+class Page(Element):
```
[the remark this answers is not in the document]
mthuurne: This PR is intended for a bug-fix release. While I appreciate ideas for further cleaning up the code, I think it would be better to consider those for the next major release, and after #299 has landed.

mthuurne:
I just noticed that we're inserting a project link into a `<title>` tag, which might be invalid and is not a good idea even if it is allowed. Another case of inconsistencies between pages.

PR #370, opened 2021-03-01
Sidebar

tristanlatr:
Hello, this PR add a major usability improvement: a left sidebar (#257) that includes : 

- The Table Of Contents if titles are defined in the module/class docstring (only works with reStructuredText, should not be an issue since the plan is to convert epytext element to nodes anyway at some point). It only shows the first 4 level tho: This is arbitrary and I'll probably make it 6. 
- for classes: information about the contents of the current class and parent module/package, sorted by types. 
- for modules/packages: information about the contents of the module and parent package,  sorted by types. 

The sidebar is collapsable in the same fashion docs.python.org/3 does. 
At the same time it is sticky and allows you to scroll it independently of the main page. 

The sidebar can be schematized as follow:
```
+---------------------------
| <Table of contents of the    
| module or class                  
| docstring if titles are 
| defined>
| --------------------------
| <Contents of the current
| module or class >
| --------------------------
| <Contents of the parent 
| of the current
| module or class >
+---------------------------
```

Oh and also it stores user preferences into cookies so we don't have to re-click on collapse sidebar button. 

See example here: https://pydoctor--370.org.readthedocs.build/en/370/api/pydoctor.templatewriter.pages.sidebar.ExpandableItem.html

Details and TODOS:

- New options:
    - --sidebar-expand-depth default is 1
    - --sidebar-toc-depth default is 6
    - --no-sidebar default is False

- TODOs
    -  create an issue one this PR is merged with the following content: “With the side bar generation, especially on project that contain lots of classes per modules and with a value higher than 1 for option --sidebar-expand-depth, files can rapidly take a few megas and be really long to load, even locally. To ensure the best user experience no matter the value of `--sidebar-expand-depth`, we should generate the sidebars for each documentable in a separate file (i.e. `.sidebar.html`) and load it with JS when the rest of the content has been loaded.”

Thanks.


tristanlatr:
This PR would gain to integrate #360 , The `Package` init module thing is making this more complicated that it has to be. And I believe you fixed it in #360

tristanlatr:
After some tests with really big projects, it seems that we need to have a JS loader that will load the sidebar after the main page content has been loaded. Files can rapidly take a few megas and be really long to load, even locally.

The first step would be to generate the sidebars for each documentable in a separate file and load it with JS when the rest of the content has been loaded.

We could even go further and try to avoid generating the same modules and classes sidebar again and again for each members (with subtle differences in the highlights). Instead we could store the members data as JSON and generate the sibebar from that. But honestly, I don’t think I have the patience and JS knowledges to go that route.

tristanlatr:
>  Files can rapidly take a few megas and be really long to load, even locally.

This is mainly due to the fact that the sidebar contain expandable items, and the default value for `--sidebar-expand-depth` is 3, which is a lot when we're talking about projects with hundreds of classes.

Generating the sidebar HTML with a different file (i.e. `<fullName>.sidebar.html)` and loading it from JS can be done with a little extra effort compared to the second approach. 

So I would be inclined to do that and consider the issue fixed.

tristanlatr:
Do not forget to bump version of the template files when required.

.gitignore
```diff
@@ -10,3 +10,4 @@ _trial_temp/
 apidocs/
 *.egg-info
 .eggs
+*.docset
```
tristanlatr, line 13:
```suggestion
```

pydoctor/epydoc/markup/plaintext.py
```diff
@@ -48,6 +48,9 @@ def __init__(self, text: str):
     def has_body(self) -> bool:
         return bool(self._text)
     
+    def get_toc(self, depth: int) -> None:
+        return None
```
tristanlatr, line 52:
```suggestion
```

pydoctor/epydoc/markup/__init__.py
```diff
@@ -109,14 +111,21 @@ def __init__(self, fields: Sequence['Field']):
 
         self._stan: Optional[Tag] = None
 
-    @abc.abstractproperty
+    @property
```
tristanlatr, line 114: Maybe it's best to keep it abstract?

pydoctor/driver.py
```diff
@@ -432,6 +448,16 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
...
 
+        # check if sidebar related arguments are valid
+        if system.options.sidebarexpanddepth < 1:
+            system._warning(None, "Invalid --sidebar-expand-depth value", detail='The value of --sidebar-expand-depth option should be greater or equal to 1, '
+                                'to suppress sidebar generation all together: use --no-sidebar')
+            system.options.sidebarexpanddepth = 1
+        if system.options.sidebartocdepth < 0:
+            system._warning(None, "Invalid --sidebar-toc-depth value", detail='The value of --sidebar-toc-depth option should be greater or equal to 0, '
+                                'to suppress sidebar generation all together: use --no-sidebar')
+            system.options.sidebartocdepth = 0
```
tristanlatr, line 459: I should add a test that ensure those warnings are getting trigered.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1355,6 +1355,9 @@ def __str__(self) -> str:
     def has_body(self) -> bool:
         return self._tree is not None
     
+    def get_toc(self, depth: int) -> None:
+        return None
+
```
tristanlatr, line 1360:
```suggestion
```

pydoctor/epydoc/markup/__init__.py
```diff
@@ -134,7 +143,6 @@ def to_stan(self, docstring_linker: 'DocstringLinker') -> Tag:
         self._stan = Tag('', children=node2stan.node2stan(self.to_node(), docstring_linker).children)
         return self._stan
     
-    @abc.abstractmethod
```
tristanlatr, line 137: I think it's best to keep it abstract.

pydoctor/epydoc2stan.py
```diff
@@ -844,6 +844,18 @@ def get_parsed_type(obj: model.Documentable) -> Optional[ParsedDocstring]:
 
     return None
 
+def format_toc(obj: model.Documentable) -> Optional[Tag]:
+    # Load the parsed_docstring if it's not already done. 
+    if not obj.parsed_docstring and obj.docstring:
+        obj.parsed_docstring = parse_docstring(obj, obj.docstring, obj)
```
tristanlatr, line 850: This is not tested because the of order of execution of methods (but it's there just in case basically because we never know), utimately we'de like all docstrings to be parsed at the same time, in a specific step of the process.

pydoctor/epydoc2stan.py
```diff
@@ -844,6 +844,18 @@ def get_parsed_type(obj: model.Documentable) -> Optional[ParsedDocstring]:
...
+
+    if obj.parsed_docstring:
+        if obj.system.options.sidebartocdepth != 0:
+            toc = obj.parsed_docstring.get_toc(depth=obj.system.options.sidebartocdepth)
+            if toc:
+                return toc.to_stan(_EpydocLinker(obj))
```
tristanlatr, line 856: Why is this untested?

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -279,6 +270,18 @@ def functionExtras(self, ob: model.Documentable) -> List["Flattenable"]:
...
         return self.docgetter.get(ob)
 
+    @renderer
+    def maindivclass(self, request: IRequest, tag: Tag) -> str:
+        return 'nosidebar' if self.ob.system.options.nosidebar else ''
+
+    @renderer
+    def sidebarcontainer(self, request: IRequest, tag: Tag) -> Union[Tag, str]:
+        if self.ob.system.options.nosidebar:
+            return ""
```
tristanlatr, line 280: --nosidebar option should be tested.

pydoctor/templatewriter/pages/sidebar.py
```diff
@@ -0,0 +1,425 @@
...
+
+    @renderer
+    def docstringToc(self, request: IRequest, tag: Tag) -> Union[Tag, str]:
+        
+        toc = self.docgetter.get_toc(self.ob)
+
+        # Only show the TOC if visiting the object page itself, in other words, the TOC do dot show up
+        # in the object's parent section or any other subsections except the main one.
+        if toc and self.documented_ob == self.ob:
+            return tag.fillSlots(titles=toc)
```
tristanlatr, line 208: Test this.

tristanlatr (commented):
A few tests to add...

tristanlatr:
Todo: create an issue one this PR is merged with the following content: “With the side bar generation, especially on project that contain lots of classes per modules and with a value higher than 1 for option `--sidebar-expand-depth` (default is 2), files can rapidly take a few megas and be really long to load, even locally. To ensure the best user experience no matter the value of `--sidebar-expand-depth`, we should generate the sidebars for each documentable in a separate file (i.e. <fullName>.sidebar.html)  and load it with JS when the rest of the content has been loaded.”

tristanlatr:
@adiroiban can you do a review of the design/usability of the table of contents for now ? Thanks

adiroiban:
it looks good. Thanks.

I am not the best person to evaluate API docs tools. I don't use such tools on a daily bases... I just read the code :)

If you are happy, we can have it released and then users can send feedback and improvements and suggestions

tristanlatr:
Maybe it would be good to have a feedback from people that actually use pydoctor interface more often. I don’t know if @glyph or @graingert could give a hand. Thanks.

adiroiban:
Some people might be busy :) 
I guess is best to just release this feature and get more feedback from end users :)

tristanlatr:
@iosonofabio 

While tou are there, can you give a little feedback on the style/usability of the new sidebar ? Would you use it for python-igraph ?

Thanks!

tristanlatr:
@ntamas maybe you have an opinion? Thanks

ntamas:
Seems like a good idea to me! In the mid-term we plan to do away with most of the custom styling we apply on the generated docs and we'll just go for the standard PyDoctor layout (we are not super picky any more about matching the style of the igraph homepage), and in this case this will definitely be useful.

Summon @iosonofabio for a second opinion.

iosonofabio:
I sensed vibrations of my name in the ether and came at once. The example link you posted looks super useful.

I'd wonder about integration with sphinx, readthedocs, and related sidebars though, is that something you want to worry about? As @ntamas says, probably fine for python-igraph as we abandon our branding in favour of doc usability.

tristanlatr:
Thanks for your feedback, to disable the collapsing items, we could set the value of option --sidebar-expand-depth to 1 by default. 

Do you also mean we should not  list the parent members items sidebar ?

not-my-profile:
Ok I am fine with showing the parent members. I don't like the expanding / collapsing at all. I think users would rather just search for something by typing what they're looking for instead of collapsing/expanding stuff in the sidebar. If we can get #381, I don't think we need the collapsing.

Regarding the styling: I think it would look better if the headings in the sidebar were centered. And I'd remove the darker background for the parent section.

I'd also prefer the sidebar to be to the left of where the documentation currently is:

current master:
image

current PR:

image

I'd rather have:

image

tristanlatr:
Collapsing/Expanding was part of the initial motivation of the sidebar, though I get your point. I think we should offer this feature as an option anyway, it can be useful not avid clicking to much around, especially in case where there are inner classes, like in django models.

I’ve tried centered headings and it doesn’t feel right to me.

> And I'd remove the darker background for the parent section.

what would you instead to make it clear it’s not the same section ? 

About the header and the sidebar location, I find it better the way it is now, but I’m not too attached to this, if there are other people that feels the same way as you, I would be inclined to change it. Maybe the readthedocs theme is designed in a better way regarding this specific issue.

glyph:
@tristanlatr OK, looking at this one instead :-) But I see a ton of conflicts here too. Can you resolve them?

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -62,26 +63,17 @@ def format_signature(function: model.Function) -> "Flattenable":
...
     Common navigation header.
     """
 
     filename = 'nav.html'
 
-    def __init__(self, system: model.System, loader: ITemplateLoader) -> None:
+    def __init__(self, system: model.System, loader: ITemplateLoader, template_lookup: TemplateLookup) -> None:
         super().__init__(loader)
         self.system = system
+        self.template_lookup = template_lookup
```
tristanlatr, line 76: Not sure why passing the template_lookup is required here?

pydoctor/themes/base/apidocs.css
```diff
@@ -269,18 +287,23 @@ ul ul ul ul ul ul ul {
         width: 35%;
     }
 
+    /* Argument name + type column table  */
+    .fieldTable tr td.fieldArgContainer {
+        width: 175px;
+    } 
+
```
tristanlatr, line 294: Is this required ?
```suggestion

```

Th current width is 250px and it's already a bit small...

pydoctor/themes/base/apidocs.css
```diff
@@ -269,18 +287,23 @@ ul ul ul ul ul ul ul {
...
 
 }
 
-@media only screen and (max-width: 650px) {
+@media only screen and (max-width: 750px) {
     /* Argument name + type column table  */
     .fieldTable tr td.fieldArgContainer {
-        width: 175px;
-    }
+        width: 155px;
```
tristanlatr, line 305:
```suggestion
        width: 175px;
```

pydoctor/themes/base/common.html
```diff
@@ -8,57 +8,77 @@
```
tristanlatr, line 8: I should update pydoctor-template-version

pydoctor/themes/base/index.html
```diff
@@ -8,11 +8,13 @@
```
tristanlatr, line 8: Update version as well

pydoctor/themes/base/nameIndex.html
```diff
@@ -9,9 +9,10 @@
```
tristanlatr, line 9: Update version

pydoctor/epydoc/markup/epytext.py
```diff
@@ -445,8 +478,9 @@ def _add_section(
     # Colorize the heading
     head = _colorize(heading_token, errors, 'heading')
 
-    # Add the section's and heading's DOM elements.
-    sec = Element('section')
+    # Add the section's and heading's DOM elements. 
+    # Add 'ids' attribute just like in docutils.
+    sec = Element('section', ids=[slugify(' '.join(gettext(head)))])
```
tristanlatr, line 483: Special care should be taken to ensure we don't generate twice the same ID for the section header.

pydoctor/test/epydoc/test_restructuredtext.py
```diff
@@ -213,3 +213,78 @@ def test_rst_directive_seealso() -> None:
...
+    assert not errors
+    
+    toc = parsed.get_toc(4)
+    assert toc is not None
+    html = flatten(toc.to_stan(NotFoundLinker()))
+    
+    expected_html="""
+     <li>
+  <p class="rst-first">
+   <a class="rst-reference internal" href="#rst-titles" id="rst-id1">
```
tristanlatr, line 253: Using the same slugify function as in epytext would produce nicer IDs.

tristanlatr:
I've solve the conflicts and I think we're getting closer. 

@glyph, I you'de like to have a look and leave your review that would be very appreciated. 
Note that it's overhaul missing a few tests, but this part of the code is a bit hard to test :/ I'll see what I ca do. 

Thanks.

pydoctor/epydoc/markup/plaintext.py
```diff
@@ -47,7 +47,7 @@ def __init__(self, text: str):
     @property
     def has_body(self) -> bool:
         return bool(self._text)
-    
+
```
tristanlatr, line 50: undo this change.

pydoctor/themes/base/summary.html
```diff
@@ -8,9 +8,10 @@
 
     <div t:render="header" />
```
tristanlatr, line 10: I should update template version here too

pydoctor/test/test_node2stan.py
```diff
@@ -84,4 +84,4 @@ def test_gettext() -> None:
     '''
 
     assert gettext(rst2node(doc)) == ['This paragraph is not in any section.', 
-    'mailto:postmaster@example.net', 'This is just a note with nested contents']
+    'mailto:postmaster@example.net', 'This is just a note with nested contents']
```
tristanlatr, line 87:
```suggestion
    'mailto:postmaster@example.net', 'This is just a note with nested contents']

```

pydoctor/epydoc/markup/epytext.py
```diff
@@ -445,7 +478,7 @@ def _add_section(
     # Colorize the heading
     head = _colorize(heading_token, errors, 'heading')
 
-    # Add the section's and heading's DOM elements.
+    # Add the section's and heading's DOM elements.
```
tristanlatr, line 481:
```suggestion
    # Add the section's and heading's DOM elements.
```

tristanlatr:
Ok we have a problem: with the current default configuration,`tox -e cpython-apidocs` took (in the github CI) 4h 2m 37s. So this is a lot of time :/ I think we need to set the default value of --sidebar-expand-depth to 1.

pydoctor/options.py
```diff
@@ -210,6 +210,18 @@ def get_parser() -> ArgumentParser:
     parser.add_argument(
         '--pyval-repr-linelen', dest='pyvalreprlinelen', default=80, type=int, metavar='INT',
         help='Maxinum number of caracters for a constant value representation line. Use 0 for unlimited.')
+    parser.add_argument(
+        '--sidebar-expand-depth', metavar="INT", type=int, default=2, dest='sidebarexpanddepth',
+        help=("How many nested modules and classes should be expandable, "
+              "first level is always expanded. (default: 2)"))
```
tristanlatr, line 216:
```suggestion
        '--sidebar-expand-depth', metavar="INT", type=int, default=1, dest='sidebarexpanddepth',
        help=("How many nested modules and classes should be expandable, "
              "first level is always expanded, nested levels can expand/collapse. Value should be 1 or greater. (default: 1)"))
```

tristanlatr:
This PR is good à for a review

tristanlatr:
43m 45s is now needed to generate cpython apidocs with the sidebar (and about 21 mn without). It's the only system test that gain as much of build time proportionnally (about 50% more time). 

For the sake of the speed, I think I'll use option `--no-sidebar` for this system test.

glyph:
> 43m 45s is now needed to generate cpython apidocs with the sidebar (and about 21 mn without). It's the only system test that gain as much of build time proportionnally (about 50% more time).

I apologize for my inability to maintain focus on this fairly complex set of changes, but if it's taking these incredibly huge amounts of time to generate these sidebars, how much *data* are we producing? Loading the pages and clicking around seems good (I am VERY excited to have these available on the public API docs!) but I'm not seeing gigabytes of data getting loaded into my browser, so I'm confused about where these hours and hours of build times are going.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1428,7 +1473,10 @@ def _to_node(self, tree: Element) -> Iterable[nodes.Node]:
             yield set_node_attributes(nodes.doctest_block(tree.children[0], tree.children[0]), document=self._document)
         elif tree.tag in ('fieldlist', 'tag', 'arg'):
             raise AssertionError("There should not be any field lists left")
-        elif tree.tag in ('section', 'epytext'):
+        elif tree.tag == 'section':
+            yield set_node_attributes(nodes.section('', ids=[self._slugify(' '.join(gettext(tree.children[0])))]),
```
tristanlatr, line 1477: This line might require more check to ensure children is not an empty list?

tristanlatr:
@glyph I think we are good to merge this PR, I'll just release another version of pydoctor before and then let's merge this PR.

tristanlatr:
@glyph, can I get a approval on this PR ?

Thanks

tristanlatr:
@adiroiban maybe ?

adiroiban (approved):
If you think that this is useful this can be merged.
There is `--no-sidebar` if people don't like it.

And if someone has feedback or suggested changes, it can always be changed in a future PR.

--------

I am not a heave user of API documentation in HTML form.

I am using the Python stdlib docs, but the sidebar over there does something else.

So I can't provide any usability feedback.

Thanks!

docs/source/conf.py
```diff
@@ -107,6 +107,7 @@
         ] + _common_args,
     'custom_template_demo': [
         '--html-output={outdir}/custom_template_demo/',
+        '--sidebar-expand-depth=1',
```
tristanlatr, line 110: `--sidebar-expand-depth=1` is the default anyway now.

```suggestion
```

tristanlatr:
Ok I'm merging this, thanks for reviewing.
Now let's look at #457

PR #371, opened 2021-03-07
Fix positioning of anchors using CSS tweaks

mthuurne:
One problem was introduced between 20.12.x and 21.2.x when I changed the page header to be a flexbox: this somehow makes browsers treat the margin on the `h1` differently.

Another problem was an inconsistency between how Chrome and Firefox render the page: Chrome considered the whitespace-only anchor to have line height 0, while Firefox applied the font's line height. By setting the font size to 0, both browsers use a line height of 0.

### DIFF
--- a/pydoctor/templates/apidocs.css
+++ b/pydoctor/templates/apidocs.css
@@ -31,7 +31,8 @@ body {
 
 a[name] {
     position: relative;
-    bottom: 62px;
+    bottom: 80px;
+    font-size: 0;
 }
 
 ul {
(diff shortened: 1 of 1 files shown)


mthuurne:
To test: follow [this link](https://pydoctor--371.org.readthedocs.build/en/371/api/pydoctor.astbuilder.html#node2fullname) and check whether the `def note2fullname` is visible at the top of the screen.

adiroiban (approved):
Looks good. Thanks.

For reference this is the current link that is broken ... https://pydoctor.readthedocs.io/en/latest/api/pydoctor.astbuilder.html#node2fullname ... but it looks like we don't have RTD builds for older version... so the link will no longer be valid after the next release...
but that is fine.

All good. Thanks!

mthuurne:
Thanks for reviewing; 21.2.2 is now available on PyPI.

PR #374, opened 2021-03-23
Update list of projects using ``pydoctor``

tristanlatr:


mthuurne:
Maybe we could also/instead link to the API docs of those projects? That's probably the thing that readers are interested in.

tristanlatr:
> That's probably the thing that readers are interested in.

It's usually possible to find it in a few clicks...
Also I've linked to the home pages because not all projects configure `--project-url` inside their API docs.

tristanlatr:
Can this be merged as is ?

mthuurne:
Yes, I'll merge it now. It would be silly to not apply a clear improvement just because it could be made better.

PR #378, opened 2021-03-25
Add support for signatures of functions and methods in extensions

ntamas:
This PR adds support for handling the signatures of functions and methods of classes declared in C extensions. The proposed changes in this PR were necessary to build the documentation of `python-igraph` using PyDoctor (see the results [here](https://igraph.org/python/doc/api/index.html)).

*Summary of changes:*

* `_introspectThing()` now handles additional types that are used when a method comes from a class declared in a C extension
* the function model was extended to provide support for Python's `__text_signature__` property
* the base template writer was extended to make use of the extended function model

*Detailed explanation of the changes:*

In particular, the function model was extended with a `text_signature` string member and the `signature` member was made optional. When introspecting a function (i.e. anything that is an instance of `BuiltinFunctionType`, `FunctionType`, `MethodDescriptorType` or `ClassMethodDescriptorType` -- the latter two are responsible for handling methods of classes and class instances in C extensions), the `signature()` function of the `inspect` module is used first to extract the function signature. Should it fail, we fall back to the `__text_signature__` property, which is filled by Python itself automatically if the docstring starts with `objectname(...)\n--\n\n`. Note that this should happen only if the text signature is mistyped -- if it is a valid Python signature, the `inspect` module should have already parsed it to provide the signature (see [here](https://github.com/python/cpython/blob/9f220e4968cf73fa60440120ee46881e7974e47d/Lib/inspect.py#L1958-L2093)).

The base template writer was also extended to use `text_signature` as a fallback if the signature is missing.

### DIFF
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -550,14 +550,13 @@ T = TypeVar('T')
 
 # Declare the types that we consider as functions (also when they are coming
 # from a C extension)
-func_types = [types.BuiltinFunctionType, types.FunctionType]
+func_types: Tuple[Type[Any], ...] = (types.BuiltinFunctionType, types.FunctionType)
 if hasattr(types, "MethodDescriptorType"):
     # This is Python >= 3.7 only
-    func_types.append(types.MethodDescriptorType)
+    func_types += (types.MethodDescriptorType, )
 if hasattr(types, "ClassMethodDescriptorType"):
     # This is Python >= 3.7 only
-    func_types.append(types.ClassMethodDescriptorType)
-func_types = tuple(func_types)
+    func_types += (types.ClassMethodDescriptorType, )
 
 
 class System:
@@ -822,7 +821,6 @@ class System:
 
     def _introspectThing(self, thing: object, parent: Documentable, parentMod: _ModuleT) -> None:
         for k, v in thing.__dict__.items():
-            # TODO(ntamas): MethodDescriptorType and ClassMethodDescriptorType are Python 3.7 only.
             if (isinstance(v, func_types)
                     # In PyPy 7.3.1, functions from extensions are not
                     # instances of the abstract types in func_types
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -553,9 +553,13 @@ func_types: Tuple[Type[Any], ...] = (types.BuiltinFunctionType, types.FunctionTy
 if hasattr(types, "MethodDescriptorType"):
     # This is Python >= 3.7 only
     func_types += (types.MethodDescriptorType, )
+else:
+    func_types += (type(str.join), )
 if hasattr(types, "ClassMethodDescriptorType"):
     # This is Python >= 3.7 only
     func_types += (types.ClassMethodDescriptorType, )
+else:
+    func_types += (type(dict.__dict__["fromkeys"]), )
 
 
 class System:
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -528,7 +528,6 @@ class Function(Inheritable):
     annotations: Mapping[str, Optional[ast.expr]]
     decorators: Optional[Sequence[ast.expr]]
     signature: Optional[Signature]
-    text_signature: str = ""
 
     def setup(self) -> None:
         super().setup()
@@ -831,8 +830,12 @@ class System:
                 f.decorators = None
                 try:
                     f.signature = signature(v)
-                except Exception:
-                    f.text_signature = (getattr(v, "__text_signature__") or "") + " (invalid)"
+                except ValueError:
+                    # function either has an invalid text signature or no signature
+                    # at all. We distinguish between the two by looking at the
+                    # __text_signature__ attribute
+                    if getattr(v, "__text_signature__", None) is not None:
+                        parent.report("Cannot parse signature of {0.name}.{1}".format(parent, k))
                     f.signature = None
                         
                 f.is_async = False
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -15,7 +15,7 @@ import platform
 import sys
 import types
 from enum import Enum
-from inspect import Signature
+from inspect import signature, Signature
 from optparse import Values
 from pathlib import Path
 from typing import (
@@ -527,7 +527,8 @@ class Function(Inheritable):
     is_async: bool
     annotations: Mapping[str, Optional[ast.expr]]
     decorators: Optional[Sequence[ast.expr]]
-    signature: Signature
+    signature: Optional[Signature]
+    text_signature: str = ""
 
     def setup(self) -> None:
         super().setup()
@@ -545,6 +546,19 @@ _ModuleT = Module
 _PackageT = Package
 
 T = TypeVar('T')
+
+
+# Declare the types that we consider as functions (also when they are coming
+# from a C extension)
+func_types = [types.BuiltinFunctionType, types.FunctionType]
+if hasattr(types, "MethodDescriptorType"):
+    # This is Python >= 3.7 only
+    func_types.append(types.MethodDescriptorType)
+if hasattr(types, "ClassMethodDescriptorType"):
+    # This is Python >= 3.7 only
+    func_types.append(types.ClassMethodDescriptorType)
+func_types = tuple(func_types)
+
 
 class System:
     """A collection of related documentable objects.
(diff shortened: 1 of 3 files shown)


ntamas:
Static checks are complaining about the fact that `func_types` (which is used in an `isinstance()` call) is first built as a list and then later turned into a tuple. I can change it to being a tuple from the very beginning if that's a problem.

pydoctor/model.py
```diff
@@ -808,15 +822,23 @@ def ensurePackage(self, package_full_name: str) -> _PackageT:
 
     def _introspectThing(self, thing: object, parent: Documentable, parentMod: _ModuleT) -> None:
         for k, v in thing.__dict__.items():
-            if (isinstance(v, (types.BuiltinFunctionType, types.FunctionType))
+            # TODO(ntamas): MethodDescriptorType and ClassMethodDescriptorType are Python 3.7 only.
```
tristanlatr, line 825: I think this is done :)
Thanks

tristanlatr (changes requested):
Hi,

This all looks good. 

Would be cool if `unittests` were added tho, are we even testing the code supporting C-modules at all currently ? How hard would it be to generate native test modules in C as part of the tests? 

Anyway, I think it this PR can me merged without tests since it clearly fixes a bug. 

But if we want to offer more serious c-module support, we should think about testing ^^. 

For the static checks, can you use a tuple to content mypy ?

Thanks again!

pydoctor/model.py
```diff
@@ -808,15 +822,23 @@ def ensurePackage(self, package_full_name: str) -> _PackageT:
...
+                    # instances of the abstract types in func_types
+                    or (hasattr(v, "__class__") and v.__class__.__name__ == 'builtin_function_or_method')):
                 f = self.Function(self, k, parent)
                 f.parentMod = parentMod
                 f.docstring = v.__doc__
                 f.decorators = None
-                f.signature = Signature()
+                try:
+                    f.signature = signature(v)
+                except Exception:
```
mthuurne, line 836: Can we be more specific than `Exception`? I'm worried about accidentally catching exceptions that are caused by bugs rather than broken input.

Looking at the `inspect` module's source, the exceptions that can be raised are `ValueError`, `TypeError` and `RuntimeError`. I assume `RuntimeError` will only be raised when something is inconsistent within CPython itself, so we shouldn't have to catch that. The same might be true for `TypeError`, but I'm less confident there.

pydoctor/model.py
```diff
@@ -808,15 +822,23 @@ def ensurePackage(self, package_full_name: str) -> _PackageT:
...
+                    f.text_signature = (getattr(v, "__text_signature__") or "") + " (invalid)"
```
mthuurne, line 837: If the input is invalid, pydoctor should report a warning. This increases the chance that the developer notices the problem (for example by using `--warnings-as-errors` in their CI), since it's unlikely developers will be reading their own API docs output regularly (it is mostly for the benefit of the users).

pydoctor/model.py
```diff
@@ -546,6 +547,19 @@ class Attribute(Inheritable):
...
+# Declare the types that we consider as functions (also when they are coming
+# from a C extension)
+func_types = [types.BuiltinFunctionType, types.FunctionType]
+if hasattr(types, "MethodDescriptorType"):
+    # This is Python >= 3.7 only
+    func_types.append(types.MethodDescriptorType)
+if hasattr(types, "ClassMethodDescriptorType"):
+    # This is Python >= 3.7 only
+    func_types.append(types.ClassMethodDescriptorType)
+func_types = tuple(func_types)
```
mthuurne, line 560: > Static checks are complaining about the fact that `func_types` (which is used in an `isinstance()` call) is first built as a list and then later turned into a tuple. I can change it to being a tuple from the very beginning if that's a problem.

Yes, we do want all static checks to pass. A possible solution would be to define `iter_func_types()` and capture the result of that in a tuple.

mthuurne (changes requested):
Thanks for the patch!

> In particular, the function model was extended with a `text_signature` string member and the `signature` member was made optional. When introspecting a function (i.e. anything that is an instance of `BuiltinFunctionType`, `FunctionType`, `MethodDescriptorType` or `ClassMethodDescriptorType` -- the latter two are responsible for handling methods of classes and class instances in C extensions), the `signature()` function of the `inspect` module is used first to extract the function signature. Should it fail, we fall back to the `__text_signature__` property, which is filled by Python itself automatically if the docstring starts with `objectname(...)\n--\n\n`. Note that this should happen only if the text signature is mistyped -- if it is a valid Python signature, the `inspect` module should have already parsed it to provide the signature (see [here](https://github.com/python/cpython/blob/9f220e4968cf73fa60440120ee46881e7974e47d/Lib/inspect.py#L1958-L2093)).

If the first line of the docstring is used to populate `__text_signature__`, does that line remain in the docstring or is it removed from there before pydoctor does the introspection?

What does the lack of `MethodDescriptorType` and `ClassMethodDescriptorType` mean when using introspection on Python 3.6? Are there other ways we could detect these method, for example by looking at `__class__.__name__` like we do for pypy?

> The base template writer was also extended to use `text_signature` as a fallback if the signature is missing.

On the one hand, showing a broken signature is more informative to the reader than showing nothing, but on the other hand I don't really like the idea of having an attribute that is only used to preserve broken input. Would it be a problem for you if broken signatures become `...` in the output? (after a warning is logged, see code diff comment)

> Would be cool if `unittests` were added tho, are we even testing the code supporting C-modules at all currently ? How hard would it be to generate native test modules in C as part of the tests?

We do have some minimal introspection tests: `test_introspection_python()` and `test_introspection_extension()` in `pydoctor/test/test_model.py`.

The documentation of `MethodDescriptorType` mentions `str.join()` as an example and `ClassMethodDescriptorType` mentions `dict.__dict__['fromkeys']`. So perhaps we can use the standard library (at least when running on CPython) as test data?

If that is insufficient, we could add a separate mini testsuite using `python-igraph` to `tox.ini`, similar to how we parse CPython's source as a robustness check.

> Anyway, I think it this PR can me merged without tests since it clearly fixes a bug.
> 
> But if we want to offer more serious c-module support, we should think about testing ^^.

Well, I wouldn't want to refuse the PR because of lack of tests, but accepting it without tests means that we might break introspection again in future changes, so it would also be in the interest of `python-igraph` to at least have some test cases that fail on the old code.

ntamas:
> If the first line of the docstring is used to populate `__text_signature__`, does that line remain in the docstring or is it removed from there before pydoctor does the introspection?

It is removed; I'm not sure where this happens within Python itself, but the docstring only contains the part after the text signature.

> What does the lack of MethodDescriptorType and ClassMethodDescriptorType mean when using introspection on Python 3.6? Are there other ways we could detect these method, for example by looking at `__class__.__name__` like we do for pypy?

Something like this could do the trick:

```
try:
    from types import MethodDescriptorType
except ImportError:
    MethodDescriptorType = type(str.join)
try:
    from types import ClassMethodDescriptorType
except ImportError:
    ClassMethodDescriptorType = type(dict.__dict__['fromkeys'])
```

The examples are taken explicitly from the Python documentation. What do you think?

pydoctor/model.py
ntamas: I'm not sure whether it's good to tie the set of exceptions that we catch to a particular implementation of `inspect.signature()`. The implementation might change any time and since there is no documentation about the exceptions that `inspect.signature()` may throw, I would say it's better to catch everything. After all, the idea is to parse `v` as a signature, and if it fails for any reason, just fall back to the text signature; we don't want to fail the entire documentation build because of that.

Anyhow, if you prefer to be more restrictive here, just let me know which exceptions I should catch and I'll be happy to change it.

ntamas: It seems like `ValueError` is the most common exception; it gets raised when there is no text signature on a method/function from a C extension as well as when there _is_ a text signature but it cannot be parsed. One can distinguish between the two by looking at the presence of the `__text_signature__` attribute. I'll change my code to catch `ValueError` only and we can extend it later if it turns out that there are legitimate cases when Python raises something else and not a `ValueError`.

ntamas:
> Would it be a problem for you if broken signatures become `...` in the output? (after a warning is logged, see code diff comment)

No, that's totally fine for me. I have updated the PR in a way that hopefully addresses all concerns outlined above.

As for the tests:

* I have extended `test_model.py` slightly to test the signature of the introspected function.
* I have tried to cover the code branch where a warning is emitted for invalid signatures, but unfortunately I cannot do that without compiling a C extension that has a function with an invalid signature. Adding an invalid text signature to a pure Python function does not work (`__text_signature__` is not filled for pure Python functions, only for C extensions), and I could not find any function in the Python standard library that has an invalid signature, so this branch is currently left uncovered. I tested this branch manually with `python-igraph` by intentionally breaking one of the text signatures.

ntamas:
Regarding the PyPy failure: does anyone know why class methods are not callable in PyPy? I have zero experience with PyPy so I don't see any logical explanation.

tristanlatr:
I'll try to pick-up this PR since it's been a while and it fixes a bug.

I would be inclined to place all code related to introspection in a new module. 

Also I need to find a way to generated a c-extension module as part of the tests in order to preperly test all use cases.

tristanlatr:
I've been able to produce a small C extension with valid in invalid signatures. I'll add the required tests to this PR. 
About pypy failure, I don't think it should be a blocker right now, C-extensions are not natively suported by pypy anyway.

tristanlatr:
I'll close this PR since #463 includes your changes @ntamas ! Thanks a lot.

Issue #364, opened 2021-02-20
Exceptions types


tristanlatr:
Exceptions shows up as `Class` in the summary list, it should show `Exception` instead. 
Same goes for the `kind`. 
Should show `exception documentation` in the "class page" instead.

mthuurne:
Technically they are classes, but I agree that they're a special category of classes that is worth giving a dedicated `kind`.

Perhaps this is a good time to convert `kind` from a `str` to an `Enum`. It has bothered me for some time that something so core to our structure is implemented as a free-form field. Maybe we should go even further and compute `kind` in each `Documentable` subclass based on other properties of the documented object.

mthuurne:
Slightly off-topic, but another thing that bothers me about our current use of `kind` is that most module "variables" are actually constants, but listed in the docs as variables. It would be nice if we could express that, for example using `typing.Final`, using a docstring field or by code analysis (variable that's assigned only once of an immutable type).

adiroiban:
I would not bother with code analysis.

As a start, I think that support for "constants" via `typing.Final` would be a nice feature.

tristanlatr:
There is an implementation of detecting a exception here https://github.com/tristanlatr/pydocspec/blob/1efc4261a83c306391a97a61112d88b4beef7488/pydocspec/__init__.py#L650

Issue #305, opened 2020-12-01
Handling of constructors


mthuurne:
Currently we don't explicitly handle constructors in the output. Instead, we document methods like `__init__()` and `__new__()` and leave it up to the reader to figure out what the constructor arguments are (`__init__()` is not the constructor, but often has the same arguments as the constructor).

The standard library isn't consistent in documenting constructors: sometimes they're documented on the class itself, sometimes like a function named after the class, sometimes they refer to the `__init__()` documentation.

epydoc allows constructor arguments to be documented on the class. We currently just show these below the class docstring, but that is not consistent with classes where the constructor arguments are documented on `__init__()`.

When using `attrs`, the `__init__()` method can be auto-generated. Should we also auto-generate the constructor documentation or leave it up to the reader to figure it out? If we ever decide to sort attributes instead of sticking to the source order, the reader wouldn't be able to figure it out anymore from just the docs.

There are classes that offer factory methods and don't want the user to call the constructor directly. It would be nice if there was a way to mark the default constructor as private, so it is hidden under the "show private API" button.

A common pattern is the following:
```py
class C:
    def __init__(self, x: int):
        self.x = x
```
To properly document this, both the `x` constructor argument and the `x` instance variable need a docstring, even though the contents are the same. It would be nice if only one of them had to be provided and the other would automatically use the same docstring.

Note that in this exact case, `attrs` could be used to auto-generate the `__init__()` method, but that won't work if one argument is trivial like this example but another is non-trivial.

tristanlatr:
I like the way Javadoc presents the constructors (i.e https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/io/File.html#%3Cinit%3E(java.io.File,java.lang.String)), more generally I like that Javadoc separates Constructor / Instance variables / Class variables / Methods / Package / Module with blocks and titles. 

I think we could present the constructor information as it is used by people using the class i.e. not displaying `__init__(*args)` but `C(*args)`. We could show this before the summary table if the` __init__` method has a docstring, else we keep the `__init__` part of the summary and we don't show anything special. 

>  If we ever decide to sort attributes instead of sticking to the source order, the reader wouldn't be able to figure it out anymore from just the docs.

We should say when a class is `attrs` based, same goes for `zope.interface`. May be a footer note ?

> There are classes that offer factory methods and don't want the user to call the constructor directly. It would be nice if there was a way to mark the default constructor as private, so it is hidden under the "show private API" button.

I don't think that exists

[...]

PR #379, opened 2021-03-25
Sort class members by privacy class and then by name

ntamas:
This PR changes the class page template in the HTML output to sort class members by privacy class first and then by name (in alphabetical order). The proposed changes in this PR were necessary to build the documentation of `python-igraph` using PyDoctor (see the results [here](https://igraph.org/python/doc/api/index.html)) -- some of igraph's classes contain quite a few members and the original unsorted list was hard to use.


pydoctor/templatewriter/pages/__init__.py
```diff
@@ -279,6 +279,11 @@ def __init__(self, ob, docgetter=None):
                 self.baselists.append((baselist, attrs))
         self.overridenInCount = 0
 
+    def children(self):
+        return sorted((o for o in self.ob.contents.values()
+                       if o.name != '__init__' and o.isVisible),
```
tristanlatr, line 284: I think this will hide the `__init__` methods and `__init__` modules. As you can in the docs generated for this PR: [here](https://pydoctor--379.org.readthedocs.build/en/379/api/pydoctor.templatewriter.writer.TemplateWriter.html) for instance, the `__init__` method do not show up...

Let's note that we are discussing about how to display of constructors [here](https://github.com/twisted/pydoctor/issues/305). And I think the PR #360 will solve the issue with `__init__` module everywhere.

About the sorting order, @mthuurne will correct me if I'm wrong, but I think it's a design decision (probably discutable) to list members in the same order they have been defined in the code. So maybe this should be an option like `--sort-members` for instance. Also sorting by kind would be useful maybe?

tristanlatr (changes requested):
Hello @ntamas ,

Thanks a lot for opening the PRs. 

I'll do a first pass of reviews :)

pydoctor/templatewriter/pages/__init__.py
mthuurne: Once #360 lands, it will no longer be necessary to filter out `__init__` modules, since `__init__.py` will be treated as the source file for the package rather than a module within the package.

I don't know whether using the source order was a design decision or just a result of the most straightforward implementation. For most projects, source order is either pure chaos or some logic that only the developer understands, so sorting the names would benefit the reader, in my opinion.

I'll post a separate reply at the top level to discuss the sort order, since that's more of a requirements discussion than a code discussion.

mthuurne: Actually, this code is inside `ClassPage`, so it would only be filtering out `__init__()` methods. If there is a reason why you wouldn't want to see the `__init__()` methods listed, please document it in #305. But for the time being, we'd lose information by simply dropping them unconditionally for all projects.

mthuurne:
Regarding the order of members, source order is usually not the most intuitive to the reader, so I agree that we should pick a different sort order by default.

While I write "by default" here, I'd like to try and define a single fixed sort order that will be acceptable for all projects. If it turns out that is not possible, we could consider adding command line options to enable different sort orders. For the sake of maintainability, I'd rather not introduce more options unless they are really needed.

I think that we should apply sorting to modules as well, not only classes.

As @tristanlatr suggested, sorting by "kind" (class method, instance method, class variable, instance variable, property etc.) might be useful. In particular, grouping things that can be accessed on a class together and things that can be accessed on an instance. Maybe it would make more sense to split this into two separate tables though instead of using it as a sort order?

In the sidebar (#370) methods and attributes are separated and each of those categories is sorted alphabetically. Would it make sense to use the same grouping and sorting for the sidebar and the main content?

pydoctor/templatewriter/pages/__init__.py
ntamas: This method was blindly copied from `PackagePage` where `__init__` is also filtered, although I guess that `__init__` refers to `__init__.py` there, not the `__init__` method, right? If that is the case, then indeed it should not be filtered.

tristanlatr:
Hi,

I agree we should group the members by kind. 

I think the exhaustive list of "kind"s is something like this :  
- Class 
- Class Method
- Instance Method
- Class Variable
- Instance Variable
- Variable
- Property
- Function
- Module
- Package
- Interface ?
- Exception (TODO): Shows as Class
- Constant (TODO): Shows as Variable

Just for the record, Dash proposes filters for Twisted docs like follows: 

by default 2021-04-08 at 2 27 19 PM

As talked in #364, we should also switch to an Enum to handle the kind values, then we can safely refactor the Pages to show subsections with all kinds. 

For the short term maybe we could replace 

```
key=lambda o2:(-o2.privacyClass.value, o2.fullName()))
```

by 
```
key=lambda o2:(-o2.privacyClass.value, o2.kind, o2.fullName()))
```

And see if the output sorted by privacy, kind, then name is acceptable ?

ntamas:
I gave it a go with the igraph documentation and it seemed perfectly fine for me; I'll push a new commit here soon.

tristanlatr:
> it seemed perfectly fine for me

I'm unsure honestly, The `__init__` used to show up first if it was declared first. Now it's lost in the middle somewhere... 
I think the order of kins should be as follow: Packages, Modules, Classes, Class Methods,  Static Methods, (Instance) Methods, Class Variables, Instance Variables, Variables,  Functions. 


We could define something like :

```python

class KindClass(Enum):
    Package = 1000
    Module = 900
    Interface = 850
    Class = 800
    Class_Method = 700
    Static_Method = 600
    Instance_Method = 500
    Async_Class_Method = 480
    Async_Static_Method = 460
    Async_Instance_Method = 440
    Function = 400
    Async_Function = 350
    Class_Variable = 300
    Instance_Variable = 200
    Property = 150
    Variable = 100
```

I'm not sure if we need to use, orderedenums: https://docs.python.org/3/library/enum.html#orderedenum .

Then, a (huge) bit if refactoring to use the new `KindClass`, and it should do the trick. 

`kind` apprears 111 times in our codes, I don't think it's really hard to refactor. 

 It's starting to be unmaintainable anyway, so we really need this. I can open a PR for that. 

But maybe there is another solution ?

tristanlatr:
So once #382 is merged we can use:
```
key=lambda o2:(-o2.privacyClass.value, -o2.kind, o2.fullName()))
```
And Methods should show up firsts is classes. According to the Enum values.

Then, more work will be needed to group the tables by kind like in javadoc.

tristanlatr:
Hi @ntamas,

The PR that transform the kind values to en Enum is merged, so we'll be able to have sorting order that makes sens :)

Could you update your PR with master ?

Also could you apply changes such that the same order is used for the ModulePage and PackagePage ?

Thank you,

ntamas:
I think this is done. Let me know if I missed anything.

tristanlatr:
It looks like only the table rows are ordered the right way but the content items in the page (starting after the tables), do not use the same order. I think the `childlist` renderer should also be changed, I assume adding the key to the `sorted` function in `CommonPage.methods()` should do it. 

by default 2021-04-26 at 10 14 53 AM

ntamas:
It seems like `CommonPage.children()` does not sort by name, only by privacy class and kind. `ModulePage` inherits from `CommonPage` so it inherits this behaviour. If I change the implementation of `CommonPage.children()`, that could have consequences for other `CommonPage` subclasses -- OTOH, it might actually be desired even for other subclasses as well. What's your opinion, should I override `.children()` in `ModulePage` or can I fix it in `CommonPage`?

tristanlatr:
I think you can change the sorting order in the base class, so the children will also be sorted in the ClassPage and PackagePage.

ntamas:
Changed it in the base class now.

tristanlatr:
Can you also edit the `CommonPage.methods()` so the child list is also using the same order ?

thanks

ntamas:
It's done now. Sorry, I was searching for `sorted()` to identify places where I needed to intervene, this one has slipped through.

tristanlatr:
Can you allow me to push some commits to `igraph/pydoctor` repo ?

Thanks

ntamas:
I thought you are allowed to push commits to my PR but in case you aren't, I have sent an invite to the repo with write access.

tristanlatr (approved):

tristanlatr:
Thanks!

I think it's because you forked the project with your organization profile that you needed to give me access.

mthuurne (approved):
Looks good; thanks!

PR #380, opened 2021-03-25
Look up identifiers in parent classes as well when parsing Epydoc markup

ntamas:
This PR changes the Eypdoc markup parser to look up identifiers in parent classes as well when trying to resolve a link target in a class context. The proposed changes in this PR were necessary to build the documentation of `python-igraph` using PyDoctor (see the results [here](https://igraph.org/python/doc/api/index.html)).

I could not find any formal description of how the method lookup should proceed with the Epydoc markup; all that I know is that we've been using Epydoc in the past with `python-igraph` and references to methods in superclasses worked fine with Epydoc. We needed to switch to PyDoctor because we dropped support for Python 2.x and Epydoc could not parse Python 3.x markup any more. This PR essentially ensures that all the internal method links that used to work with Epydoc keep on working with PyDoctor as well.

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -184,6 +184,11 @@ class _EpydocLinker(DocstringLinker):
             target = src.resolveName(identifier)
             if target is not None:
                 return target
+            if isinstance(src, model.Class):
+                for base in src.allbases():
+                    target = base.resolveName(identifier)
+                    if target is not None:
+                        return target
             src = src.parent
 
         # Walk up the object tree again and see if 'identifier' refers to an


tristanlatr (commented):
Hi again,

This PR looks good. 

I'm not super familial with how the `EpydocLinker` works, so @mthuurne would probably be best suited to comment on the logic this addition. 

All I can say is that I'm happy if that fills a feature that Epydoc had and not PyDoctor, but it would be good to have a test tho. All `epydoc2stan` tests can be found here https://github.com/twisted/pydoctor/blob/master/pydoctor/test/test_epydoc2stan.py

pydoctor/epydoc2stan.py
```diff
@@ -184,6 +184,11 @@ def _resolve_identifier_xref(self,
             target = src.resolveName(identifier)
             if target is not None:
                 return target
+            if isinstance(src, model.Class):
+                for base in src.allbases():
```
mthuurne, line 188: Does the `Class.find()` method provide what you need? That would reduce the amount of new code considerably.

mthuurne (commented):
It would be good to have a unit test case for this, to make sure we don't accidentally break this functionality in the future.

pydoctor/epydoc2stan.py
ntamas: I don't know to be honest. I have replaced this with `target = src.find(identifier)` and it seems to do the trick (i.e. all references that were resolved with the previous version of the code in `python-igraph`'s documentation are also resolved with the new version), but I don't know what the difference is. if you think it's better to use `Class.find()`, I trust your assessment and I'll just use that here.

tristanlatr:
Hi @ntamas ,

Thanks for updating the PR.

As pointed out by @mthuurne , could you a test case that demonstrate this new feature of the `_EpydocLinker` ?

Thanks,

ntamas:
Yes, I will, I'm just swamped with other things at the moment. Give me a few days and I'll try to get back to this.

tristanlatr:
I can write the tests. 

@ntamas , can you provide an example of a case that was failing before ?

Thanks

ntamas:
Hi @tristanlatr 

Sorry, I don't have a clean reproduction on this yet; it looks like the situation is not that simple as I originally thought and the problem appears only when the class being documented has a base class that is in a C extension and you are linking to a method that is defined in the base class. It definitely happens in `python-igraph`'s documentation, where `igraph.Graph` is derived from `igraph._igraph.GraphBase` and the docstring of `igraph.Graph.indegree()` (implemented in Python) links to `igraph._igraph.GraphBase.degree()`(implemented in the C base class) using the `L{degree}` Epydoc syntax in the docstring. However, I haven't managed to create a small self-contained example yet. I wanted to use Cython to generate the code for the C module, but apparently Cython does something differently as I can't even make the Cython-defined functions appear in `pydoctor`'s output even with `--introspect-c-modules` turned on. In fact, the `pydoctor` throws quite a few exceptions with my test module due to things like `is_async` not being defined on function objects coming from the C module etc; I managed to work around these problems but the function is still not there in the docs.

I have placed an example project in my Dropbox [here](https://www.dropbox.com/s/bsvcjzeaft81ymh/pydoctor-c-ext-doc-test.zip?dl=0); usage instructions are in the README. I'm not sure whether this project reproduces the issue due to reasons outlined above, but it shows you how the stock PyDoctor version crashes with it.

tristanlatr:
Hi @ntamas ,

Thanks for complete investigation and report on the issue. I hope I can pick it up and add proper tests for it.

Regarding the code on the dropbox, I haven't tested it yet, but it looks like there is no README file. Maybe you forgot to upload it ?
We can think of a more complex test case that would need to compile a c-extension prior running the test. 

Regarding how pydoctor crashes with c extensions and `is_async`, do you have a more clear test case for how your [patch](https://github.com/igraph/igraph.org/blob/acdeb71a0294928c4eb432a44f83a9e8ba216bc1/_tools/pydoctor-21.2.0.patch#L84) fixes the issue ?

Thanks again

ntamas:
Sorry, I've updated the ZIP file, it should have a README now.

When you follow the instructions in the README, the build should first fail with the following exception:

```python
  File "/private/tmp/x/.venv/bin/pydoctor", line 8, in <module>
    sys.exit(main())
  File "/private/tmp/x/.venv/lib/python3.9/site-packages/pydoctor/driver.py", line 393, in main
    system.process()
...
  File "/private/tmp/x/.venv/lib/python3.9/site-packages/pydoctor/model.py", line 925, in getProcessedModule
    self.processModule(mod)
  File "/private/tmp/x/.venv/lib/python3.9/site-packages/pydoctor/model.py", line 945, in processModule
    self.unprocessed_modules.remove(mod)
KeyError: ZopeInterfaceModule 'mymodule.base_class'
```

I work around this with the following patch (although I have no idea why this is needed):

```patch
--- pydoctor/model.py.orig	2021-06-25 20:17:55.000000000 +0200
+++ pydoctor/model.py	2021-06-25 20:18:12.000000000 +0200
@@ -942,7 +942,8 @@
             mod.state = ProcessingState.PROCESSED
             head = self.processing_modules.pop()
             assert head == mod.fullName()
-        self.unprocessed_modules.remove(mod)
+        if mod in self.unprocessed_modules:
+            self.unprocessed_modules.remove(mod)
         self.progress(
             'process',
             self.module_count - len(self.unprocessed_modules),
```

After applying this patch, the build starts to work, but the introspection does not discover the `func()` function from `BaseClass` (which is the one that is implemented in the C extension), even though `func` is in `dir(BaseClass)`. This is because none of the `isinstance(v, types.SomethingType)`-based tests in `_introspectThing` recognize `func` as a function; my investigation shows that `inspect.isfunction(BaseClass.func)` is `False` but `inspect.isroutine(BaseClass.func)` is `True` so it seems like this is the only way to recognize an instance method defined by Cython. Then I extend `_introspectThing` to also test things with `inspect.isroutine()`:

```patch
--- pydoctor/model.py.orig	2021-06-25 20:17:55.000000000 +0200
+++ pydoctor/model.py	2021-06-25 20:35:40.000000000 +0200
@@ -808,10 +808,11 @@

     def _introspectThing(self, thing: object, parent: Documentable, parentMod: _ModuleT) -> None:
         for k, v in thing.__dict__.items():
-            if (isinstance(v, (types.BuiltinFunctionType, types.FunctionType))
+            if (isinstance(v, (types.BuiltinFunctionType, types.FunctionType, types.MethodDescriptorType, types.ClassMethodDescriptorType))
                     # In PyPy 7.3.1, functions from extensions are not
                     # instances of the above abstract types.
-                    or v.__class__.__name__ == 'builtin_function_or_method'):
+                    or v.__class__.__name__ == 'builtin_function_or_method'
+                    or inspect.isroutine(v)):
                 f = self.Function(self, k, parent)
                 f.parentMod = parentMod
                 f.docstring = v.__doc__
```

and this is the point where I get the exception related to a missing `is_async` on a `ZopeInterfaceFunction`:

```
  File "/private/tmp/x/.venv/lib/python3.9/site-packages/twisted/web/_flatten.py", line 309, in _flattenTree
    element = next(stack[-1])
  File "/private/tmp/x/.venv/lib/python3.9/site-packages/twisted/web/_flatten.py", line 227, in _flattenElement
    result = renderMethod(request, rootClone)
  File "/private/tmp/x/.venv/lib/python3.9/site-packages/pydoctor/templatewriter/pages/table.py", line 27, in kind
    if isinstance(child, Function) and child.is_async:
AttributeError: 'ZopeInterfaceFunction' object has no attribute 'is_async'
```

This is the exception that's solved by [my patch in the igraph repo](https://github.com/igraph/igraph.org/blob/acdeb71a0294928c4eb432a44f83a9e8ba216bc1/_tools/pydoctor-21.2.0.patch#L84) -- and this is the point where I'm stuck because after applying all these patches, I cannot create a situation similar to the one I observed in the `python-igraph` repo where a function defined on the base class and referred by a docstring in a derived class was not linked properly; here in this example it works, and I still don't know what the difference is between this small example and `python-igraph`.

ntamas:
Also note that the patches in my comments above are more like workarounds for the immediate problems that I've encountered so I could get the build to finish, but I don't claim that they are correct; in particular, I don't quite understand what `inspect.isroutine()` matches and whether it would also match things that we don't want to match during introspection.

wsanchez (changes requested):
This is adding new code paths, it really should come with tests…

tristanlatr:
This PR has been replaced with #495! Thanks @ntamas for taking the time to report all these issues related to c introspection and name resolving. 

I think `python-igraph` will be good to use the next `pydoctor` version without pathing it ;) Also see https://github.com/twisted/pydoctor/pull/494 which adds a test that run `python-igraph` docs with the latest pydoctor version and ensures `igraph._igraph.BaseGraph` pages effectively links to `igraph.Graph` and the opposite way arround too. 

Talk to you later,

ntamas:
Thanks for your work on this! So, shall we try PyDoctor 22.2.0 on `python-igraph` without any patching and see how it goes?

tristanlatr:
> So, shall we try PyDoctor 22.2.0 on python-igraph without any patching and see how it goes?

Yes, It think it will work.

ntamas:
Looks good to me as well!

@iosonofabio Can you also take a quick look at it in the `develop` branch to see if it's okay (in case I missed something). This is the diff I used to disable the patching:

```patch
diff --git a/scripts/mkdoc.sh b/scripts/mkdoc.sh
index f8b89b8..1bf29fd 100755
--- a/scripts/mkdoc.sh
+++ b/scripts/mkdoc.sh
@@ -46,11 +46,7 @@ if [ ! -d ".venv" ]; then
   python3 -m venv .venv

   # Install sphinx, matplotlib, wheel, and pydoctor into the venv
-  .venv/bin/pip install -U sphinx sphinxbootstrap4theme matplotlib wheel
-
-  echo "Patching PyDoctor..."
-  .venv/bin/pip install -U pydoctor
-  $SCRIPTS_FOLDER/patch-pydoctor.sh
+  .venv/bin/pip install -U pip sphinx sphinxbootstrap4theme matplotlib wheel "pydoctor>=22.2.0"
 fi

 #echo "Set PyDoctor theme"
```

PR #381, opened 2021-03-25
Search function with Lunr integration

tristanlatr:
This PR adds a simple search function that works locally with Javascript. 

Here is a test: https://pydoctor--381.org.readthedocs.build/en/381/api/search-results.html?search-query=test


tristanlatr:
Not sure why there is an encoding error on windows the CI.. :/

pydoctor/templatewriter/__init__.py
```diff
@@ -31,6 +32,15 @@ def parse_xml(text: str) -> minidom.Document:
     except Exception as e:
         raise ValueError(f"Failed to parse template as XML: {e}") from e
 
+def scandir(path: Path) -> Iterable['Template']:
```
mthuurne, line 35:
```suggestion
def scandir(path: Path) -> Iterator['Template']:
```

pydoctor/templatewriter/__init__.py
```diff
@@ -31,6 +32,15 @@ def parse_xml(text: str) -> minidom.Document:
...
+    """
+    Scan a directory for templates. 
+    """
+    for entry in os.scandir(path):
```
mthuurne, line 39: `path.iterdir()` should work as well, then you don't have to convert the strings back into a `Path`s.

tristanlatr: It's not string it's os.DirEntry, but it does not have the suffix property so we have to cast it to Path

pydoctor/templatewriter/__init__.py
```diff
@@ -84,7 +94,7 @@ def __init__(self, name: str, text: str):
         self.text = text
         """Template text: contents of the template file."""
 
-    TEMPLATE_FILES_SUFFIX = ('.html', '.css', '.js')
+    TEMPLATE_FILES_SUFFIX = ('.html', '.xml', '.xhtml', '.css', '.js', '.svg')
```
mthuurne, line 97: I'm curious: why are you adding XML and XHTML?

For XML, it seems strange to treat that as HTML later, since most XML formats are not HTML serializations.

For XHTML, the Twisted templating system only accepts XHTML as far as I know, so we could consider renaming the template files from `.html` to `.xhtml`.

tristanlatr: Yeah, I can roll back this changes, it was because all-document template was xhtml at the beginning but it could not get rendered by the browser with this file extension so everything uses .html. 

I'll roll back this

pydoctor/templatewriter/writer.py
```diff
@@ -56,13 +57,28 @@ def prepOutputDirectory(self) -> None:
         """
         Write static CSS and JS files to build directory.
         """
-        os.makedirs(self.base, exist_ok=True)
-        for template in self.template_lookup.templates:
-            if isinstance(template, _StaticTemplate):
-                with self.base.joinpath(template.name).open('w', encoding='utf-8') as jobj:
-                    jobj.write(template.text)
+        os.makedirs(self.output_dir, exist_ok=True)
```
mthuurne, line 60: [Path.mkdir()](https://docs.python.org/3/library/pathlib.html?highlight=pathlib#pathlib.Path.mkdir) can also create a directory including parents.

pydoctor/templatewriter/writer.py
```diff
@@ -56,13 +57,28 @@ def prepOutputDirectory(self) -> None:
...
+        self._writeStaticTemplates(self.template_lookup.templates)
+    
+    def _writeStaticTemplates(self, templates: Iterable[Template], subfolder: Optional[str] = None) -> None:
+        """
+        Write all L{_StaticTemplate} to output directory, inspect L{_TemplateSubFolder} 
+        and reccursively write the static templates in subfolders.
+        """
+        _template_rel_path_t = f"{subfolder + os.sep if subfolder else ''}%s"
```
mthuurne, line 68: This is risky: if the path contains percent characters, the interpolation could do the wrong thing.

Instead, I think you could take `self.output_dir` if `subfolder` is empty and look up the subdir otherwise. So operate on `Path` objects instead of on strings.

pydoctor/templatewriter/writer.py
```diff
@@ -56,13 +57,28 @@ def prepOutputDirectory(self) -> None:
...
+        for template in templates:
+            _template_rel_path = _template_rel_path_t % template.name
+            outfile = self.output_dir.joinpath(_template_rel_path)
+            if isinstance(template, _TemplateSubFolder):
+                os.makedirs(outfile, exist_ok=True)
+                self._writeStaticTemplates(template.lookup.templates, subfolder=_template_rel_path)
+                
+            elif isinstance(template, _StaticTemplate):
+                with outfile.open('w', encoding='utf-8') as fobj:
+                    fobj.write(template.text)
```
mthuurne, line 79: Would it make sense to have these two fragments in a method on `Template`, implemented in both subclasses instead?

tristanlatr: You mean for instance to add a `write(path: Path)` method to `Template` superclass ? 

Then the subclass would implement it.

mthuurne:
Would it be possible to split off the changes to the template system (cleanups, XML/XHTML/SVG support and subdirs) into a separate PR? Also some of the remarks I entered about replacing `os` functions with `pathlib` methods could be resolved before the actual search functionality is added.

pydoctor/templatewriter/__init__.py
mthuurne: Indeed it's a dir entry. But the only thing you use the dir entry for is to create the `Path` object, which you could get directly from `iterdir()` instead.

mthuurne:
Regarding SVG support, do we actually parse the SVG files or just copy them over as binary or text data? In the latter case, maybe it would be good to treat all unknown file extensions as binary blobs and copy them over as-is. That way a project that wants to use for example PNG images in their theme can do so as well.

tristanlatr:
Re-: SVG .. The thing is there is no "generic" support for any file extension, we ignore the files if it doesn't match the suffix. The names of the templates, includes the suffixes and are used as identifier for the template. So no, people cannot use PNG instead of SVG.

tristanlatr:
But it's copied just like other CSS and JS files are copied, yes.

mthuurne:
> Re-: SVG .. The thing is there is no "generic" support for any file extension, we ignore the files if it doesn't match the suffix. The names of the templates, includes the suffixes and are used as identifier for the template. So no, people cannot use PNG instead of SVG.

But does it ignore unknown file types because that's how it is currently implemented or because the design wouldn't work if those files are instead copied from the user template to the output as-is?

tristanlatr:
I think it would work but it's a bit dangerous to allow just any file to be passed to the output as is. Also The thing is the filenames are used as identifier , so there is no point supporting arbitrary file extension because they will never be referenced in the code. 

So all in all I don't see a great value into supporting unknown file types n template directory, because people can just add the files after to their output directory.

tristanlatr:
TODO: Split the PR in two  separate the cleanup and the template subfolder support first, then, mark this PR as ready for review

tristanlatr:
I've generated these API docs with the search feature: https://github.com/tristanlatr/apidocs

pydoctor/themes/base/search.js
```diff
@@ -0,0 +1,442 @@
...
+
+  let nb_functions = search_results_documents.filter(function(value){
+    return (value.querySelector('.kind').innerHTML.endsWith("Function") && (value.querySelector('.privacy').innerHTML.endsWith(priv ? "PRIVATE" : "VISIBLE")));
+  }).length;  
+
+  let nb_modules = search_results_documents.filter(function(value){
+    return (value.querySelector('.kind').innerHTML.endsWith("Module") || value.querySelector('.kind').innerHTML.endsWith("Package")) && (value.querySelector('.privacy').innerHTML.endsWith(priv ? "PRIVATE" : "VISIBLE"));
+  }).length;  
+  
+  // We don't go in the details of property/attribute/instance variable.
```
tristanlatr, line 123: Maybe it would be more consistent to go in the details.

pydoctor/themes/base/search.js
```diff
@@ -0,0 +1,442 @@
...
+    }
+    setInfos('')
+    setWarning('')
+    
+    console.log("Message received from worker: ")
+    console.dir(response.data)
+
+    if (!response.data.results){
+      setErrorStatus();
+      throw("No data received from worker")
```
tristanlatr, line 329:
```suggestion
      throw("No data received from worker");
```

pydoctor/themes/base/search.js
```diff
@@ -0,0 +1,442 @@
...
+  worker.onmessage = function (response) {
+
+    if (_setLongSearchInfosTimeout){
+      clearTimeout(_setLongSearchInfosTimeout)
+    }
+    setInfos('')
+    setWarning('')
+    
+    console.log("Message received from worker: ")
+    console.dir(response.data)
```
tristanlatr, line 325:
```suggestion
    console.dir(response.data);
```

pydoctor/themes/base/search.js
```diff
@@ -0,0 +1,442 @@
...
+  
+  worker.onmessage = function (response) {
+
+    if (_setLongSearchInfosTimeout){
+      clearTimeout(_setLongSearchInfosTimeout)
+    }
+    setInfos('')
+    setWarning('')
+    
+    console.log("Message received from worker: ")
```
tristanlatr, line 324:
```suggestion
    console.log("Message received from worker: ");
```

pydoctor/themes/base/search.js
```diff
@@ -0,0 +1,442 @@
...
+    query: _query,
+  });
+  
+  worker.onmessage = function (response) {
+
+    if (_setLongSearchInfosTimeout){
+      clearTimeout(_setLongSearchInfosTimeout)
+    }
+    setInfos('')
+    setWarning('')
```
tristanlatr, line 322:
```suggestion
    setWarning('');
```

pydoctor/themes/base/search.js
```diff
@@ -0,0 +1,442 @@
...
+  worker.postMessage({
+    query: _query,
+  });
+  
+  worker.onmessage = function (response) {
+
+    if (_setLongSearchInfosTimeout){
+      clearTimeout(_setLongSearchInfosTimeout)
+    }
+    setInfos('')
```
tristanlatr, line 321:
```suggestion
    setInfos('');
```

pydoctor/themes/base/search.js
```diff
@@ -0,0 +1,442 @@
...
+    if (response.data.results.length == 0){
+      setStatus('No results matches "' + _query + '"');
+      return ;
+    }
+    else{
+      setStatus('Fetching documents...');
+    }
+
+    // Get result data
+    httpGet("all-documents.html", function(response2) {
```
tristanlatr, line 341: Would be best to start loading "all-documents.html" right at the beginning of the search process. Not only once we receive the response from the Worker (see `worker.onmessage = function (response)`). This would cut a couple of seconds per search, which is not negligible.

PR #382, opened 2021-04-10
Use an Enum to store the possible kinds

tristanlatr:
We need this to be able to properly groups objects members by types in the HTML.

Related: https://github.com/twisted/pydoctor/issues/364 and https://github.com/twisted/pydoctor/pull/379#issuecomment-817157046


mthuurne (changes requested):
Thanks for the PR!

The main issue is how we want to separate model from presentation. I don't think we need to be super strict at this point in time, since there is a lot of existing code that mixes model and presentation. That said, I think the main motivation for using an enum is to get rid of presentation strings being used in analysis code, so it would be good to separate model from presentation where we can in the new code.

pydoctor/model.py
```diff
@@ -78,6 +78,34 @@ class PrivacyClass(Enum):
...
+    @property
+    def name(self) -> str: # type: ignore[override]
+        return super().name.replace("_", " ")
+
+class KindClass(_KindEnum):
+    """
+    Values indicating the possible object types.
+
+    @note: Some new kinds can be dynamically created by using `zope.schema`.
+        New kinds will have lower values.
```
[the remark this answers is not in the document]
tristanlatr: It’s really the same to me, we have to sort with the minus sign and it does the trick. This is how we sort by privacy

pydoctor/model.py
```diff
@@ -112,7 +153,7 @@ def css_class(self) -> str:
         """A short, lower case description for use as a CSS class in HTML."""
```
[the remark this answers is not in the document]
tristanlatr: This is outside the scope of this or I think

pydoctor/model.py
```diff
@@ -78,6 +78,34 @@ class PrivacyClass(Enum):
...
+    """
+    @property
+    def name(self) -> str: # type: ignore[override]
+        return super().name.replace("_", " ")
+
+class KindClass(_KindEnum):
+    """
+    Values indicating the possible object types.
+
+    @note: Some new kinds can be dynamically created by using `zope.schema`.
```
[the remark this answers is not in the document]
tristanlatr: I don’t see other way to deal with zope schema thing. We need to have dynamic kinds names. I don’t understand what you mean by being just a shortcut. In any case, this pattern is sketchy... and I don’t see how we could meet the requirements with static enum values.

pydoctor/model.py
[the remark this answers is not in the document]
mthuurne: We should probably just use `enum.auto()`, since these values are purely internal details that aren't exported or persisted.

Deriving presentation order from the enum values at all is also an example of putting presentation details in the model, but separating it would require extra code for little practical benefit, so I don't think that's necessary.

pydoctor/model.py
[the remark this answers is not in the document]
mthuurne: Yes, it's not necessary to address it now, but it's something you could keep in mind when designing new code, to make sure there is a way it can be fixed later.

pydoctor/model.py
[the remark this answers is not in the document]
mthuurne: I have never used zope schemas, only interfaces, so I'll have to dive into the zope docs to figure out what this is even for.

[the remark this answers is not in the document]
mthuurne: With "shortcut" I meant that when the original author decided to create new `kind` values dynamically, I wonder if that was a deliberate design decision or just the easiest way to implement it at the time and that initial implementation stuck around.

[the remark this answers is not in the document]
tristanlatr: Just look at our tests regarding zope.schema. No idea how we could reproduce this behaviour with a fix amount of enum values. 

https://github.com/twisted/pydoctor/blob/e32d0bac24d4c901454ec076becb031fa834bd15/pydoctor/test/test_zopeinterface.py#L152

pydoctor/zopeinterface.py
```diff
@@ -6,6 +6,17 @@
 
 from pydoctor import astbuilder, model
 
+def new_kind(name: str) -> model._KindEnum:
+    """
+    Dynamically create a new `_KindEnum` with `name` 
+    as it's only value and return the enum value. 
+    """
```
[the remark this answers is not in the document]
tristanlatr: How would you go to be able to create arbitrary named kind values then ?

pydoctor/model.py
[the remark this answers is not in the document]
tristanlatr: I guess the right way to do it would be to list all possible schemas here: https://zopeschema.readthedocs.io/en/latest/api.html#schema-apis

But that's definitely overkill in my opinion, I mean who is using this anyway ? 

I think we can go with a new kind value like `ZOPE.SCHEMA_FIELD`. Then we could rather store the type information inside the `parsed_type` attribute ? (It's currently `None`) It makes more sens to me anyway. 

What think ye ?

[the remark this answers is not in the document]
mthuurne: `zope.schema` looks similar to `attr.ib` from `attrs`. The schema class is more or less the type of the attribute, so I think `kind` wasn't really the best place to store the schema class to begin with.

Having a dedicated `SCHEMA_FIELD` kind would be an option indeed. Or maybe we could use the existing `ATTRIBUTE` kind.

I'm not sure what the criteria are for defining "kinds". An attribute is an instance variable, but we have it as a separate kind and I'm not sure why. Similarly, there is #364 which asks for a separate kind for exceptions.

There is some connection between the `Documentable` subclass and the `kind`, in that for example a documentable of kind `METHOD` is always an instance of `Function`. But there can be more than one kind of each subclass: for `Function` there is also `CLASS_METHOD`, `STATIC_METHOD` and plain `FUNCTION`. And in some cases it's not even a 1:1 mapping, for example a property is implemented using 1 to 3 methods (get/set/del), which isn't handled all that well at the moment.

I don't think the `kind` is used much for analysis, it seems to be mainly a presentation thing. So perhaps it doesn't even belong in the model. So we might want to overhaul the design in a more thorough way, but I think that it's best to first finish the conversion from string to enum and merge this PR before we start making bigger changes.

To come back to the current issue: I don't think we format schema fields in any particular way, except that we display the `kind` field itself. I don't know of any project that uses `zope.schema` (Twisted doesn't use it), so I wouldn't know what works better in practice. So I'll leave it up to you to pick between `SCHEMA_FIELD` and `ATTRIBUTE`, whichever is easier. If a user disagrees they can file a ticket and we make it nicer then; we'll then have someone that can show us real use cases.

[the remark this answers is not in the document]
tristanlatr: Ok I've pushed what I think is the right thing to do: Using AnnotationDocstring to link to zope.schema type in the apidocs.

pydoctor/model.py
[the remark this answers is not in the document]
tristanlatr: I've moved it to the utils module

tristanlatr:
It ready for another review @mthuurne :)

pydoctor/epydoc2stan.py
```diff
@@ -722,27 +722,28 @@ def format_summary(obj: model.Documentable) -> Tag:
...
         assert kind is not None # if kind is None, object is invisible
         tag(
-            "No ", kind.lower(), " docstring; ",
+            "No ", format_kind(kind), " docstring; ",
             ', '.join(
-                f"{subdocstrings[k]}/{subcounts[k]} "
-                f"{plurals.get(k, k + 's')}"
-                for k in sorted(subcounts)
+                f"{sub_objects_with_docstring_count[k]}/{sub_objects_total_count[k]} "
+                f"{format_kind(k, plural=sub_objects_with_docstring_count[k]>=2).lower()}"
```
mthuurne, line 744: Maybe `format_kind()` should return the description in lower case and we can call `.title()` when we want title case in the output.

pydoctor/model.py
```diff
@@ -77,7 +77,27 @@ class PrivacyClass(Enum):
     HIDDEN = 0
     PRIVATE = 1
     VISIBLE = 2
+    
+class KindClass(Enum):
+    """
+    L{Enum} containing values indicating the possible object types.
 
+    @note: Presentation order is derived from the enum values
```
mthuurne, line 85: I still think it would be simpler to use `auto()` and remove the `-` from the sorting.

pydoctor/model.py
```diff
@@ -77,7 +77,27 @@ class PrivacyClass(Enum):
     HIDDEN = 0
     PRIVATE = 1
     VISIBLE = 2
+    
+class KindClass(Enum):
```
mthuurne, line 81: `KindClass` is more or less saying the same thing twice. Maybe name it just `Kind` or `DocumentableKind`?

pydoctor/templatewriter/summary.py
```diff
@@ -183,7 +183,7 @@ def link(obj: model.Documentable) -> Tag:
             # category (class, method, etc.) an object belongs to.
             tag: Tag = tags.code(
                 epydoc2stan.taglink(obj, NameIndexPage.filename),
-                **{"data-type": obj.kind}
+                **{"data-type": epydoc2stan.format_kind(obj.kind) if obj.kind else ''}
```
mthuurne, line 186: Can the `kind` be `None` at this stage? If not, you can assert that it is not `None`. If it can be `None`, maybe it's better to omit the `data-type` attribute than to add it with an empty value.

pydoctor/templatewriter/summary.py
```diff
@@ -273,7 +273,7 @@ def roots(self, request, tag):
     def rootkind(self, request, tag):
         rootkinds = {}
         for o in self.system.rootobjects:
-            rootkinds[o.kind.lower() + 's']  = 1
+            rootkinds[epydoc2stan.format_kind(o.kind, plural=True).lower()] = 1
         return tag.clear()('/'.join(sorted(rootkinds)))
```
mthuurne, line 277: I think this is ancient code from a time before Python supported generator comprehensions or even sets. In modern Python it can be written more elegantly like this:
```py
         return tag.clear()('/'.join(sorted(
             epydoc2stan.format_kind(o.kind, plural=True).lower()
             for o in self.system.rootobjects
             )))
```

pydoctor/test/test_astbuilder.py
```diff
@@ -452,7 +452,7 @@ class Processor:
     f = P.contents['clientFactory']
     assert unwrap(f.parsed_docstring) == """Callable that returns a client."""
     assert f.privacyClass is model.PrivacyClass.VISIBLE
-    assert f.kind == 'Instance Variable'
+    assert f.kind == model.KindClass.INSTANCE_VARIABLE
```
mthuurne, line 455: You can use `is` instead of `==` here as well. Performance is not that important for the tests, but it's good for consistency.

pydoctor/test/test_zopeinterface.py
```diff
@@ -160,13 +161,16 @@ class IMyInterface(interface.Interface):
     mod = fromText(src, modname='mod', systemcls=ZopeInterfaceSystem)
     text = mod.contents['IMyInterface'].contents['text']
     assert text.docstring == 'fun in a bun'
-    assert text.kind == "TextLine"
+    assert str(text.parsed_type.to_stan(NotFoundLinker())) == "Tag('code', children=[Tag('', children=['schema.TextLine'])])"
```
mthuurne, line 164: In other tests we use the following approach:
```py
assert type2html(c) == '<code>str</code>'
```
I think that is a bit more robust against changes in the HTML generation process, as well as being easier to read.

pydoctor/zopeinterface.py
```diff
@@ -221,13 +220,19 @@ def _handleAssignmentInClass(self,
                     section='zopeinterface')
         else:
             match = schema_prog.match(funcName)
+
             if match:
```
mthuurne, line 224:
```suggestion
            if schema_prog.match(funcName):
```
Since we no longer use the match object anywhere else, we don't need to save it to a local variable either.

pydoctor/zopeinterface.py
```diff
@@ -221,13 +220,19 @@ def _handleAssignmentInClass(self,
...
+                attr.kind = model.KindClass.SCHEMA_FIELD
+
             else:
                 cls = self.builder.system.objForFullName(funcName)
                 if not (isinstance(cls, ZopeInterfaceClass) and cls.isschemafield):
                     return
-                attr.kind = cls.name
+                attr.kind = model.KindClass.SCHEMA_FIELD
+
+            # Linking to zope.schema.* class if setup with intersphinx
```
mthuurne, line 233:
```suggestion
            # Link to zope.schema.* class, if setup with intersphinx.
```
It's nice that we get slightly richer output now, by linking to the zope docs.

mthuurne (changes requested):
Still some things that can be improved, but it's all details now.

docs/source/conf.py
```diff
@@ -96,6 +96,7 @@
     '--intersphinx=https://urllib3.readthedocs.io/en/latest/objects.inv',
     '--intersphinx=https://requests.readthedocs.io/en/latest/objects.inv',
     '--intersphinx=https://www.attrs.org/en/stable/objects.inv',
+    '--intersphinx=http://code.nabla.net/doc/docutils/objects.inv',
```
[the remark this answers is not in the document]
tristanlatr: Ok I’ll just put my own docutils apidocs site...

Here : 
https://tristanlatr.github.io/apidocs/docutils/docutils.html

Docutils still misses apidocs, crazy.

pydoctor/epydoc2stan.py
tristanlatr: I think the opposite actually, lower() function drops casing information, where title() adds cases where it thinks it the best. What is there a adverb I the kind and we don’t want to upper case it ?

pydoctor/model.py
tristanlatr: I think auto() makes it more complicated to understand . With numbers it’s clear which will pop up first. Also it keeps it consistent with other enigma

pydoctor/test/test_astbuilder.py
tristanlatr: What should I understand  here ? Please don’t tell me that I should replace All double equals with is when possible ...

mthuurne: You could do an automated search-and-replace to replace the obvious ones; if there are any left after that, it's not really a problem.

pydoctor/model.py
mthuurne: The other enums were added when using the `enum34` compatiblity package when we were still supporting Python 2.7. If I added them today, I'd probably use `auto()`.

pydoctor/epydoc2stan.py
mthuurne: That's true. While we currently don't have such word like "of" in the descriptions, it's likely we'll get more kinds in the future and they might have. So it's fine to keep it like this.

docs/source/conf.py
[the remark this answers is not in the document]
mthuurne: pydoctor didn't have published API docs either until recently ;)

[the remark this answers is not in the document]
tristanlatr: Yes, but pydoctor is not a library

pydoctor/templatewriter/summary.py
tristanlatr: Ok, in any case the doc2dash parser needs an update and will use intersphinx so this doesn't really matter anymore.

pydoctor/epydoc2stan.py
```diff
@@ -722,27 +722,28 @@ def format_summary(obj: model.Documentable) -> Tag:
...
     tag: Tag = tags.span(class_='undocumented')
-    if subdocstrings:
-        plurals = {'class': 'classes', 'property': 'properties'}
+    if sub_objects_with_docstring_count:
+        
         kind = obj.kind
         assert kind is not None # if kind is None, object is invisible
         tag(
-            "No ", kind.lower(), " docstring; ",
+            "No ", format_kind(kind), " docstring; ",
```
mthuurne, line 741:
```suggestion
            "No ", format_kind(kind).lower(), " docstring; ",
```

mthuurne (approved):
There is one case where the original `.lower()` call wasn't preserved. I'll re-add that and then merge.

Issue #377, opened 2021-03-23
Representation of constants values


tristanlatr:
Having the value of the constant defined here and there would be very useful.

Epydoc had a great thing: the `epydoc.markup.pyval_repr. PyvalColorizer`, and it happens that some people already made the code work with Python3: https://github.com/nltk/epydoc/blob/master/src/epydoc/test/py3/pyval_repr.doctest

What we need to figure is how to know when displaying the value? I'm unsure of how Epydoc did this. `typing.Final` is an idea, whether the variable name is upper cased is another clue to indicate that the variable is final. Or we could just always try to render the value. How difficult is it to construct initial the value from AST when it's a single assignment ? It should not be complicated, or at least we should not handle that complexity!

Related: https://github.com/twisted/pydoctor/issues/364#issuecomment-803660886

mthuurne:
> What we need to figure is how to know when displaying the value? I'm unsure of how Epydoc did this. `typing.Final` is an idea, whether the variable name is upper cased is another clue to indicate that the variable is final. Or we could just always try to render the value.

I think we should only display values for constants, not variables that might change their value at runtime.

> How difficult is it to construct initial the value from AST when it's a single assignment ? It should not be complicated, or at least we should not handle that complexity!

Let's say we have these definitions:
```py
SPECIES_PIG = 1 << 11
SPECIES_FROG = 1 << 10

COLOR_RED = 1 << 2
COLOR_GREEN = 1 << 1
COLOR_BLUE = 1 << 0

KERMIT_LIKE = SPECIES_FROG | COLOR_GREEN
"""It's not easy."""
```

We could use `ast.literal_eval()` to convert the AST to a Python value. In the example above, that would result in `KERMIT_LIKE = 1026`, which is not very informative.

We could use the `astor` library to get a string representation of the AST value. This would be simple to implement, but it would limit our ability to colorize and link the output. In the example above, `astor` would produce `KERMIT_LIKE = SPECIES_FROG | COLOR_GREEN` as flat text.

Probably the best user experience would be if we'd implement a custom presenter as an AST visitor, similar to `pydoctor.epydoc2stan._AnnotationFormatter`. We could then hyperlink `SPECIES_FROG` and `COLOR_GREEN` to their respective documentation locations. And other formatting ideas could be taken from the `PyvalColorizer` that you mentioned.

[...]

PR #386, opened 2021-04-24
Docutils move

tristanlatr:
The goal of this PR is to integrate more closely with `docutils` and be able to generate docutils documents from Epytext parsed Elements. This way, we don't have to maintain an epytext and a restructuredtext version of every new features that needs a complete control over the document tree (like TOC). 

Parts of the short-terms benefits are: 

- Allows to generate the summary docstrings as described in #86 
- We can look at merging #370 and #402. The TOC generation and the pyval_repr depends on the new `node2stan` module.
-  The new `gettext()` function, will help us sanitize the docstrings to be indexed by the search system (#381, it's currently indexed with the raw docstring) 

What does it changes ?

- It adds a new method to the `ParsedDocstring` abstract class: `to_node()`. 
- `ParsedEpytextDocstring` has been modified to output a docutils document,just like `ParsedRstDocstring`. 

- Part of the `epydoc.markup.restructuredtext` module have been moved to a new module named `pydoctor.node2stan`. This module currently offers 3 functions: 
    - `node2stan(node: Union[Node, List[Node], docstring_linker: 'DocstringLinker') -> Tag` 
    - `node2html(node: Node) -> List[str]`
    - `gettext(node: Union[Node, List[Node]]) -> List[str]`

- All the tests have been adapted, some have been added

- But 3 are marked as `xfail` because `AnnotationDocstring` is not implementing `to_node()` and generates Tags directly.
I don't thinks this is a big issue (it will be fixed when #377 is fixed - in the PR #402)


tristanlatr:
I still need to add a test for the fall back to plaintext when using epytext

tristanlatr:
Once #389 and #397 are merged, I'll mark this PR as ready for review.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -131,11 +131,18 @@
...
 #   5. testing
 
-from typing import Any, List, Optional, Sequence, Union, cast, overload
+from typing import Any, Iterable, List, Optional, Sequence, Union, cast, overload
 import re
 
-from twisted.web.template import CharRef, Tag, tags
-from pydoctor.epydoc.doctest import colorize_doctest
+from docutils import utils
+from docutils.nodes import (Node, Text, document, inline, paragraph, literal, reference,
```
tristanlatr, line 138: We should probably use `from docutils import nodes` instead, this would avoid a lot of risk to create bugs related to variable names overrides.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -341,12 +344,12 @@ def parse(text: str, errors: Optional[List[ParseError]] = None) -> Optional[Elem
...
-        if raise_on_error:
-            raise errors[0]
-        else:
-            return None
+    # If there was a fatal error, then raise exception!
+    # The exception was not raised by default leading into displaying literally nothing
+    # when an invalid epytext docstring was used. Now it will fall back to plaintext thanks to this exception.
+    for e in errors:
+        if e.is_fatal():
+            raise e
```
tristanlatr, line 353: This change is concurrent to the changes proposed in https://github.com/twisted/pydoctor/pull/407

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1356,107 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
+            self._document.children.extend(set_nodes_parent(children[0].children, self._document))
+        
+        return self._document
+    
+    def _to_node(self,
             tree: Union[Element, str],
-            linker: DocstringLinker,
-            seclevel: int = 0
-            ) -> Any:
+            seclevel: int = 0,
```
tristanlatr, line 1380: The seclevel argument could be removed from the function.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1356,107 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
-            return value
+            yield set_node_attributes(Text(value), document=self._document)
+
         elif tree.tag == 'italic':
-            return tags.i(*variables)
+            yield set_node_attributes(emphasis('', ''), document=self._document, children=variables)
         elif tree.tag == 'math':
-            return tags.i(*variables, class_='math')
+            node = set_node_attributes(math('', ''), document=self._document, children=variables)
+            node.set_class('math')
```
tristanlatr, line 1437: This triggers a deprecation warning: 

`DeprecationWarning: docutils.nodes.Element.set_class deprecated; append to Element['classes'] list attribute directly`

pydoctor/epydoc/markup/plaintext.py
```diff
@@ -34,10 +35,20 @@ class ParsedPlaintextDocstring(ParsedDocstring):
...
     def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
         return tags.p(self._text, class_='pre')
+    
+    def to_node(self) -> nodes.document:
+        if self._document is not None:
+            return self._document
+        else:
+            self._document = utils.new_document('plaintext')
+            self._document.children.append(nodes.literal_block(rawsource=self._text, text=self._text))
+            return self._document
```
tristanlatr, line 54: This code is currently unused, maybe it should raise NotImplementedError() instead ?

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -41,29 +41,25 @@
...
+from typing import Iterable, List, Optional, Sequence, Set
 
 from docutils.core import publish_string
 from docutils.writers import Writer
-from docutils.writers.html4css1 import HTMLTranslator, Writer as HTMLWriter
 from docutils.readers.standalone import Reader as StandaloneReader
 from docutils.utils import Reporter, new_document
-from docutils.nodes import Node, NodeVisitor, SkipNode, Text
-from docutils.frontend import OptionParser
+from docutils.nodes import Node, NodeVisitor, Text, document
```
tristanlatr, line 50: We should use `from docutils import nodes`

pydoctor/node2stan.py
```diff
@@ -0,0 +1,241 @@
...
+
+from pydoctor.epydoc.markup import (
+    DocstringLinker, flatten, html2stan
+)
+from pydoctor.epydoc.doctest import colorize_codeblock, colorize_doctest
+
+def _node2html(node: Node, docstring_linker: 'DocstringLinker') -> List[str]:
+    visitor = _PydoctorHTMLTranslator(node.document, docstring_linker)
+    node.walkabout(visitor)
+    return visitor.body
```
tristanlatr, line 23:
```suggestion
def node2html(node: Node, docstring_linker: 'DocstringLinker') -> List[str]:
    """
    Convert a L{docutils.nodes.Node} object to HTML strings.
    """
    visitor = _PydoctorHTMLTranslator(node.document, docstring_linker)
    node.walkabout(visitor)
    return visitor.body
```

pydoctor/node2stan.py
```diff
@@ -0,0 +1,241 @@
...
+def node2stan(node: Union[Node, Iterable[Node]], docstring_linker: 'DocstringLinker') -> Tag:
+    """
+    Convert L{docutils.nodes.Node} objects to a Stan tree.
+
+    @param node: An docutils document.
+    @return: The element as a stan tree.
+    """
+    html = []
+    if isinstance(node, Node):
+        html += _node2html(node, docstring_linker)
```
tristanlatr, line 34:
```suggestion
        html += node2html(node, docstring_linker)
```

pydoctor/node2stan.py
```diff
@@ -0,0 +1,241 @@
...
+    else:
+        for child in node:
+            html += _node2html(child, docstring_linker)
```
tristanlatr, line 37:
```suggestion
            html += node2html(child, docstring_linker)
```

pydoctor/test/test_epydoc2stan.py
```diff
@@ -54,15 +58,15 @@ def docstring2html(obj: model.Documentable) -> str:
 
 def summary2html(obj: model.Documentable) -> str:
     stan = epydoc2stan.format_summary(obj)
-    assert stan.tagName == 'span', stan
+    assert stan.tagName == 'span' or stan.tagName == '', stan
```
tristanlatr, line 61:
```suggestion
    assert stan.tagName == '', stan
```

tristanlatr (commented):
Few adjustments to do still.

pydoctor/epydoc/markup/epytext.py
tristanlatr: Done

pydoctor/epydoc/markup/epytext.py
tristanlatr: I should probably revert this changes and propose another PR for that or add a commit to #407. 
Please advise.

pydoctor/epydoc/markup/epytext.py
tristanlatr: Done

pydoctor/epydoc/markup/epytext.py
tristanlatr: Replaced with `node['classes'].append('math')`

pydoctor/epydoc/markup/restructuredtext.py
tristanlatr: Done

tristanlatr:
The issue with mypy is that the "type: ignore" comment is actually needed for python 3.9 but not for python 3.8 or before. 

`pydoctor/epydoc2stan.py:854: error: unused "type: ignore" comment`

pydoctor/epydoc2stan.py
```diff
@@ -850,8 +849,9 @@ def visit_Subscript(self, node: ast.Subscript) -> Tag:
         sub: ast.AST = node.slice
         if isinstance(sub, ast.Index):
             # In Python < 3.9, non-slices are always wrapped in an Index node.
-            if sys.version_info < (3, 9):
-                sub = sub.value
+            # If tox runs with python 3.9, mypy will get a warning. 
+            # But we can't reach this code with python 3.9 anyway.
+            sub = sub.value # type: ignore[attr-defined]
```
mthuurne, line 854: mypy is supposed to [understand `sys.version_info`](https://mypy.readthedocs.io/en/stable/common_issues.html#version-and-platform-checks), so in theory this shouldn't be a problem.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1352,101 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
+            
+            # Figure the line number to warn on precise lines. 
+            # This is needed only for links currently.
+            lineno: Optional[int] = None
+            elem = tree.children[1]
+            if isinstance(elem, Element):
+                lineno = int(elem.attribs['lineno'])
+
+            yield set_node_attributes(nodes.title_reference(
+                   '', '', **args), document=self._document, lineno=lineno, children=_label.children)
```
mthuurne, line 1414: Unlike the old code, we no longer call `link_xref()` here. This means that linking is now done when converting the docutils document to Stan. Is that the best time to resolve crosslinks or would it be better to do it here and change the interface of `DocstringLinker` to use docutils types instead?

pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -558,9 +386,9 @@ class PythonCodeDirective(Directive):
 
     has_content = True
 
-    def run(self) -> List[Node]:
+    def run(self) -> List[nodes.Node]:
         text = '\n'.join(self.content)
-        node = docutils.nodes.doctest_block(text, text, codeblock=True)
+        node = nodes.doctest_block(text, text, codeblock=True)
         return [ node ]
```
mthuurne, line 392: It would be good to include these `nodes.` changes in the cleanup PR as well.

mthuurne (commented):
I've done a first review. Overall the approach looks good.

It would be nice to have some of the changes separated into a cleanup PR that can be merged before the actual docutils move. That should decrease the size of the diff considerably and will make future reviews easier.

We might want to consider were exactly the docutils/Stan boundary should be. If it's feasible, I think it would be better to move all uses of Stan out of `pydoctor.epydoc`: docutils is our document format while Stan is used to convert that document to HTML. This will allow other ways of outputting the document in the future, for example passing the document to Sphinx or supporting other output formats than HTML.

pydoctor/epydoc/markup/epytext.py
mthuurne: One advantage of resolving early would be that we could skip HTML generation entirely if we want to implement a "lint mode" that checks docstrings without producing output.

pydoctor/epydoc/docutils.py
```diff
@@ -0,0 +1,33 @@
+"""
+Collection of helper functions related to the creation L{docutils} nodes.
+"""
+from typing import Iterable, Iterator, Optional, List
+
+from docutils import nodes
+
+def set_nodes_parent(nodes: Iterable[nodes.Node], parent: nodes.Node) -> Iterator[None]:
```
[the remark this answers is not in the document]
tristanlatr: LOL you are right

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1352,101 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
-                return tags.p(*variables)
+            # we yield a paragraph node even if tree.attribs.get('inline') is True because
+            # the choice to render the <p> tags is handled in _PydoctorHTMLTranslator.should_be_compact_paragraph(), not here anymore
+            yield set_node_attributes(nodes.paragraph('', ''), document=self._document, children=variables)
         elif tree.tag == 'code':
-            return tags.code(*variables)
+            yield set_node_attributes(nodes.literal('', ''), document=self._document, children=variables)
         elif tree.tag == 'uri':
-            return tags.a(variables[0], href=variables[1], target='_top')
+            _label, _target = variables[0], variables[1]
```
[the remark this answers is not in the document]
tristanlatr: doing the unpacking is triggering some error, the variables can be any length, bu we consider only the two first elements.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1352,101 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
+
+            args = {}
+            if _target.astext() != _label.astext():
+                args['refuri']=_target.astext()
+            
+            # Figure the line number to warn on precise lines. 
+            # This is needed only for links currently.
+            lineno: Optional[int] = None
+            elem = tree.children[1]
+            if isinstance(elem, Element):
```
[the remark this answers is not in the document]
tristanlatr: IDK, it's safer ?

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1352,101 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
+                lineno = int(elem.attribs['lineno'])
+
+            yield set_node_attributes(nodes.title_reference(
+                   '', '', **args), document=self._document, lineno=lineno, children=_label.children)
+
+        elif tree.tag in ('name',):
```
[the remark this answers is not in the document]
tristanlatr: Thanks, that a left over of the older code.

pydoctor/epydoc/markup/epytext.py
tristanlatr: Honestly, I hope we can merge this Pr without breaking it in several smaller pieces. 
I'm OK to remove some of the changes like this one, but removing all the cleanups regarding the imports and put them in a separate PR would be extra work for nothin gin my opinion. 

The diff is under 1000 lines, that is what the Twisted changes process advices. please don't ask me to change everything in this PR.

pydoctor/epydoc2stan.py
tristanlatr: Well.. I get a warning on python 3.9...

pydoctor/epydoc/markup/plaintext.py
tristanlatr: I'm not sure what you mean, the `to_stan()` default implementation raises NotImplementedError, so all subclasses of ParsedDocstring MUST override `to_stan()`.

I agree with trying to keep markup modules deal only with nodes tree, but it's not the way it has been working for years. Please be more specific regarding what changes you what to see in this PR, what changes you want to see in another, future, PR.

pydoctor/epydoc/markup/epytext.py
tristanlatr: >  Is that the best time to resolve crosslinks or would it be better to do it here and change the interface of DocstringLinker to use docutils types instead?

This is probably something that could be left for a next PR. Also it has been working this way forever in the the `restructuredtext.py` module. So this change of behaviour makes us more consistent. 

If the goal is only to provide a "lint mode", then we could probably do it with a little extra code, that do not required us to touch this "sensible" part of the `epytex.py` module. But instead process the `ParsedDocstring` of all documentables in the systems, and report errors appropriately.

pydoctor/epydoc/markup/restructuredtext.py
tristanlatr: I assume your talking about the changes made in the following commits:

Use "from docutils import nodes" in node2stan.py
b0968b0

Use "from docutils import nodes" in restructuredtext.py module, too.
a1ec921

Do the same for epytext.py
d4a4df3

Cleanups
3ef32d2

Sorry, I'm not willing to apply this changes on a new clean branch from master, this would be a headache to sort out because this changes almost uniquely applies to lines that needs also changing in this PR.

pydoctor/epydoc2stan.py
tristanlatr: So you'r saying that we should add an `assert sys.version_info < (3, 9)` check ?

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1352,101 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
+        if self._tree is not None:
+            children = list(self._to_node(self._tree))
+            assert len(children)==1
+            # The contents is encapsulated inside a section node. 
+            # Reparent the contents of the second level to the root level. 
+            self._document.children.extend(set_nodes_parent(children[0].children, self._document))
+        
+        return self._document
+    
+    def _to_node(self, tree: Union[Element, str]) -> Iterable[nodes.Node]:
```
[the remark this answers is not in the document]
tristanlatr: I see what you mean, but it looks like there are subtle differences. 

The following diff represents what I've tried, but it makes 33 test fails with assertions errors. I'm sure it's worth dedicating a lot of time trying to make this work since we have sufficient testing to stay with the current implementation in my opinion. 

```diff
diff --git a/pydoctor/epydoc/markup/epytext.py b/pydoctor/epydoc/markup/epytext.py
index 224219c0..a1434e7a 100644
--- a/pydoctor/epydoc/markup/epytext.py
+++ b/pydoctor/epydoc/markup/epytext.py
@@ -1371,16 +1371,15 @@ class ParsedEpytextDocstring(ParsedDocstring):
         
         return self._document
     
-    def _to_node(self, tree: Union[Element, str]) -> Iterable[nodes.Node]:
-        
-        if isinstance(tree, str):
-            yield set_node_attributes(nodes.Text(tree), document=self._document)
-            return
+    def _to_node(self, tree: Element) -> Iterable[nodes.Node]:
 
         # Process the children first.
         variables: List[nodes.Node] = []
-        for c in tree.children:
-            variables.extend(self._to_node(c))
+        for child in tree.children:
+            if isinstance(child, str):
+                variables.append(set_node_attributes(nodes.Text(tree), document=self._document))
+            else:
+                variables.extend(self._to_node(child))
 
         # Perform the approriate action for the DOM tree type.
         if tree.tag == 'para':
```

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
tristanlatr: OK never mind, it passes the tests with `_label, _target = variables`.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1352,101 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
         if self._tree is None:
             self._stan = Tag('')
```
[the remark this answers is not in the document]
tristanlatr: I see, but should the whole functionality move be done on this PR ?

I can work on having a single `to_stan` implementation, in `ParsedDocstring`, that is going to call `node2stan`, that is already in another module than `epydoc`.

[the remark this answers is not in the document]
tristanlatr: I have something ready that I could push to this branch. 
It's not too big of a change, only thing , I needed to move `html2stan` and `flatten` function to another, new, module - `pydoctor.stanutils` - such that we avoid a circular import issue.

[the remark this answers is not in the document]
tristanlatr: I've pushed bd25740afd00f9a35c3976e348f9589c916ae641. 

Please tell me if that's the kind of thing you would expect. I can always revert it.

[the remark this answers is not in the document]
tristanlatr: I've made the ParsedDocstring provide a default implementation for the to_stan() method, Only ParsedPlaintextDocstring must override this method, see bd25740afd00f9a35c3976e348f9589c916ae641 . 

I think it's a good step in the right direction, but the `markup/__init__.py` is still pretty couple with stan related stuff. But it should taken care of later I think.

pydoctor/epydoc/markup/epytext.py
tristanlatr: I've reverted these changes.

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1352,101 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
+            return self._document
+        
+        self._document = utils.new_document('epytext')
+        
+        if self._tree is not None:
+            children = list(self._to_node(self._tree))
+            assert len(children)==1
+            # The contents is encapsulated inside a section node. 
+            # Reparent the contents of the second level to the root level. 
+            self._document.children.extend(set_nodes_parent(children[0].children, self._document))
```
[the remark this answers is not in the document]
tristanlatr: Thanks!

pydoctor/epydoc/markup/plaintext.py
tristanlatr: OK. 

I've implemented a middle ground approach where `ParsedDocstring` provides a default implementation for `to_stan()`, this implementation is really simple and relies on code in `node2stan` module.

pydoctor/test/epydoc/test_epytext2html.py
```diff
@@ -152,15 +142,38 @@ def test_epytext_sections() -> None:
...
+           <h2 class="heading">
+            Section 2
+           </h2>
+           <p>
+            This is a paragraph in section 2.
+           </p>
+          </div>
+
         '''
-    assert epytext2html(doc) == squash(expected)
```
tristanlatr, line 163: I'll resurrect this `squash` function since the same code is called again and again...

pydoctor/epydoc/markup/epytext.py
```diff
@@ -1352,69 +1352,101 @@ def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
...
 
         # Perform the approriate action for the DOM tree type.
         if tree.tag == 'para':
-            if tree.attribs.get('inline'):
-                return variables
-            else:
-                return tags.p(*variables)
+            # we yield a paragraph node even if tree.attribs.get('inline') is True because
+            # the choice to render the <p> tags is handled in _PydoctorHTMLTranslator.should_be_compact_paragraph(), not here anymore
+            yield set_node_attributes(nodes.paragraph('', ''), document=self._document, children=variables)
```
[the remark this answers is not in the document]
tristanlatr: Are you suggesting removing the `Token.inline` attribute ?

tristanlatr (commented):
Thanks for the review @mthuurne , 

I hope I have fixed the issues you raised. 

I could not simplify the `_to_node(self, tree: Union[Element, str])` call such that it does not use the Union, though. 

Thanks

pydoctor/test/epydoc/test_epytext2html.py
tristanlatr: Done.

pydoctor/node2stan.py
```diff
@@ -0,0 +1,246 @@
+"""
+Helper function to convert `docutils` nodes to Stan tree.
```
tristanlatr, line 2:
```suggestion
Helper function to convert L{docutils} nodes to Stan tree.
```

pydoctor/epydoc/markup/epytext.py
```diff
@@ -264,16 +266,15 @@ def parse(text: str, errors: Optional[List[ParseError]] = None) -> Optional[Elem
     @return: a DOM tree encoding the contents of an epytext string,
         or C{None} if non-fatal errors were encountered and no C{errors}
         accumulator was provided.
-    @raise ParseError: If C{errors} is C{None} and an error is
-        encountered while parsing.
+    @raise ParseError: If a fatal error is encountered while parsing.
```
tristanlatr, line 269:
```suggestion
    @raise ParseError: If C{errors} is C{None} and an error is
        encountered while parsing.
```

pydoctor/epydoc2stan.py
mthuurne: I can't reproduce the problem under Python 3.9 if the code looks like this:
```py
        if isinstance(sub, ast.Index):
            if sys.version_info < (3, 9):
                # In Python < 3.9, non-slices are always wrapped in an Index node.
                sub = sub.value
```
Only if I remove the `sys.version_info < (3, 9)` check, I see the warning.

To test, I added `basepython = python3.9` under `[testenv:mypy]` in `tox.ini` and added `python --version` to the command; it prints 3.9.5.

As for a solution, I think this is probably the safest approach:
```py
        if sys.version_info < (3, 9) and isinstance(sub, ast.Index):
            # In Python < 3.9, non-slices are always wrapped in an Index node.
            sub = sub.value
```
This will still work if the `Index` class is removed at some point.

pydoctor/epydoc/markup/epytext.py
mthuurne: I'm not going to block this PR because it's changing too many things, but in general I'd be quicker and more motivated to review PRs that are smaller. So next time you notice that you'll need to do some cleanup before adding a feature, please create aim for multiple smaller PRs instead of one big one.

Think of Git history as a series of patches rather than a chronology of development. You can use `git rebase -i` to re-order commits, putting all the cleanups at the start of the chain. That way, it's easy to split them off into a separate branch later if necessary. If you do this while developing, you may have to occasionally do some conflict resolution, but it shouldn't be nearly as much work as doing it after all the code has been written.

tristanlatr: > So next time you notice that you'll need to do some cleanup before adding a feature, please aim for multiple smaller PRs instead of one big one.

Yes I Will try to do that. 

> You can use `git rebase -i` to re-order commits, putting all the cleanups at the start of the chain.

I'll try to get my head around git rebasing, it's still something that I'm not quite familiar with yet.

pydoctor/epydoc2stan.py
tristanlatr: Done!

pydoctor/epydoc/markup/plaintext.py
mthuurne: That's a step in the right direction.

However, what I don't understand is why `ParsedPlaintextDocstring` needs to override `to_stan()`. A comment says "Also, the consolidated fields handling in restructuredtext.py relies on this "pre" class.", but I can't find that back in the code. I do see `ParsedPlaintextDocstring` being used as a fallback for field splitting, but I don't see any dependency on Stan there. Maybe I'm just overlooking it though.

If at some point `to_stan()` is not overridden by any subclass, it could be moved out of `ParsedDocstring`.

Note that the epydoc code was only imported into pydoctor as part of the Python 3 porting in 2019. A few months later I reduced the epytext -> HTML -> Stan -> HTML pipeline to just epytext -> Stan -> HTML, which is when the Stan dependency was added to the epytext code. So the current situation hasn't existed for all that long.

Stan isn't very good for introspection, both because its API isn't designed for that and because the generated HTML is not semantic enough. Therefore implementing things like summary auto-generation on Stan trees would be very error prone. I hope it will be easier to do on docutils trees.

pydoctor/epydoc/markup/epytext.py
mthuurne: Indeed the `restructuredtext` code doesn't resolve links, so maybe doing it very early isn't a good idea. It would be good to resolve links before generating HTML though. I'm fine with that being done in a separate PR.

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
mthuurne: I was actually suggesting removing `inline` from `Element`, but if there is no other use for it, maybe it can be removed from `Token` as well.

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
mthuurne: If `tree.children[1]` is somehow not an `Element`, I'd rather have the code crash than silently produce incomplete output.

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
tristanlatr: Should this be done now, though?

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
tristanlatr: Very well, I'll revert back to original behaviour.

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
tristanlatr: sorry I just saw an issue in my code, I'll retry it..

pydoctor/epydoc/markup/plaintext.py
tristanlatr: Happy that you like the changes. 

Regarding why ParsedPlaintextDocstring needs to override to_stan(). Please look at this [test case](https://github.com/twisted/pydoctor/blob/master/pydoctor/test/epydoc/restructuredtext.doctest#L43). My opinion it that it touches stuff outside the scope of this PR, so I'de rather not modify this behaviour now. 

> I hope it will be easier to do on docutils trees.

Yes it is, I think.

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
mthuurne: You forgot to substitute `child` for `tree` in this line:
```py
variables.append(set_node_attributes(nodes.Text(tree), document=self._document))
```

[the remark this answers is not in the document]
tristanlatr: Yes that's what I see. It's fixed now! Thanks

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
tristanlatr: Done.

pydoctor/epydoc/markup/plaintext.py
tristanlatr: The new "pipeline" after this PR will be: 

Epytext -> Docutils -> Stan (with `HTMLTranslator` and `html2stan()`) -> HTML

tristanlatr:
Thanks for the hints, @mthuurne. We should be good now. 

I hope I was clear on the why `ParsedEpytextDocstring` needs to override `to_stan()` method. 

Tell me if there is anything else. 

Thanks.

pydoctor/epydoc/markup/epytext.py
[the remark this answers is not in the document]
mthuurne: Not strictly, but unlike some of the other changes that could be postponed, this is one that's easy to forget about. So I would be in favor of doing it now.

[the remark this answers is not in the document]
tristanlatr: Done!

pydoctor/epydoc/markup/epytext.py
tristanlatr: The `link_xref()` call is done in our `HTMLTranslator` when converting docutils documents to stan. 

I suppose the right thing to do would be to use a custom Transform and apply it to all docstring documents. But this means that instead of having a `ParsedDocstring.to_node() -> nodes.document` method, we should provide a `ParsedDocstring.document: nodes.document` attribute such that we can transform it in place. I guess it could work with the `to_node()` method  too(since it essentially returns the self._document attribute, bu that would be a bit confusing for future maintenance.). Also the thing to consider is that, currently not all `ParsedDocstring` subclasses implements `to_node()`. Namely the `AnnotationDocstring` and the `ParsedPlaintextDocstring`. `AnnotationDocstring` will be removed in #402, and we discussed `ParsedPlaintextDocstring` already.

tristanlatr: So maybe we can create a separate ticket to take care of that ?

tristanlatr:
So I’ve removed all usages of the inline attribute in the `epytext` parser. 

The last thing I see that could be added here, if the `ParsedPlaintexDocstring` situation is not acceptable, is to use a custom node class to store the plaintext docstring, then, edit the HTMLTranslator to output the right `<p class=« pre »>` tags when visiting this new node class. This would solve the plaintext issue with to_stan()

pydoctor/epydoc/markup/epytext.py
tristanlatr: Done #421

tristanlatr:
> The last thing I see that could be added here, if the ParsedPlaintexDocstring situation is not acceptable, is to use a custom node class to store the plaintext docstring, then, edit the HTMLTranslator to output the right <p class=« pre »> tags when visiting this new node class. This would solve the plaintext issue with to_stan()

I've tried on my local branch to solve this issue with the above approach but it seems that I can't make it work for some reason :/ 

Should we merge this PR in the current state ? Or should I spend more time trying to make this work ?

Thanks

tristanlatr:
Hi @mthuurne, 

Are we good with this PR ? 

Thanks

tristanlatr:
I've tried again to make it work but it seems that there is something that I don't understand. 

I've tried to use this sphinx helper function: https://github.com/sphinx-doc/sphinx/blob/4.x/sphinx/util/docutils.py#L103 but I cannot make the custom `visit_plaintext()` method to run within the visitor... 

I suggest we leave this for later, this code already relatively de-coupled stan related stuff from ParsedDocstring. 

Tell me what you think.

Thanks.

adiroiban (approved):
Hi, since it takes too long to review, I think that this should be merged to not block future development.

@tristanlatr looks like the only current active developer and to me it's fair to access this PR as it is.

Thanks @tristanlatr

tristanlatr:
🙌🥳

Issue #376, opened 2021-03-23
AssertionError


tristanlatr:
I'm trying to generate the [Natural Language Toolkit](https://github.com/nltk/nltk) API docs, they used to do it with Epytext.

I'm running into an `AssertionError`, I believe we should not run into those right?

Also, the traceback is not consistent from one call to another so here are two:

```
% git clone git@github.com:tristanlatr/nltk.git
% cd nltk                              
% pydoctor --docformat=plaintext ./nltk
adding directory /Users/tristan/Documents/pydoctor/nltk/nltk
369 modules and packages discovered
Guessing 'nltk' for project name.
32/342 modules processed, 0 warnings
moving 'nltk.lm.models.MLE' into 'nltk.lm.__init__'
moving 'nltk.lm.models.Lidstone' into 'nltk.lm.__init__'
moving 'nltk.lm.models.Laplace' into 'nltk.lm.__init__'
moving 'nltk.lm.models.WittenBellInterpolated' into 'nltk.lm.__init__'
moving 'nltk.lm.models.KneserNeyInterpolated' into 'nltk.lm.__init__'
moving 'nltk.lm.counter.NgramCounter' into 'nltk.lm.__init__'
moving 'nltk.lm.vocabulary.Vocabulary' into 'nltk.lm.__init__'
185/342 modules processed, 0 warnings
/Users/tristan/Documents/pydoctor/nltk/nltk/corpus/reader/wordnet.py:1806: Unable to figure out value for __doc__ assignment, maybe too complex
/Users/tristan/Documents/pydoctor/nltk/nltk/corpus/reader/wordnet.py:1811: Unable to figure out value for __doc__ assignment, maybe too complex
/Users/tristan/Documents/pydoctor/nltk/nltk/corpus/reader/wordnet.py:1816: Unable to figure out value for __doc__ assignment, maybe too complex
/Users/tristan/Documents/pydoctor/nltk/nltk/corpus/reader/wordnet.py:1821: Unable to figure out value for __doc__ assignment, maybe too complex
/Users/tristan/Documents/pydoctor/nltk/nltk/corpus/reader/wordnet.py:1826: Unable to figure out value for __doc__ assignment, maybe too complex
/Users/tristan/Documents/pydoctor/nltk/nltk/corpus/reader/wordnet.py:1831: Unable to figure out value for __doc__ assignment, maybe too complex
[...]

PR #387, opened 2021-04-25
Pass filename to ast.parse() for more useful warnings

mthuurne:
If any warnings are issued by `ast.parse()`, we'll now see the file path instead of `<unknown>` being printed.

To see parse warnings being logged, `export PYTHONWARNINGS=always` prior to running pydoctor. To trigger a warning, you can use for example `\s` or `\]` in a string literal without the `r` (raw) prefix.

I encountered this problem when investigating #376, but to reproduce it on `nltk` you might have to merge #360 first to avoid the assert failing.


tristanlatr (approved):
Nice thanks!

PR #392, opened 2021-05-07
Annotate templatewriter

mthuurne:
Now that `twisted.web.template` has type annotations, we can fully annotate `pydoctor.templatewriter`.

### DIFF
--- a/pydoctor/templatewriter/pages/__init__.py
+++ b/pydoctor/templatewriter/pages/__init__.py
@@ -1,17 +1,25 @@
 """The classes that turn  L{Documentable} instances into objects we can render."""
 
-from typing import Any, Dict, Iterator, List, Optional, Mapping, Sequence, Tuple, Union, Type
+from typing import (
+    TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Mapping, Sequence,
+    Tuple, Type, Union
+)
 import ast
 import abc
 
-from twisted.web.template import Element, Flattenable, Tag, renderer, tags
 import astor
-
 from twisted.web.iweb import IRenderable, ITemplateLoader, IRequest
+from twisted.web.template import Element, Flattenable, Tag, renderer, tags
+
 from pydoctor import epydoc2stan, model, zopeinterface, __version__
 from pydoctor.astbuilder import node2fullname
 from pydoctor.templatewriter import util, TemplateLookup, TemplateElement
 from pydoctor.templatewriter.pages.table import ChildTable
+
+if TYPE_CHECKING:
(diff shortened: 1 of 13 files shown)


pydoctor/templatewriter/pages/__init__.py
```diff
@@ -189,78 +195,81 @@ def namespace(self, obj: model.Documentable) -> Sequence[Union[Tag, str]]:
...
         attr_loader = AttributeChild.lookup_loader(self.template_lookup)
 
         for c in self.methods():
             if isinstance(c, model.Function):
                 r.append(FunctionChild(self.docgetter, c, self.functionExtras(c), func_loader))
-            else:
+            elif isinstance(c, model.Attribute):
                 r.append(AttributeChild(self.docgetter, c, self.functionExtras(c), attr_loader))
+            else:
+                assert False, type(c)
```
tristanlatr, line 262: Maybe we could raise a `RuntimeError` instead ? saying something like `"unexpected child item in {self.ob.fullName()}"`

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -445,32 +467,33 @@ def baseTables(self, request, item):
...
+            tail: List["Flattenable"] = []
             for b in reversed(bases_to_mention):
                 tail.append(tags.code(epydoc2stan.taglink(b, page_url, b.name)))
                 tail.append(', ')
             del tail[-1]
             r.extend([' (via ', tail, ')'])
         return r
 
-    def functionExtras(self, ob):
+    def functionExtras(self, data: model.Documentable) -> List["Flattenable"]:
```
tristanlatr, line 485: We name the documentable `obj` everywhere, why do you change that here ?

tristanlatr (commented):
This is great!

Thanks a lot.

pydoctor/templatewriter/pages/__init__.py
mthuurne: This is an internal consistency check, so I think an assert is fine here. It checks the types returned by `method()`, which is defined inside the same class. Poor naming though, since it doesn't return methods.

pydoctor/templatewriter/pages/__init__.py
mthuurne: It was named `data` in `CommonPage` and I copy-pasted the annotated header. I could rename it in `CommonPage` instead.

mthuurne:
I want to address the `_deprecatedInfo` and `data` comments; I'll merge after that.

tristanlatr (approved):
Looks great!

PR #396, opened 2021-05-27
Delete apilinks_sphinxext.py

tristanlatr:
This is probably a vestige from a time when intersphinx was not supported by pydoctor yet.


mthuurne:
Yes, apilinks was a custom solution from a time before pydoctor supported Intersphinx. It was used by Twisted until I removed it half a year ago, see [Twisted ticket 10057](https://twistedmatrix.com/trac/ticket/10057). Twisted was using its own copy of the apilinks extension, so it would be safe to remove this from pydoctor even if Twisted was still using it.

Issue #13, opened 2015-01-12
Does not work with SWIG-generated classes


mwhudson:
affects pydoctor

Pydoctor does not generate documentation for classes wrapped by SWIG
(www.swig.org). SWIG creates classes like this:

class event(object):
    thisown = _swig_property(lambda x: x.this.own(), lambda x, v:
x.this.own(v), doc='The membership flag')
    __repr__ = _swig_repr
    def __init__(self, *args, **kwargs):
        _events.event_swiginit(self,_events.new_event(*args, **kwargs))
    __swig_destroy__ = _events.delete_event
event.loop_once = new_instancemethod(_events.event_loop_once,None,event)
event.loop_wait = new_instancemethod(_events.event_loop_wait,None,event)
event_swigregister = _events.event_swigregister
event_swigregister(event)

pydoctor will in this case only document event.__init__(). It would be
nice if it would also document event.loop_once and event.loop_wait.
-- 
Jelmer Vernooij <email address hidden> - http://samba.org/~jelmer/
Jabber: <email address hidden>

mwhudson:
(by mwhudson)
Yuck :)

Something can probably be done, though.

PR #397, opened 2021-05-29
Fixes #385

### LINKED ISSUE #385: Separate constants from variables
Most module "variables" are actually constants, but are listed in the docs as variables. It would be nice if they could be listed as constants instead.

To detect whether a name is a constant, we could:
- have the user annotate it with `typing.Final`
- use a naming convention, where all-caps indicates a constant (PEP8 states this is common, but doesn't prescribe it)
- use code analysis (variable that's assigned only once of an immutable type)
- introduce a docstring field to mark a name as a constant

Original discussion was here: https://github.com/twisted/pydoctor/issues/364#issuecomment-788652903

tristanlatr:
Mark the variables as contants when their names is all caps or if using `typing.Final` annotation.

Introduces a new kind: CONSTANT.

Also new: Attribute.value, this will be used when adding the feature #377 in the near future.

### DIFF
--- a/docs/epytext_demo/demo_epytext_module.py
+++ b/docs/epytext_demo/demo_epytext_module.py
@@ -9,6 +9,11 @@ from somelib import SomeInterface
 import zope.interface
 import zope.schema
 
+LANG = 'Fr'
+"""
+This is a constant.
+"""
+
 def demo_fields_docstring_arguments(m, b):  # type: ignore
     """
     Fields are used to describe specific properties of a documented object.
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -168,6 +168,17 @@ def attrib_args(expr: ast.expr, ctx: model.Documentable) -> Optional[BoundArgume
                 )
     return None
 
+def is_constant(name:str, annotation: Optional[ast.expr], 
+                ctx: model.Documentable) -> bool:
+    """
+    Determine if the given assignment is a constant. 
+
+    To detect whether a assignment is a constant, this checks two things:
+        - all-caps variable name
+        - typing.Final annotation
+    """
+    return name == name.upper() or \
+        node2fullname(annotation, ctx) == "typing.Final"
 
 class ModuleVistor(ast.NodeVisitor):
     currAttr: Optional[model.Documentable]
@@ -447,6 +458,11 @@ class ModuleVistor(ast.NodeVisitor):
             obj.kind = model.DocumentableKind.VARIABLE
             if annotation is None and expr is not None:
                 annotation = _infer_type(expr)
+            
+            if is_constant(target, annotation, obj):
+                obj.kind = model.DocumentableKind.CONSTANT
+                obj.value = expr
+            
             obj.annotation = annotation
             obj.setLineNumber(lineno)
             self.newAttr = obj
@@ -488,6 +504,11 @@ class ModuleVistor(ast.NodeVisitor):
                 annotation = self._annotation_from_attrib(expr, cls)
             if annotation is None:
                 annotation = _infer_type(expr)
+
+        if is_constant(name, annotation, obj):
+            obj.kind = model.DocumentableKind.CONSTANT
+            obj.value = expr
+    
         obj.annotation = annotation
         obj.setLineNumber(lineno)
         self.newAttr = obj
@@ -512,6 +533,12 @@ class ModuleVistor(ast.NodeVisitor):
         obj.kind = model.DocumentableKind.INSTANCE_VARIABLE
         if annotation is None and expr is not None:
             annotation = _infer_type(expr)
+
+        if is_constant(name, annotation, obj):
+            # stuff like: self.DOCUMENT = ""
+            obj.kind = model.DocumentableKind.CONSTANT
+            obj.value = expr
+
         obj.annotation = annotation
         obj.setLineNumber(lineno)
         self.newAttr = obj
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -95,6 +95,7 @@ class DocumentableKind(Enum):
     STATIC_METHOD       = 600
     METHOD              = 500
     FUNCTION            = 400
+    CONSTANT            = 310
     CLASS_VARIABLE      = 300
     SCHEMA_FIELD        = 220
     ATTRIBUTE           = 210
@@ -504,7 +505,7 @@ class Attribute(Inheritable):
     kind: Optional[DocumentableKind] = DocumentableKind.ATTRIBUTE
     annotation: Optional[ast.expr]
     decorators: Optional[Sequence[ast.expr]] = None
-
+    value: Optional[ast.expr] = None
 
 # Work around the attributes of the same name within the System class.
 _ModuleT = Module
--- a/pydoctor/test/test_astbuilder.py
+++ b/pydoctor/test/test_astbuilder.py
@@ -1612,3 +1612,46 @@ def test_ignore_function_contents(systemcls: Type[model.System]) -> None:
     ''', systemcls=systemcls)
     outer = mod.contents['outer']
     assert not outer.contents
+
+@systemcls_param
+def test_constant_module(systemcls: Type[model.System]) -> None:
+    mod = fromText('''
+    LANG = 'FR'
+    ''', systemcls=systemcls)
+    assert getattr(mod.contents['LANG'], 'kind') == model.DocumentableKind.CONSTANT
+    assert getattr(mod.contents['LANG'], 'value') is not None
+
+@systemcls_param
+def test_constant_module_with_final(systemcls: Type[model.System]) -> None:
+    mod = fromText('''
+    from typing import Final
+    lang: Final = 'fr'
+    ''', systemcls=systemcls)
+    assert getattr(mod.contents['lang'], 'kind') == model.DocumentableKind.CONSTANT
+    assert getattr(mod.contents['lang'], 'value') is not None
+
+@systemcls_param
+def test_constant_class(systemcls: Type[model.System]) -> None:
+    mod = fromText('''
+    from typing import Final
+    from datetime import datetime
+    class Clazz:
+        """Class."""
+        LANG = 'FR'
+    ''', systemcls=systemcls)
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    assert getattr(mod.resolveName('Clazz.LANG'), 'value') is not None
+
+
+@systemcls_param
+def test_constant_init(systemcls: Type[model.System]) -> None:
+    mod = fromText('''
+    from typing import Final
+    from datetime import datetime
+    class Clazz:
+        """Class."""
+        def __init__(**args):
+            self.LANG = 'FR'
+    ''', systemcls=systemcls)
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    assert getattr(mod.resolveName('Clazz.LANG'), 'value') is not None
--- a/docs/restructuredtext_demo/demo_restructuredtext_module.py
+++ b/docs/restructuredtext_demo/demo_restructuredtext_module.py
@@ -4,6 +4,11 @@ This is a module demonstrating reST code documentation features.
 Most part of this documentation is using Python type hinting.
 """
 
+LANG = 'Fr'
+"""
+This is a constant.
+"""
+
 def demo_fields_docstring_arguments(m, b):  # type: ignore
     """
     Fields are used to describe specific properties of a documented object.
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -917,6 +917,7 @@ def format_kind(kind: model.DocumentableKind, plural: bool = False) -> str:
         model.DocumentableKind.PROPERTY        : 'Property',
         model.DocumentableKind.VARIABLE        : 'Variable',
         model.DocumentableKind.SCHEMA_FIELD    : 'Attribute',
+        model.DocumentableKind.CONSTANT        : 'Constant',
     }
     plurals = {
         model.DocumentableKind.CLASS           : 'Classes', 


pydoctor/test/test_astbuilder.py
```diff
@@ -1612,3 +1612,79 @@ def func():
...
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    val = getattr(mod.resolveName('Clazz.LANG'), 'value')
+    assert val is not None
+    assert ast.literal_eval(val) == 'FR'
+
+    captured = capsys.readouterr().out
+    assert "mod:4: Assignment to constant \"LANG\" overrides previous assignment.\n" in captured
+
+@systemcls_param
+def test_constant_override_do_not_warns_when_defined_in_docstring(systemcls: Type[model.System], capsys: CapSys) -> None:
```
adiroiban, line 1673: I think that a docstring that describes what is the purpose of this test will help.

as a new person to this code, I had to read the test name multiple times and try to "decode" its name.

The problem is that from the test name I don't know what is the expected behaviour.

I guess that the purpose / test scenario is:

```
Constant variables can be defined at docstring level without any warnings.
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1612,3 +1612,79 @@ def func():
...
+        def __init__(**args):
+            self.LANG = 'FR'
+    ''', systemcls=systemcls, modname="mod")
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    val = getattr(mod.resolveName('Clazz.LANG'), 'value')
+    assert val is not None
+    assert ast.literal_eval(val) == 'FR'
+
+    captured = capsys.readouterr().out
+    assert "mod:4: Assignment to constant \"LANG\" overrides previous assignment.\n" in captured
```
adiroiban, line 1670: This is not that bad... but I am not sure if pydoctor should go this path... this looks like a pyflakes or mypy type of warning.

pydoctor/test/test_astbuilder.py
```diff
@@ -1612,3 +1612,79 @@ def func():
...
+@systemcls_param
+def test_constant_init(systemcls: Type[model.System]) -> None:
+    mod = fromText('''
+    class Clazz:
+        """Class."""
+        def __init__(**args):
+            self.LANG = 'FR'
+    ''', systemcls=systemcls)
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    assert getattr(mod.resolveName('Clazz.LANG'), 'value') is not None
```
adiroiban, line 1653: why not assert the actual value here?

pydoctor/test/test_astbuilder.py
```diff
@@ -1612,3 +1612,79 @@ def func():
...
+    class Clazz:
+        """Class."""
+        LANG = 'FR'
+    ''', systemcls=systemcls)
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    assert getattr(mod.resolveName('Clazz.LANG'), 'value') is not None
+
+
+@systemcls_param
+def test_constant_init(systemcls: Type[model.System]) -> None:
```
adiroiban, line 1645: Maybe as docstring
```
It can recognize constants define as instance members.
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1612,3 +1612,79 @@ def func():
...
+def test_constant_module_with_final(systemcls: Type[model.System]) -> None:
+    mod = fromText('''
+    from typing import Final
+    lang: Final = 'fr'
+    ''', systemcls=systemcls)
+    assert getattr(mod.contents['lang'], 'kind') == model.DocumentableKind.CONSTANT
+    assert getattr(mod.contents['lang'], 'value') is not None
+
+@systemcls_param
+def test_constant_class(systemcls: Type[model.System]) -> None:
```
adiroiban, line 1634: Maybe as docstring
```
It can recognize constants define as class members.
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1612,3 +1612,79 @@ def func():
     ''', systemcls=systemcls)
     outer = mod.contents['outer']
     assert not outer.contents
+
+@systemcls_param
+def test_constant_module(systemcls: Type[model.System]) -> None:
```
adiroiban, line 1617: Maybe as docstring
```
It can recognize constants define as module members.
```

adiroiban:
Thanks for the PR.

------------

I only did a quick review but it looks very good. Only minor comments.

For easy reference, the new LANG constant is rendered as 

https://pydoctor--397.org.readthedocs.build/en/397/docformat/epytext/epytext_demo.demo_epytext_module.html#LANG

And this is the rendered HTML

```
<tr class="constant">
    
    <td>Constant</td>
    <td><code><a href="#LANG">LANG</a></code></td>
    <td><span>This is a constant.</span></td>
  </tr>
```

-------------

I have no comment on the CSS styling.
I think that to keep the size of this PR small, is best if we can go with the current implementation.

--------------

I see that in the module is defined at the top of the file, but in the table and HTML it is at the end.
is this on purpose?

-----------

I see that in the RST / Epytext examples there is a `LANG` contant. Would it be possible to update the example to also include a contant which is not all uppercase but has the `typing.Final` ?

pydoctor/test/test_astbuilder.py
```diff
@@ -1612,3 +1612,79 @@ def func():
...
+    from enum import Enum
+    class PrivacyClass(Enum):
+        """
+        @cvar HIDDEN: Don't show the object at all.
+        @cvar PRIVATE: Show, but de-emphasize the object.
+        @cvar VISIBLE: Show the object as normal.
+        """
+        HIDDEN = 0
+        PRIVATE = 1
+        VISIBLE = 2
```
adiroiban, line 1684: Maybe add some diversity to the testing contants...and a "real" variable overwriting.

I expect that defining the constant multiple time can be found via an `if/else` expression.

I read this test multiple time trying to understand why we need 3 constants defined for this test... and I was not sure why a single constant definition is not enough.

```suggestion
        @cvar UPPER_CASE: Don't show the object at all.
        @cvar lower_case: Show, but de-emphasize the object.
        """
        UPPER_CASE = 0
        lower_case = 1
        UPPER_CASE = 2  # Overwrite the constant.
```

pydoctor/test/test_astbuilder.py
mthuurne: I think the purpose of this warning isn't to act as a static code checker, but to inform the user that pydoctor is disregarding the other definitions, which means that if the code is not redundant, the generated docs are incomplete.

Are there valid reasons to define a constant more than once?
- maybe the value depends on the presence of certain API (OS type, Python version or optional dependency)
- maybe a base value is assigned first and later augmented to add detail

I don't know how common these situations are. But until we know, having a warning would be useful, so we can find real-world examples of duplicate definitions to consider for refining the handling.

pydoctor/test/test_astbuilder.py
tristanlatr: Thanks @adiroiban and @mthuurne for the quick comments. 

I agree we don't need 3 constants for this test, I agree. A lowercase name, even in an `Enum, will not be recognized as Constant unless we add `typing.Final` annotation though.

pydoctor/test/test_astbuilder.py
tristanlatr: One word a bout this warning. It will be useful primarily when #377 is implemented. Then it would mean that the value has been overridden and the generated repr of this value is not the original. 
Right now it does not serve any particular purpose.

pydoctor/test/test_astbuilder.py
tristanlatr: Also the override is tested in another test case.

tristanlatr:
> I see that in the module is defined at the top of the file, but in the table and HTML it is at the end.
is this on purpose?

No. I mean the constants are usually defined at the top of the module. Then I assigned `310` as the `DocumentableKind` value, meaning that it will display right after functions, before class variables. I agree, It might not be the best place, please advise another organisation for the `DocumentableKind` values if you think it should be changed. I think of grouping the class stuff together (class methods and variables), before the instance stuff, So the constants would go right before ?

tristanlatr:
RE CSS styling: Do we want to a special color ? It should not be the same as the variables though.

pydoctor/test/test_astbuilder.py
tristanlatr: No particular reason. I assert the values everywhere now.

pydoctor/test/test_astbuilder.py
tristanlatr: Thanks!

pydoctor/test/test_astbuilder.py
tristanlatr: Thanks!

pydoctor/test/test_astbuilder.py
tristanlatr: Thanks!

tristanlatr: Thanks!

pydoctor/test/test_astbuilder.py
tristanlatr: Thanks!

tristanlatr:
I've implemented most of your comments @adiroiban , Thanks.

There is still something that bothers me, it's that the warning line number is pointing to the original assign line where it should point to the overriding assignment instead.

tristanlatr:
It's all good now.

It should be clean.

docs/epytext_demo/demo_epytext_module.py
```diff
@@ -8,6 +8,17 @@
...
 import zope.interface
 import zope.schema
+from typing import Final
+
+LANG = 'Fr'
+"""
+This is a constant.
+"""
+
+lang: Final = ['Fr', 'En']
```
adiroiban, line 18: Maybe something like this for complete typing ?

```suggestion
lang: Final[Sequence[str]] = ['Fr', 'En']
```

tristanlatr: That's interesting stuff, not even sure `typing. Final[something]` will actually work!

Let me check...

adiroiban:
Thanks for the update  and info.

----

> RE CSS styling: Do we want to a special color ? It should not be the same as the variables though.

I don't have any suggestion for color.
I think that we can leave it as it is...and if someone has a suggestion later, it can be done in a separate PR.

------

I think that for now, this is good enough.

Since this is a constant and the value is stored inside pydoctor I was expecting to see the default value presented in the HTML rendering 

image


But is ok if that will be done in a separate PR.


------------

my review was only for the test code and user interface.
it would help if  @mthuurne  could review the internal implementation.
To me it looks ok.

Thanks again

mthuurne:
> Since this is a constant and the value is stored inside pydoctor I was expecting to see the default value presented in the HTML rendering
> 
> image
> 
> But is ok if that will be done in a separate PR.

We already have a feature request for this: #377. I would prefer this to be handled in a separate PR, since it's not a trivial feature.

Regarding the HTML rendering: do we want to show the annotation or the type here? Let's say the code looks like this:
```py
FOO = 1
BAR: Final[int] = 2
```
Should the documentation read:
```
FOO: int = 1
BAR: Final[int] = 2
```
or
```
FOO: int = 1
BAR: int = 2
```
I think the latter would be more consistent and easier on the eyes.

> my review was only for the test code and user interface.
> it would help if @mthuurne could review the internal implementation.
> To me it looks ok.

I'll have a look.

docs/epytext_demo/demo_epytext_module.py
mthuurne: Whether it already works or not, it would be good to have a unit test case for it.

pydoctor/astbuilder.py
```diff
@@ -426,6 +438,21 @@ def _handleOldSchoolMethodDecoration(self, target: str, expr: Optional[ast.expr]
...
+                obj: model.Attribute, is_first_assignment: bool, lineno: int) -> None:
+        # Must be called after obj.setLineNumber() to have the right line number in the warning.
+        
+        if is_constant(name, obj):
+            obj.kind = model.DocumentableKind.CONSTANT
+            
+            if not is_first_assignment and obj.value is not None:
+                # This warning will be useful when #377 is implemented.
+                obj.report(f'Assignment to constant "{name}" overrides previous assignment.', 
+                            section='ast', lineno_offset=lineno-obj.linenumber)
```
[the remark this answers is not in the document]
mthuurne: Maybe it would make sense to mention both the initial and the override line number in the log message.

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
...
+
+@systemcls_param
+def test_constant_module(systemcls: Type[model.System]) -> None:
+    """
+    It can recognize constants define as module members.
+    """
+    mod = fromText('''
+    LANG = 'FR'
+    ''', systemcls=systemcls)
+    assert getattr(mod.contents['LANG'], 'kind') == model.DocumentableKind.CONSTANT
```
mthuurne, line 1693:
```suggestion
    lang = mod.contents['LANG']
    assert isinstance(lang, model.Attribute)
    assert lang.kind is model.DocumentableKind.CONSTANT
```
This provides mypy with the information it needs to actually check the types, unlike `getattr()` which returns `Any`.

Also you can use `is` instead of `==` to compare enum members.

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
...
+    assert ast.literal_eval(getattr(mod.resolveName('LANG'), 'value') or '') == 'FR'
```
mthuurne, line 1694: Why `resolveName()` here instead of `contents[]`?

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
     ''', systemcls=systemcls)
     outer = mod.contents['outer']
     assert not outer.contents
+
+@systemcls_param
+def test_constant_module(systemcls: Type[model.System]) -> None:
+    """
+    It can recognize constants define as module members.
```
mthuurne, line 1688:
```suggestion
    Module variables with all-uppercase names are recognized as constants.
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
...
+    mod = fromText('''
+    LANG = 'FR'
+    ''', systemcls=systemcls)
+    assert getattr(mod.contents['LANG'], 'kind') == model.DocumentableKind.CONSTANT
+    assert ast.literal_eval(getattr(mod.resolveName('LANG'), 'value') or '') == 'FR'
+
+@systemcls_param
+def test_constant_module_with_final(systemcls: Type[model.System]) -> None:
+    """
+    It can recognize constants defined with typing.Final
```
mthuurne, line 1699:
```suggestion
    Module variables annotated with typing.Final are recognized as constants.
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
...
+    from typing import Final
+    lang: Final = 'fr'
+    ''', systemcls=systemcls)
+    assert getattr(mod.contents['lang'], 'kind') == model.DocumentableKind.CONSTANT
+    assert ast.literal_eval(getattr(mod.resolveName('lang'), 'value') or '') == 'fr'
+
+@systemcls_param
+def test_constant_class(systemcls: Type[model.System]) -> None:
+    """
+    It can recognize constants define as class members.
```
mthuurne, line 1711:
```suggestion
    Class variables with all-uppercase names are recognized as constants.
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
...
+
+@systemcls_param
+def test_constant_init(systemcls: Type[model.System]) -> None:
+    """
+    It can recognize constants define as instance members.
+    """
+    mod = fromText('''
+    class Clazz:
+        """Class."""
+        def __init__(**args):
```
mthuurne, line 1730: This is missing `self`.

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
...
+        LANG = 'EN'
+        def __init__(**args):
+            self.LANG = 'FR'
+    ''', systemcls=systemcls, modname="mod")
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    assert ast.literal_eval(getattr(mod.resolveName('Clazz.LANG'), 'value') or '') == 'FR'
+
+    captured = capsys.readouterr().out
+    assert "mod:6: Assignment to constant \"LANG\" overrides previous assignment.\n" in captured
+
```
mthuurne, line 1753: The test case can be split here.

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
...
+        """
+        @cvar LANG: French.
+        """
+        LANG = 99
+    ''', systemcls=systemcls, modname="mod")
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    assert ast.literal_eval(getattr(mod.resolveName('Clazz.LANG'), 'value') or '') == 99
+    captured = capsys.readouterr().out
+    assert not captured
+
```
mthuurne, line 1783: The test case can be split here.

pydoctor/test/test_astbuilder.py
```diff
@@ -1681,3 +1681,113 @@ def func():
...
+    It can recognize constants define as instance members.
+    """
+    mod = fromText('''
+    class Clazz:
+        """Class."""
+        def __init__(**args):
+            self.LANG = 'FR'
+    ''', systemcls=systemcls)
+    assert getattr(mod.resolveName('Clazz.LANG'), 'kind') == model.DocumentableKind.CONSTANT
+    assert ast.literal_eval(getattr(mod.resolveName('Clazz.LANG'), 'value') or '') == 'FR'
```
mthuurne, line 1734: Should we actually consider immutable instance variables to be constants?

The convention of constants with all-caps names originates from languages that have compile-time constants, like C and Java. If we'd translate that to Python, a constant would characterized by being both immutable and global.

If we instead define a constant as an immutable named value, then instances could have constants too. We'd need to recognize a whole lot of additional code constructs though to handle this consistently. For example, all attributes in a class decorated with `@attr.s(frozen=True)` would be constants. And perhaps even read-only properties should be considered constants then.

I think it would be simplest to for now only have globals as constants and later decide how we want to document immutable (read-only) instance attributes (both variables and properties).

mthuurne (changes requested):
Some comments on implementation details, but the overall approach looks good.

My only concern about the approach is whether instance variables should be recognized as constants, or whether that should be reserved for global names.

pydoctor/astbuilder.py
```diff
@@ -426,6 +438,21 @@ def _handleOldSchoolMethodDecoration(self, target: str, expr: Optional[ast.expr]
                     target_obj.kind = model.DocumentableKind.CLASS_METHOD
                 return True
         return False
+    
+    @staticmethod
```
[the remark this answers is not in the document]
tristanlatr: By design.

If the function doesn't need to be an instance method, I prefer it to be a class method, and if it doesn't need to be a class method, I prefer to it to be a static method. `self` will be unused anyway.

pydoctor/astbuilder.py
```diff
@@ -426,6 +438,21 @@ def _handleOldSchoolMethodDecoration(self, target: str, expr: Optional[ast.expr]
...
+    def _handleMaybeConstant(name:str, value: Optional[ast.expr], 
+                obj: model.Attribute, is_first_assignment: bool, lineno: int) -> None:
+        # Must be called after obj.setLineNumber() to have the right line number in the warning.
+        
+        if is_constant(name, obj):
+            obj.kind = model.DocumentableKind.CONSTANT
+            
+            if not is_first_assignment and obj.value is not None:
```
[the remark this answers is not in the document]
tristanlatr: Yes it's possible with variables defined in the module or the class docstring (see test `test_constant_override_do_not_warns_when_defined_in_docstring `). 

Well spotted, I needed a failing test to see it ;-)

pydoctor/test/test_astbuilder.py
tristanlatr: Cmon!

pydoctor/test/test_astbuilder.py
tristanlatr: No reason.

pydoctor/test/test_astbuilder.py
tristanlatr: I understand, I kindda agree too.

tristanlatr (commented):
Thank for the review.

tristanlatr:
Regarding 

> Should the documentation read:
    FOO: int = 1
    BAR: Final[int] = 2
    or
    FOO: int = 1
    BAR: int = 2

Of course the latter, but please advise on how to remove the first part of the name from the AST expression.

docs/epytext_demo/demo_epytext_module.py
tristanlatr: Spoiler: it's not working @mthuurne

pydoctor/astbuilder.py
```diff
@@ -512,6 +512,8 @@ def _handleModuleVar(self,
             
             if is_constant(target, obj):
                 self._handleConstant(obj=obj, value=expr, lineno=lineno)
+                if isinstance(obj.annotation, ast.Subscript):
```
mthuurne, line 515: I think this is incorrect: it is possible for `is_constant()` to return true based on the naming convention, in which case the type being subscripted could be for example `List` or `Mapping` instead of `Final`.

tristanlatr: Indeed, thanks. Anyway it's not displaying the right way as you can see here: https://pydoctor--397.org.readthedocs.build/en/397/docformat/epytext/epytext_demo.demo_epytext_module.html#lang ...

pydoctor/test/test_astbuilder.py
tristanlatr: I've fixed this now. There is now 2 different warnings depending on the situation.

pydoctor/model.py
```diff
@@ -523,9 +524,14 @@ def setup(self) -> None:
 
 class Attribute(Inheritable):
     kind: Optional[DocumentableKind] = DocumentableKind.ATTRIBUTE
-    annotation: Optional[ast.expr]
+    annotation: Optional[ast.AST]
```
mthuurne, line 527: I think an annotation should always be an expression. It seems the code that handles the Python 3.9 version of `Subscript` is responsible for this change. I think the mismatch should be handled there instead of by changing the type here and in `node2fullname()` and `node2dottedname()`.

pydoctor/astbuilder.py
```diff
@@ -426,6 +462,51 @@ def _handleOldSchoolMethodDecoration(self, target: str, expr: Optional[ast.expr]
...
+        # A hack to to display variables annotated with Final with the real type instead.
+        if is_using_typing_final(obj):
+            if isinstance(obj.annotation, ast.Subscript):
+                # Will not display as "Final[str]" but rather only "str"
+                ann_slice = obj.annotation.slice
+                if isinstance(ann_slice, ast.Index):
+                    obj.annotation = ann_slice.value
+                else:
+                    # Python 3.9
+                    obj.annotation = ann_slice
```
mthuurne, line 502: In Python 3.8's documentation of the `ast` module, it is stated that a subscript's slice can be `Slice`, `ExtSlice` or `Index`. Therefore it is not valid to conclude that if `ann_slice` is not an `Index`, we are running under Python 3.9+.

Of course in practice people shouldn't be using for example `Final[str, int:3:2]` as an annotation, but if they do for some reason, pydoctor should handle it gracefully.

I think that for `Slice` and `ExtSlice`, we can issue a warning that the annotation is invalid. And then proceed by not using `ann_slice`.

If `ann_slice` is neither `Index`, `Slice` nor `ExtSlice`, then we do know we are on Python 3.9+ and if mypy doesn't consider `ann_slice` an expression then, we can `assert` that it is.

pydoctor/astbuilder.py
```diff
@@ -439,14 +520,26 @@ def _handleModuleVar(self,
...
+            
             obj.annotation = annotation
             obj.setLineNumber(lineno)
+            
+            if is_constant(target, obj):
+                self._handleConstant(obj=obj, value=expr, lineno=lineno)
+            else:
+                obj.kind = model.DocumentableKind.VARIABLE
+                # We store the expr value for all Attribute in order to be able to 
+                # check if they have been initiated or not.
```
mthuurne, line 540:
```suggestion
                # check if they have been initialized or not.
```

pydoctor/test/test_astbuilder.py
mthuurne: The main point of type-checking the tests is to have additional checks on whether the annotations in the code under test are correct. By using `getattr()`, we lose that advantage. That doesn't mean you can't ever use `getattr()`, but in  this case it can be avoided without too much hassle.

pydoctor/astbuilder.py
```diff
@@ -426,6 +462,51 @@ def _handleOldSchoolMethodDecoration(self, target: str, expr: Optional[ast.expr]
...
+                else:
+                    # Python 3.9
+                    obj.annotation = ann_slice
+            else:
+                # Just plain "Final" annotation.
+                # Simply ignore it because it's duplication of information.
+                if value is not None:
+                    obj.annotation = _infer_type(value)
+                else:
+                    obj.annotation = None
```
mthuurne, line 509:
```suggestion
                obj.annotation = None if value is None else _infer_type(value)
```

mthuurne (changes requested):
Two main things that need changing:
- the way the subscript index is handled
- minimizing the use of `getattr()`

I want to do another review after those are addressed, because I don't think I understand the flow of the `astbuilder` code well enough yet to spot potential problems there.

pydoctor/astbuilder.py
```diff
@@ -167,6 +167,42 @@ def attrib_args(expr: ast.expr, ctx: model.Documentable) -> Optional[BoundArgume
...
+    if fullName == "typing.Final":
+        return True
+    if isinstance(obj.annotation, ast.Subscript):
+        value = obj.annotation.value
+        if isinstance(value, ast.Name) and value.id == 'Final':
+            # Final[...] expressions
+            return True
+        elif isinstance(value, ast.Attribute) and value.attr == 'Final':
+            # typing.Final[...] expressions
+            return True
```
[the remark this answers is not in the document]
tristanlatr: Thanks!

pydoctor/model.py
tristanlatr: Right.

pydoctor/astbuilder.py
tristanlatr: I think we're good on this issue

tristanlatr:
We should be good now for the final review @mthuurne

tristanlatr:
> we should do some static analysis instead

The can easily become a very complex exercise, though. Since variables can be assigned outside of their class scope for instance (#13), Andy what’s an immutable type anyway? Attrs classes can be frozen... I think it’s too complex right now since we don’t have a static analysis that is well decoupled from ast data gathering (thing that I’m trying to think  about with #430).

tristanlatr:
> I'm a little wary of having all caps imply that a symbol is a constant

Let’s also note that only attributes can be constants, not any symbol.

tristanlatr:
> maybe only mark as constant the variables that have "Final" type hinting

99% of actual constants are written with all upper case names, this is a very common practice. Whereas using `Final`, which is very marginal. I don’t want to force the users putting `Final` everywhere when the naming convention can be sufficient. What we can do, nevertheless, is to provide fields `:constant:` and `:variable:` that could be used to override the default behavior when the naming convention is misleading. But this mean extra coding for little use cases, though. 

What is sure, it’s that some documentation should be added to describe the constant recognition

adiroiban:
> Not sure to see what you are talking about.

The page is rendered as

```
LANG: str = (source)
```

So the name is `LANG` , the type is `str`. Up to here all good.
But then the value is **(source)** ?

I think that is should be rendered as something like this, showing that `Fr` is the value.
But this can be done in a separate PR.

```
LANG: str = 'Fr' (source)
```


Makes sense? Thanks

tristanlatr:
Yes, see the PR #402

tristanlatr:
Can I get an approval on this PR ?

PR #398, opened 2021-05-30
Add -vv flag to pytest

tristanlatr:
I'm always adding it locally and it creates conflicts all the time.
I figured it's more simple to just make the goddam tests verbose.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ deps =
 
 commands =
     test: coverage erase
-    test: coverage run -m pytest {posargs:pydoctor}
+    test: coverage run -m pytest {posargs:pydoctor} -vv
     test: coverage report -m
 
     ; Publish coverage data on codecov.io
--- a/tox.ini
+++ b/tox.ini
@@ -35,7 +35,7 @@ deps =
 
 commands =
     test: coverage erase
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -35,7 +35,7 @@ deps =
 
 commands =
     test: coverage erase
-    test: coverage run -m pytest {posargs:pydoctor}
+    test: coverage run -m pytest {posargs:pydoctor} -vv
```
adiroiban, line 38: Maybe add is as the default...but make it possible to disable it.

```suggestion
    test: coverage run -m pytest {posargs:-vv pydoctor}
```

adiroiban (approved):
Thanks for the update.

I only have a minor comment.
I also prefer verbose runs for automated tests.

I think that you can also use `tox -e test -vv pydoctor`to send extra argument to the tox command.

I am approving it as I think that this can be merged as it is... but let me know if you prefer to add `-vv` as "default arguments", as opposed to "always arguments"

tristanlatr:
Thanks!

adiroiban:
Thanks for the update. I have merged it. I hope @mthuurne  is also happy with the update :)

PR #399, opened 2021-06-07
Add Python3.9 to tox envs

tristanlatr:


adiroiban (approved):
All good. Thanks! 

I guess that this is here to help with local development.

I see that the main branch already  runs py3.9 tests https://github.com/twisted/pydoctor/runs/2712437807

PR #400, opened 2021-06-09
Use Dict[str, Documentable] as annotation for Documentable.contents.

tristanlatr:
Adjust the rest of the annotation in the codebase.

Add assertion when needed, like before reparent() call to assert the parent is a module.

Fix a potentially problematic None value in the Documentable.kind property in format_undocumented().

Change the unwrap() function in the test_astbuilder module to accept Optional[ParsedDocstring] but actually raises if the parsed_docstring is not an ParsedEpytextDocstring. This avoids a lot of extra cast() call or assertions in test_astbuilder.

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -366,6 +366,8 @@ class ModuleVistor(ast.NodeVisitor):
                             "astbuilder",
                             "moving %r into %r" % (ob.fullName(), current.fullName())
                             )
+                        # Must be a Module since the exports is set to an empty list if it's not.
+                        assert isinstance(current, model.Module)
                         ob.reparent(current, asname)
                         continue
 
@@ -470,7 +472,8 @@ class ModuleVistor(ast.NodeVisitor):
         assert isinstance(cls, model.Class)
         if not _maybeAttribute(cls, name):
             return
-        obj: Optional[model.Attribute] = cls.contents.get(name)
+        # Class variables can only be Attribute, so it's OK to cast
+        obj = cast(Optional[model.Attribute], cls.contents.get(name))
         if obj is None:
             obj = self.builder.addAttribute(name=name, kind=None, parent=cls)
         if obj.kind is None:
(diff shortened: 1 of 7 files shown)


tristanlatr:
It should be quick to review @adiroiban :)

pydoctor/astbuilder.py
```diff
@@ -470,7 +472,8 @@ def _handleClassVar(self,
         assert isinstance(cls, model.Class)
         if not _maybeAttribute(cls, name):
             return
-        obj: Optional[model.Attribute] = cls.contents.get(name)
+        # Class variables can only be Attribute, so it's OK to cast
+        obj = cast(Optional[model.Attribute], cls.contents.get(name))
```
tristanlatr, line 476: Should we use `assert` here too ?

pydoctor/astbuilder.py
```diff
@@ -504,8 +507,9 @@ def _handleInstanceVar(self,
             return
         if not _maybeAttribute(cls, name):
             return
-        obj = cls.contents.get(name)
-        if obj is None:
+        # Class variables can only be Attribute, so it's OK to cast
+        obj = cast(Optional[model.Attribute], cls.contents.get(name))
```
tristanlatr, line 511: Should we use `assert` here too ?

tristanlatr:
I would be tempted to merge this PR under the advice of @adiroiban (in #419). Let’s note that this PR makes really minor changes and would be pretty quick to review.

adiroiban:
I think that it can be reviewed... the tests are green.

if something goes wrong, we will receive error reports.

I think that with pydoctor we can be a bit more flexible. This is not a tool that is deployed in production and with which end-users are interacting.

I guess there might be such a scenario... but I don't think is that common.

Thanks @tristanlatr for all your help! Much appreciated

tristanlatr:
> I think that it can be reviewed

You mean merged ? If so, please leave an approval on the PR. Don’t want to go against the twisted protocol without a clear consensus of what should be done instead.

pydoctor/astbuilder.py
wsanchez: I believe you should be able to assert before this line, and then the cast should be unnecessary, because mypy tries to narrow types when `isinstance` assertions are made.

tristanlatr:
Thanks for the review @wsanchez

PR #402, opened 2021-06-16
Constant values and decorators colorizing

### LINKED ISSUE #377: Representation of constants values

### LINKED ISSUE #385: Separate constants from variables

tristanlatr:
Fixes #377. 

This PR introduces a new object from Epydoc, with major modifications to support a variety of AST expressions. 

The `_AnnotationFormatter` class has been completely removed and replaced with the more powerfull `PyvalColorizer`, `_ValueFormatter` uses the `PyvalColorizer` under hood, and present a colorized HTML string with links when possible.

And all of this is pretty well tested!

Check it out here: https://pydoctor--402.org.readthedocs.build/en/402/docformat/restructuredtext/restructuredtext_demo.constants.html

Example of decorators colorizing: https://pydoctor--402.org.readthedocs.build/en/402/api/pydoctor.test.test_sphinx.html#test_prepareCache

The colorizer docs: https://pydoctor--402.org.readthedocs.build/en/402/api/pydoctor.epydoc.markup._pyval_repr.PyvalColorizer.html?private=1#colorize


tristanlatr:
This PR is marked as draft until #386 is merged

tristanlatr:
This PR should be split in two: 

- First we import the `PyvalColorizer` and remove the `_AnnotationFormatter` class and make `_ValueFormatter` uses the `PyvalColorizer` under hood. 
- Then we introduce constant values rendering.

tristanlatr:
> First we import the PyvalColorizer and remove the _AnnotationFormatter class and make _ValueFormatter uses the PyvalColorizer under hood.

This first step is the big one since the pyval colorizer file is about 1000 new lines of code, and about the same for the tests. The pyval colorizer itself is type agnostic, meaning that Any kind of objects can be passed to the colorizer and it should produce an output that makes some sens in any case. I've added the AST traversing feature on top of what was already existing in the Epydoc code, the colorizer already had standard types and regex colorizing support. I find the result very convenient and flexible enough for our usage, meaning, almost everywhere: annotations, parameters defaults values and constant values. 

As a reference, here is the colorizer code: https://github.com/twisted/pydoctor/blob/d2e0b386bcd06d9e8a2478853f9398c6b434b390/pydoctor/epydoc/markup/_pyval_repr.py

I'll work on splitting this PR once #386 merged

tristanlatr:
For links to be red in function signature, they should be incorporated into a ``<code>`` tag. 

by default 2021-10-30 at 7 52 32 PM

tristanlatr:
I still need to add an entry to the changelog.

setup.cfg
```diff
@@ -40,6 +40,8 @@ install_requires =
     astor
     attrs
     docutils
+    sre_parse36
+    typing_extensions; python_version < "3.8"
```
tristanlatr, line 44: It would be best not to have typing_extensions module as a runtime dependency but only for the type checking. This means importing typing_extensions.Final inside a if TYPE_CHECKING block and use string annotation. Pydoctor should support that already.

docs/restructuredtext_demo/demo_restructuredtext_module.py
```diff
@@ -46,13 +52,13 @@ def demo_consolidated_fields(a:float, b):  # type: ignore
...
-def demo_typing_arguments(name: str, size: bytes) -> bool:
+def demo_typing_arguments(name: str, size: Optional[bytes] = None) -> bool:
     """
     Type documentation can be extracted from standard Python type hints.
 
     :param name: The human readable name for something.
-    :param size: How big the name should be.
-    :return: Always `True`.
+    :param size: How big the name should be. Leave none if you don't care.
+    :return: Always C{True}.
```
adiroiban, line 61: is this RST or epydoc?

pydoctor/driver.py
```diff
@@ -254,6 +254,9 @@ def verbose_about_callback(option: Option, opt_str: str, value: str, parser: Opt
         default='1d',
         help=MAX_AGE_HELP,
     )
+    parser.add_option(
+        '--pyval-repr-maxlines', dest='pyvalreprmaxlines', default=7, type=int,
+        help='Maxinum number of lines for a python value representation.')
```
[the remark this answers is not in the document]
adiroiban: I don't know what this means ? Is there a separate documentation for this CLI option ?

`Maxinum number of lines for a python value representation` is what I can extract from `pyval-repr-maxline` ... but I think the documentation should explain why I should care for this feature and when I might want to use it.

as a general rule, if you need to leave a PR comment to explain an end user interface / text / gui , I think that this is a sign that the existing implementation / docs  / text is not ok :)

For example, your comment from the PR might be a better text than "Maxinum number of lines for a python value representation." :)

And mabye keep it short to just `--repr-maxlines` ... do we have repr for non python values ?

README.rst
```diff
@@ -73,6 +73,7 @@ What's New?
 in development
 ~~~~~~~~~~~~~~
 
+* Colorize constant values (new), decorators, parameters default values and annotations with new colorizer adapted from epydoc. This enables to link to name locations automatically.
```
adiroiban, line 76: From this text, it is not clear what is the new functionality or why I should be excited about this change?

It this an internal refactoring ?

Does this provide any new functionality to end users?

docs/restructuredtext_demo/demo_restructuredtext_module.py
tristanlatr: Thanks, it should be RST.

README.rst
tristanlatr: > From this text, it is not clear what is the new functionality or why I should be excited about this change?

Please advise another formulation for the changelog, I would be glad to change it.

> It this an internal refactoring ?

There a bit of refactoring, please read the main text presenting the pr: https://github.com/twisted/pydoctor/pull/402#issue-670858879

> Does this provide any new functionality to end users?

Example of constant value colorizing: https://pydoctor--402.org.readthedocs.build/en/402/docformat/restructuredtext/restructuredtext_demo.constants.html

Example of decorators colorizing: https://pydoctor--402.org.readthedocs.build/en/402/api/pydoctor.test.test_sphinx.html#test_prepareCache

adiroiban: I think that we don't need to include information about refactoring inside the public changelog.

But the changelog might include information that there is now a new command line option `--pyval-repr-maxlines` that allows you to restrict/configure the number of lines for on output. 

-----------

I guess that the changelog can contain multiple items, as this PR add a lot of new features

* the availability of --pyval-repr-maxlines 
* Syntax highlight for constant values
* Embedded documentation links inside the value of a constant

docs/epytext_demo/constants.py
```diff
@@ -0,0 +1,49 @@
...
+        (?:\n           [ ]*  \.\.\. .*)*   # PS2 lines
+        \n?)
+    # Want consists of any non-blank lines that do not start with PS1.
+    (?P<want> (?:(?![ ]*$)    # Not a blank line
+                 (?![ ]*>>>)  # Not a line starting with PS1
+                 .*$\n?       # But any other line
+              )*)
+    ''', re.MULTILINE | re.VERBOSE)
+"""
+Multiline regex patterns are rendered as string.
```
adiroiban, line 48: Only regex multi-line, or any multi line text ?

docs/epytext_demo/constants.py
```diff
@@ -0,0 +1,49 @@
...
+
+import re
+from .demo_epytext_module import demo_fields_docstring_arguments, _PrivateClass
+
+A_DICT = {'1':33, '2':[1,2,3,{7:'oo'*20}], 
+         '3': demo_fields_docstring_arguments, 
+         '4': _PrivateClass.method_inside_private, 
+         '5': re.compile('^<(?P<descr>.*) at (?P<addr>0x[0-9a-f]+)>$') }
+"""
+Pydoctor does a pretty good job at analyzing constant values ;-)
```
adiroiban, line 13: Can you use this docstring to describe the exemplified features.

Like

```
The value of a constant is rendered with syntax highlighting.
Internal and external links are generated to references of classes/methods used inside the constant
```

docs/epytext_demo/constants.py
```diff
@@ -0,0 +1,49 @@
...
+
+A_LIST = [1,2,[5,6,[(11,22,33),9],10],11]+[99,98,97,96,95]
+
+ANOTHER_LIST = list(range(100))
+
+RE_STR = re.compile("(foo (?P<a>bar) | (?P<boop>baz))")
+
+RE_RAW_STR = re.compile(r"abc_raw \t\r\n\f\v \xff \uffff")
+
+RE_WITH_UNICODE = re.compile("abc 😀")
```
adiroiban, line 24: Maybe add a docstring that unicode is supported.

it would be nice to have a docstring for each of these exemplified constants to describe the presented functionality.

For example, I am not sure what is the different between `RE_RAW_STR` and `RE_RAW_STR2`

Also,  I see `A_LIST` which I guess it exemplified that you can have nested lists.
And `ANOTHER_LIST` which exemplified that you can have generated or constructed lists.

docs/epytext_demo/constants.py
tristanlatr: Not sure to understand what you mean.
But multiline strings are always represented in triple quotes, this behaviour is inherited from epydoc.

adiroiban (approved):
Sorry for the delay.
I think that this can be merged.
The end result looks good and the changes are documented and tested.

If we found any issues, we can always fix them later.

Thanks again for this great work!

tristanlatr:
Thanks @adiroiban

PR #403, opened 2021-06-16
Fix most mypy errors.

tristanlatr:


tristanlatr:
Opinions on the two issues @adiroiban @mthuurne ?
```
pydoctor/epydoc/markup/restructuredtext.py:55: error: Module "docutils.parsers.rst" has no attribute "directives"; maybe "Directive"?  [attr-defined]
pydoctor/epydoc/markup/restructuredtext.py:395: error: Argument 1 to "OptionParser" has incompatible type "List[Any]"; expected "Optional[str]"  [arg-type]
```

mthuurne:
Sorry for my slow responses; it's been a busy week and we had a bit of a heatwave here. I'll have a look now.

mthuurne:
> ```
> pydoctor/epydoc/markup/restructuredtext.py:55: error: Module "docutils.parsers.rst" has no attribute "directives"; maybe "Directive"?  [attr-defined]
> pydoctor/epydoc/markup/restructuredtext.py:395: error: Argument 1 to "OptionParser" has incompatible type "List[Any]"; expected "Optional[str]"  [arg-type]
> ```

The first error is weird: I can reproduce it under `tox`, but in my manually maintained venv it doesn't occur, even if I'm using the same version of `mypy`, `docutils`, and `types-docutils`.

The second error I can reproduce in both setups.

tristanlatr:
Ok. So let’s ignore them ?

mthuurne:
> Ok. So let’s ignore them ?

I want to dig a bit deeper first.

mthuurne:
The problem with `OptionParser` seems to be that `docutils.frontend.OptionParser` inherits from `optparse.OptionParser`, but has a different `__init__()` signature, which is missing from the type stubs. Therefore this one can be ignored indeed, with a link to https://github.com/python/typeshed/issues/5667 so we know what is happening and when we can remove the suppression.

mthuurne:
`docutils.parsers.rst.directives` is a module, which is present in `docutils` itself but missing from `typing-docutils`. So this is another case of the `docutils` stubs being incomplete.

I still don't understand why I don't get the error in my dev venv: I thought the reason might be because I have `docutils` installed, but the `mypy` tox environment also installs `docutils` since it's a dependency of `sphinx`, which we depend on since they ship type annotations in their main distribution itself.

So ignoring the error would be the correct short-term solution here as well. You could link to https://github.com/python/typeshed/issues/1269 as documentation of the problem.

pydoctor/model.py
```diff
@@ -808,6 +808,7 @@ def introspectModule(self,
             module_full_name = f'{package.fullName()}.{module_name}'
 
         spec = importlib.util.spec_from_file_location(module_full_name, path)
+        assert spec is not None, f"Cannot find spec for module {module_full_name} at {path}"
```
mthuurne, line 811: We should only use `assert` if we can we guarantee that the condition is always true. If we're not that sure, it would be better to raise an exception instead.

I think there is at least one unlikely but possible scenario in which this condition is not true: if a file is deleted or moved between the moment its name is discovered and the moment it is introspected.

pydoctor/templatewriter/__init__.py
```diff
@@ -27,7 +27,7 @@ def parse_xml(text: str) -> minidom.Document:
     Create a L{minidom} representaton of the XML string.
     """
     try:
-        return minidom.parseString(text)
+        return cast(minidom.Document, minidom.parseString(text))
```
mthuurne, line 30: I agree with the cast as a temporary solution, but we should also submit a PR to `typeshed` to add a return type for `parseString()`.

mthuurne (changes requested):

pydoctor/model.py
tristanlatr: A RuntimeError is OK ?

tristanlatr:
> You could link to python/typeshed#1269 as documentation of the problem.

I've tried using the [docutils-stubs](https://pypi.org/project/docutils-stubs/) as mentioned in this ticket but it's becoming a real headache, more than 50 errors that require a lot of additional checks in the restructuredtext module. I don't want to fix all mypy errors honestly. I will revert back to use `types-docutils` for the principal `mypy` env and create a separate one that uses `docutils-stubs` so we can look at it later. 

Also the HTMLTranslator class has been moved to pydoctor.node2stan._PydoctorHTMLTranslator in #386

tox.ini
```diff
@@ -104,21 +104,44 @@ commands =
...
 
 commands =
     mypy                                       \
         --cache-dir="{toxworkdir}/mypy_cache"  \
         {tty:--pretty:}                        \
         {posargs:pydoctor docs/epytext_demo}
 
+[testenv:mypy-docutils-stubs]
+description = run mypy with docutils-stubs (does not pass for now)
+; See: https://github.com/python/typeshed/issues/1269
```
glyph, line 126: Do we want an issue for this repo as well, to track turning it on in CI when this is done?

pydoctor/model.py
```diff
@@ -808,6 +808,8 @@ def introspectModule(self,
             module_full_name = f'{package.fullName()}.{module_name}'
 
         spec = importlib.util.spec_from_file_location(module_full_name, path)
+        if spec is None: 
+            raise RuntimeError(f"Cannot find spec for module {module_full_name} at {path}")
```
glyph, line 812: Can you do a test for this apparently untested case in a follow-up?

glyph (approved):
🚢

pydoctor/model.py
tristanlatr: Like @mthuurne was saying, this error is raised when  a file is deleted or moved between the moment its name is discovered and the moment it is introspected. So honestly I don't know how to test such a case.

tox.ini
tristanlatr: See https://github.com/twisted/pydoctor/issues/409

tristanlatr:
Thank @glyph for your review. 

I'll let @mthuurne merge this PR. 

@glyph, If you want to help out reviewing more PRs, that would be great.  
There is #386 waiting for a review  ;)

glyph:
> @glyph, If you want to help out reviewing more PRs, that would be great.

I'd love to, but WYSIWYG in this case — I already have a recurring task in my personal task management system to review open source PRs as frequently as I can, and … this is about it ;)

tristanlatr:
@mthuurne Can I merge that ? Thanks

tristanlatr:
@mthuurne I will merge this PR now since it's blocking the CI from being green on all other branches. 
If there is anything you'd like to be modified, please speak up and I'll submit a patch.

pydoctor/model.py
glyph: This is why `FilePath` narrows down its interface to a single object which returns other objects, so you can pass a parameter along to the system under test. In the absence of such compositional design, the way to test this is to mock out `spec_from_file_location` with `patch` in a test.

Issue #406, opened 2021-06-20
Handle overloaded functions


tristanlatr:
Overload signatures do not currently show up in the  output.

The `signature` attribute of the Function class could simply de transformed to `signatures` containing a list of ast expressions.

mthuurne:
Having `signatures` as a list sounds like a good idea.

Should we support per-signature docstrings?

We have the following overloaded function in `pydoctor/epydoc/markup/epytext.py`:
```py
@overload
def parse(text: str) -> Element: ...

@overload
def parse(text: str, errors: List[ParseError]) -> Optional[Element]: ...

def parse(text: str, errors: Optional[List[ParseError]] = None) -> Optional[Element]:
    """
    Return a DOM tree encoding the contents of an epytext string.  Any
    errors generated during parsing will be stored in C{errors}.

    @param text: The epytext string to parse.
    @param errors: A list where any errors generated during parsing
        will be stored.  If no list is specified, then fatal errors
        will generate exceptions, and non-fatal errors will be
        ignored.
    @return: a DOM tree encoding the contents of an epytext string,
        or C{None} if fatal errors were encountered and an C{errors}
        accumulator was provided.
    @raise ParseError: If C{errors} is C{None} and an error is
        encountered while parsing.
    """
    # (implementation)
```
Note that the `@return` clause is different from the version in Git; I'll submit a PR to fix the current docstring.

This docstring is complicated by the fact that each of the overloaded variants has a slightly different behavior. Perhaps it would be clearer if they could be documented separately:
```py
@overload
def parse(text: str) -> Element: ...
    """
    Return a DOM tree encoding the contents of an epytext string.

    @param text: The epytext string to parse.
    @return: a DOM tree encoding the contents of an epytext string.
    @raise ParseError: If a fatal error is encountered while parsing.
        Non-fatal errors are ignored.
    """"

@overload
def parse(text: str, errors: List[ParseError]) -> Optional[Element]: ...
    """
    Return a DOM tree encoding the contents of an epytext string.  Any
    errors generated during parsing will be stored in C{errors}.

    @param text: The epytext string to parse.
    @param errors: A list where any errors (fatal and non-fatal) generated
        during parsing will be stored.
    @return: a DOM tree encoding the contents of an epytext string,
        or C{None} if fatal errors were encountered.
    """

def parse(text: str, errors: Optional[List[ParseError]] = None) -> Optional[Element]:
    # (implementation)
```
[...]

PR #407, opened 2021-06-20
Remove overloading of epytext parse() function

mthuurne:
In https://github.com/twisted/pydoctor/issues/406#issuecomment-864608911 I promised a PR to fix the docstring. However, after checking the callers of `parse()`, it turns out that the one-argument version of this function was called only by the tests. Therefore I removed that overloaded variant and I moved the code that raises the first exception into the test helper.

This PR won't pass CI until #403 is solved, but I thought it would be useful to create it already, so I don't forget to submit it.

### DIFF
--- a/pydoctor/epydoc/markup/epytext.py
+++ b/pydoctor/epydoc/markup/epytext.py
@@ -131,7 +131,7 @@ __docformat__ = 'epytext en'
 #   4. helpers
 #   5. testing
 
-from typing import Any, List, Optional, Sequence, Union, cast, overload
+from typing import Any, List, Optional, Sequence, Union, cast
 import re
 
 from twisted.web.template import CharRef, Tag, tags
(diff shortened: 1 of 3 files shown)


tristanlatr:
I don’t think it’s the right thing to do.
Please look at https://github.com/twisted/pydoctor/pull/386/commits/05b842f888237db3661b6dc59ceb07809d705480

mthuurne:
> I don’t think it’s the right thing to do.
> Please look at [05b842f](https://github.com/twisted/pydoctor/commit/05b842f888237db3661b6dc59ceb07809d705480) (I’ve improved the implementation afterwards to raise the fatal error, not the first one, bu that commit contains the core of the fix)

Raising the first fatal error is indeed better than raising the first error. However, for a test helper, I'm not sure it's worth the effort if none of our tests require it.

> Not raising the error is a a bug imo.

Raising the error leads to simpler code, but it goes against the principle that we'd like to report as many errors as we can in a single run of pydoctor. If we raise the first fatal error, we discard any other errors on the same docstring.

If it is likely that most errors after the first are bogus, that could be an argument to only report the first error. I don't think that is typically the case for docstring parsing though.

---

A possible alternative would be to pass errors to a logger instead of storing them in a list. In that approach, we can raise an exception on a fatal error without losing the details of other errors. That is something we could do in a follow-up PR though: it should be easier to implement after removing the overloading.

In a different project, I implemented [a context manager on a logger](https://github.com/mthuurne/retroasm/blob/master/src/retroasm/linereader.py#L317) to raise an exception if any errors were logged. The code would look like this:
```py
def parse(text: str, logger: CustomLogger) -> Element:
    with logger.raise_on_errors():
        # Parse while logging warnings and errors to logger.
```
Or perhaps it's cleaner if we don't make the context manager part of the logger class:
```py
def parse(text: str, logger: Logger) -> Element:
    with raise_on_logged_errors(logger) as checked_logger:
        # Parse while logging warnings and errors to checked_logger.
```

tristanlatr:
As far as I understand, if there is a fatal error while parsing Epytext, it will simply show nothing in the generated output, whereas when I changed the code to always raise the exception, then it would fall back to plaintext. 

I don’t understand why we need a logger here, simply reporting the errors even if the parse_docstring() functions fails seems enough to report a maximum of errors to me.

glyph:
If this has been superseded, should it be closed?

Issue #2, opened 2015-01-12
Only first value of multi-value parameter tilde-expanded


mwhudson:
Originally reported as Twisted issue 2623:

> This is due to the fact that the config file parsing code invokes os.expanduser before splitting on commas.

---

Imported from Launchpad using lp2gh.
- date created: 2009-08-06T18:50:01Z
- owner: exarkun
- the launchpad url was https://bugs.launchpad.net/bugs/410004

tristanlatr:
I think we should completely deprecate the actual config file parsing since it's really horrible.
At the same time, the `optparse` options can be refactored to use something that belongs more to the twenty-first century.

We could use https://pypi.org/project/ConfigArgParse/ to handle the fact that all options can be read from file and overridden from CLI.

tristanlatr:
I have to send a few patches to ConfigArgParse for this module to be the real deal for us. First there is https://github.com/bw2/ConfigArgParse/pull/229, there would be a few other stuff also...

Issue #23, opened 2015-01-12
"packages" .cfg field only seems able to hold one package


mwhudson:
In the absense of examples which used more than one package in the packages .cfg field, I treid both:

packages: bzr bzrlib

and:

packages: bzr, bzrlib

and neither syntax worked.

---

Imported from Launchpad using lp2gh.
- date created: 2009-04-10T19:41:20Z
- owner: naesten
- the launchpad url was https://bugs.launchpad.net/bugs/359253

mwhudson:
(by mwhudson)
Maybe "bzr,bzrlib" will work?

The config file/argument parsing stuff in pydoctor is pretty terrible.  I should use something like configglue instead.

PR #411, opened 2021-06-23
Use to ConfigArgparse to parse arguments and load config file

### LINKED ISSUE #410: Replace `optparse`
I was thinking maybe using https://github.com/bw2/ConfigArgParse . 
That would allow to configure the optional arguments with a config file, like `pydoctor.ini` for instance, then the positionals arguments should still be passed to `pydoctor`

Tell me what you think @mthuurne, thanks.

tristanlatr:
Fixes #410 

This PR kills the optparse dependency and replace it with ConfigArgparse.

This adds a new file to load config from: `pydoctor.ini`.  I've asked if the config file could be integrated to other "standards"
 files (like `setup.cfg`) here https://github.com/bw2/ConfigArgParse/issues/227, but it seems that's there is not plan to support that.

This also fixes issues #2 and #23 in the sense that the positional arguments cannot be set from config file with configargparse, so that solves the issues!

I still need to work on adding some documentation, so this PR as marked as draft (also until #403 is merged)


tristanlatr:
Also, I think I will delete the help output extension and simply use https://sphinx-argparse.readthedocs.io/en/stable/index.html

tristanlatr:
We might want to wait until https://github.com/bw2/ConfigArgParse/pull/232 and https://github.com/bw2/ConfigArgParse/pull/229 are merged to ConfigArgParse/master branch and a pypi release is done in order to use these options in pydoctor!

tristanlatr:
Just one thing also, I needed to remove the option `--verbose-about` the verbose details stuff. I feel like it's not useful to be able to choose such a option when we don't even know what all the stages of the API doc generation. 

Also it was harder to make the current code work with argparse the way it was handled, so I choose to remove the option alltogether.

Issue #313, opened 2020-12-04
Include documented aliases in output


mthuurne:
From `src/twisted/words/protocols/jabber/client.py`: (irrelevant parts cut out)
```py
class BasicAuthenticator(xmlstream.ConnectAuthenticator):
    """
    Authenticates an XmlStream against a Jabber server as a Client.

    @cvar INVALID_USER_EVENT: See L{IQAuthInitializer.INVALID_USER_EVENT}.
    @type INVALID_USER_EVENT: L{str}

    @cvar AUTH_FAILED_EVENT: See L{IQAuthInitializer.AUTH_FAILED_EVENT}.
    @type AUTH_FAILED_EVENT: L{str}

    @cvar REGISTER_FAILED_EVENT: Token to signal that registration failed.
    @type REGISTER_FAILED_EVENT: L{str}
    """

    INVALID_USER_EVENT = IQAuthInitializer.INVALID_USER_EVENT
    AUTH_FAILED_EVENT = IQAuthInitializer.AUTH_FAILED_EVENT
    REGISTER_FAILED_EVENT = "//event/client/basicauth/registerfailed"
```
Currently pydoctor will handle `INVALID_USER_EVENT` and `AUTH_FAILED_EVENT` as aliases, which means that it keeps track of the indirections, but doesn't include them in the output. However, as they are documented names, they should be given documentation entries.

From `src/twisted/python/constants.py`:
```py
# Import and re-export Constantly
from constantly import NamedConstant, ValueConstant, FlagConstant, Names, Values, Flags

__all__ = ["NamedConstant", "ValueConstant", "FlagConstant", "Names", "Values", "Flags"]
```
Since these names are explicitly exported via `__all__`, they should end up in the output. But currently the documentation for `twisted.python.constants` is empty.

So I think aliases should sometimes be Documentables.

tristanlatr:
I'de like to work on this issue. This is a major thing that pydoctor does not do yet.

I'm just not sure about skipping undocumented aliases tho ? 
I think alias should be always given a documentation entry, even when not documented.

tristanlatr:
To clarify the difference between aliases and constants, we could do the following: 
- create the new alias kind
- create a `is_alias(obj) -> bool` function and change `ModuleVistor._handleAliasing()` method such that, instead of hacking the redirection with `ctx._localNameToFullName_map[target] = full_name`, it creates a legit Attribute documentable and it returns `None`, like other `_handle*` methods. 
[...]

PR #414, opened 2021-06-26
Aliases

tristanlatr:
Add aliases to the output and add them the the system with a special kind: `ALIAS`.  

See #313 for details. 

Screen Shot 2021-06-26 at 2 33 46 AM

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -46,25 +46,7 @@ def _maybeAttribute(cls: model.Class, name: str) -> bool:
         inherited) attribute, L{False} otherwise
     """
     obj = cls.find(name)
-    return obj is None or isinstance(obj, model.Attribute)
-
-
-def _handleAliasing(
-        ctx: model.CanContainImportsDocumentable,
-        target: str,
-        expr: Optional[ast.expr]
-        ) -> bool:
-    """If the given expression is a name assigned to a target that is not yet
-    in use, create an alias.
-    @return: L{True} iff an alias was created.
-    """
-    if target in ctx.contents:
-        return False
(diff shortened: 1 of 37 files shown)


tristanlatr:
Should undocumented aliases be shown in output ? I current implementation does, maybe we should think about it twice, because it cause create a lot of noise in the docs.

pydoctor/astbuilder.py
```diff
@@ -194,13 +176,18 @@ def extract_final_subscript(annotation: ast.Subscript) -> ast.expr:
...
+
+
 class ModuleVistor(NodeVisitor):
 
     def __init__(self, builder: 'ASTBuilder', module: model.Module):
         super().__init__()
         self.builder = builder
         self.system = builder.system
         self.module = module
+        self._moduleLevelAssigns: List[str] = []
```
tristanlatr, line 190: This looks like dead code, I'll remove it.

pydoctor/model.py
```diff
@@ -387,7 +479,23 @@ def report(self, descr: str, section: str = 'parsing', lineno_offset: int = 0) -
...
+        @note: It seems that the list is not always complete, though.
+        """
+        aliases: List['Attribute'] = []
+        for alias in filter(lambda ob: ob.kind is DocumentableKind.ALIAS and isinstance(ob, Attribute), 
+                         self.system.allobjects.values()):
+            assert isinstance(alias, Attribute)
+            assert alias.parent is not None
+            if alias.parent._resolveDocumentable(alias) == self.fullName():
+                aliases.append(alias)
+        return aliases
```
tristanlatr, line 497: This is not the right implementation, the aliases should be computed at post-processing time.

PR #417, opened 2021-06-30
@overload decorator support

tristanlatr:
Add `@overload` decorator support.

Also, it adds a clear `Function.return_type` attribute.

And some documentation to the ASTBuilder!

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -642,19 +642,26 @@ class ModuleVistor(ast.NodeVisitor):
             return
 
         lineno = node.lineno
+
+        # setting linenumber from the start of the decorations
         if node.decorator_list:
             lineno = node.decorator_list[0].lineno
 
+        # extracting docstring
         docstring: Optional[ast.Str] = None
         if len(node.body) > 0 and isinstance(node.body[0], ast.Expr) \
                               and isinstance(node.body[0].value, ast.Str):
             docstring = node.body[0].value
-
+        
         func_name = node.name
+
+        # determine the function's kind
         is_property = False
         is_classmethod = False
         is_staticmethod = False
-        if isinstance(parent, model.Class) and node.decorator_list:
+        is_overload_func = False
+        if node.decorator_list:
+            # analyze decorators
             for d in node.decorator_list:
                 if isinstance(d, ast.Call):
                     deco_name = node2dottedname(d.func)
@@ -662,26 +669,42 @@ class ModuleVistor(ast.NodeVisitor):
                     deco_name = node2dottedname(d)
                 if deco_name is None:
                     continue
-                if deco_name[-1].endswith('property') or deco_name[-1].endswith('Property'):
-                    is_property = True
-                elif deco_name == ['classmethod']:
-                    is_classmethod = True
-                elif deco_name == ['staticmethod']:
-                    is_staticmethod = True
-                elif len(deco_name) >= 2 and deco_name[-1] in ('setter', 'deleter'):
-                    # Rename the setter/deleter, so it doesn't replace
-                    # the property object.
-                    func_name = '.'.join(deco_name[-2:])
+                if isinstance(parent, model.Class):
+                    if deco_name[-1].endswith('property') or deco_name[-1].endswith('Property'):
+                        is_property = True
+                    elif deco_name == ['classmethod']:
+                        is_classmethod = True
+                    elif deco_name == ['staticmethod']:
+                        is_staticmethod = True
+                    elif len(deco_name) >= 2 and deco_name[-1] in ('setter', 'deleter'):
+                        # Rename the setter/deleter, so it doesn't replace
+                        # the property object.
+                        func_name = '.'.join(deco_name[-2:])
+                if deco_name[-1] == 'overload':
+                    # determine if the function is decorated with overload
+                    is_overload_func = True
 
         if is_property:
+            # handle property and return
             attr = self._handlePropertyDef(node, docstring, lineno)
             if is_classmethod:
                 attr.report(f'{attr.fullName()} is both property and classmethod')
             if is_staticmethod:
                 attr.report(f'{attr.fullName()} is both property and staticmethod')
             return
-
-        func = self.builder.pushFunction(func_name, lineno)
+        
+        # Check if it's a new func
+        func = parent.contents.get(func_name)
+        if func is not None and any(
+            'overload' in node2dottedname(deco) or () for 
+            deco in func.decorators or ()):
+            # If the function is annottated with overload right before:
+            # do not re create a Function object, simply enter it again.
+            self.builder.push(func, lineno)
+        else:
+            # new function
+            func = self.builder.pushFunction(func_name, lineno)        
+        
         func.is_async = is_async
         if docstring is not None:
             func.setDocstring(docstring)
@@ -700,6 +723,8 @@ class ModuleVistor(ast.NodeVisitor):
         num_pos_args = len(posonlyargs) + len(node.args.args)
         defaults = node.args.defaults
         default_offset = num_pos_args - len(defaults)
+        annotations, return_type = self._annotations_from_function(node)
+
         def get_default(index: int) -> Optional[ast.expr]:
             assert 0 <= index < num_pos_args, index
             index -= default_offset
@@ -708,7 +733,11 @@ class ModuleVistor(ast.NodeVisitor):
         parameters = []
         def add_arg(name: str, kind: Any, default: Optional[ast.expr]) -> None:
             default_val = Parameter.empty if default is None else _ValueFormatter(default)
-            parameters.append(Parameter(name, kind, default=default_val))
+            # pick up annotations in the functions sigantures, this will link to names when the pyval_repr PR is merged (https://github.com/twisted/pydoctor/pull/402)
+            annotation = Parameter.empty if name not in annotations or annotations[name] is None else _ValueFormatter(annotations[name])
+            parameters.append(Parameter(name, kind, 
+                                        default=default_val, 
+                                        annotation=annotation))
 
         for index, arg in enumerate(posonlyargs):
             add_arg(arg.arg, Parameter.POSITIONAL_ONLY, get_default(index))
@@ -728,14 +757,25 @@ class ModuleVistor(ast.NodeVisitor):
         if kwarg is not None:
             add_arg(kwarg.arg, Parameter.VAR_KEYWORD, None)
 
+        # pick up return-type annotations, ignore "-> None" annotations.
+        return_annotation = Parameter.empty if return_type is None or epydoc2stan.is_none_literal(return_type) else _ValueFormatter(return_type)
+
         try:
-            signature = Signature(parameters)
+            signature = Signature(parameters, return_annotation=return_annotation)
         except ValueError as ex:
             func.report(f'{func.fullName()} has invalid parameters: {ex}')
             signature = Signature()
 
         func.signature = signature
-        func.annotations = self._annotations_from_function(node)
+        func.annotations = annotations
+        func.return_type = return_type
+
+        # Handle overloaded functions
+        if is_overload_func:
+            func.overloads.append(
+                self.system.FunctionOverload(
+                    signature=func.signature))
+        
         self.default(node)
         self.builder.popFunction()
 
@@ -796,12 +836,13 @@ class ModuleVistor(ast.NodeVisitor):
 
     def _annotations_from_function(
             self, func: Union[ast.AsyncFunctionDef, ast.FunctionDef]
-            ) -> Mapping[str, Optional[ast.expr]]:
+            ) -> Tuple[Mapping[str, Optional[ast.expr]], Optional[ast.expr]]:
         """Get annotations from a function definition.
         @param func: The function definition's AST.
-        @return: Mapping from argument name to annotation.
-            The name C{return} is used for the return type.
-            Unannotated arguments are omitted.
+        @return: 2-tuple:
+            - Mapping from argument name to annotation.
+              Unannotated arguments are omitted.
+            - The return type annotation.
         """
         def _get_all_args() -> Iterator[ast.arg]:
             base_args = func.args
@@ -821,16 +862,15 @@ class ModuleVistor(ast.NodeVisitor):
         def _get_all_ast_annotations() -> Iterator[Tuple[str, Optional[ast.expr]]]:
             for arg in _get_all_args():
                 yield arg.arg, arg.annotation
-            returns = func.returns
-            if returns:
-                yield 'return', returns
-        return {
+        annotations_map = {
             # Include parameter names even if they're not annotated, so that
             # we can use the key set to know which parameters exist and warn
             # when non-existing parameters are documented.
             name: None if value is None else self._unstring_annotation(value)
             for name, value in _get_all_ast_annotations()
             }
+        
+        return annotations_map, func.returns
 
     def _unstring_annotation(self, node: ast.expr) -> ast.expr:
         """Replace all strings in the given expression by parsed versions.
@@ -982,6 +1022,9 @@ def _annotation_for_elements(sequence: Iterable[object]) -> Optional[ast.expr]:
 DocumentableT = TypeVar('DocumentableT', bound=model.Documentable)
 
 class ASTBuilder:
+    """
+    Keeps tracks of the state of the AST build, creates documentable and adds objects to the system.
+    """
     ModuleVistor = ModuleVistor
 
     def __init__(self, system: model.System):
@@ -992,6 +1035,9 @@ class ASTBuilder:
         self.ast_cache: Dict[Path, Optional[ast.Module]] = {}
 
     def _push(self, cls: Type[DocumentableT], name: str, lineno: int) -> DocumentableT:
+        """
+        Create and enter a new object of the given type and add it to the system.
+        """
         obj = cls(self.system, name, self.current)
         self.system.addObject(obj)
         self.push(obj, lineno)
@@ -1002,6 +1048,9 @@ class ASTBuilder:
         self.pop(self.current)
 
     def push(self, obj: model.Documentable, lineno: int) -> None:
+        """
+        Enter a documentable.
+        """
         self._stack.append(self.current)
         self.current = obj
         if isinstance(obj, model.Module):
@@ -1018,24 +1067,44 @@ class ASTBuilder:
             obj.setLineNumber(lineno)
 
     def pop(self, obj: model.Documentable) -> None:
+        """
+        Leave a documentable.
+        """
         assert self.current is obj, f"{self.current!r} is not {obj!r}"
         self.current = self._stack.pop()
         if isinstance(obj, model.Module):
             self.currentMod = None
 
     def pushClass(self, name: str, lineno: int) -> model.Class:
+        """
+        Create and a new class in the system.
+        """
         return self._push(self.system.Class, name, lineno)
+
     def popClass(self) -> None:
+        """
+        Leave a class.
+        """
         self._pop(self.system.Class)
 
     def pushFunction(self, name: str, lineno: int) -> model.Function:
+        """
+        Create and enter a new function in the system.
+        """
         return self._push(self.system.Function, name, lineno)
+
     def popFunction(self) -> None:
+        """
+        Leave a function.
+        """
         self._pop(self.system.Function)
 
     def addAttribute(self,
             name: str, kind: Optional[model.DocumentableKind], parent: model.Documentable
             ) -> model.Attribute:
+        """
+        Add a new attribute to the system.
+        """
         system = self.system
         parentMod = self.currentMod
         attr = system.Attribute(system, name, parent)
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -376,7 +376,7 @@ class FieldHandler:
 
         self.parameter_descs: List[FieldDesc] = []
         self.return_desc: Optional[FieldDesc] = None
-        self.raise_descs: List[RaisesDesc] = []
+        self.raise_descs: List[FieldDesc] = []
         self.seealsos: List[Field] = []
         self.notes: List[Field] = []
         self.authors: List[Field] = []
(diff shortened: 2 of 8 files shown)


tristanlatr:
Feedback welcome on the proposed ideas @mthuurne @glyph . 

Thanks

mthuurne:
> Type annotations are now always added to the function signature. I realize this might be a duplication of information, but for functions with a lot of parameters, having a concise and complete view of the signature is a good thing to have.

For functions with a lot of parameters, the annotations might actually make it less concise, since the signature won't fit onto a single line then.

> Right now, it only shows the annotations coming from AST. We probably shouldn’t try to integrate the type fields values to the signatures since fields can contain any kind of markup. The duplication of the the type informations might be an issue if project are using very long types.

Type fields can contain either formal or informal types. Showing informal types in the signature wouldn't make sense, in my opinion. Showing formal types in the signature could be useful, but in that case it might be more useful if we'd provide a tool/mode for converting formal type documentation to type annotations.

> This is an example of how it renders :
> 
> Screen Shot 2021-07-04 at 9 12 31 PM

I think we should omit the final variant (the one without the `@overload` decoration in the source), since those annotations only exist to let mypy verify the implementation. The interface of the function is defined by the `@overload` definitions and we're generating interface documentation.

> * Maybe we should allow to turn-off annotations in all signatures with an option ? And turn it on automatically for overloaded signatures.
> * Another thing that could be done, and that could avoid creating a new option, is to always add annotations when present and stop listing undocumented parameters in the « Parameters » section, or list only their names in one line. This way we have a consistent solution for the signatures and we avoid duplication of information of undocumented parameters.

When it comes to options, I think we should first figure out what the default behavior will be. If there are a significant number of people who don't want that default behavior for their project, we can introduce an option for alternative behavior.

> Simply transform this
> 
> by default 2021-07-09 at 10 54 04 AM
> to this, WITH tLINKS :
> 
> Screen Shot 2021-07-04 at 9 17 12 PM
> 
> For undocumented parameters only, of course. It's more concise and still present all the information available.

I think that undocumented parameters standing out a bit is useful to remind authors to document them. We could list undocumented parameters without listing parameter types though.

> Also, this pr adds a clear `Function.return_type` attribute. And some documentation to the ASTBuilder!

Nice! However, it would be good to have such cleanups in a separate PR, so they could be reviewed and merged more easily.

---

There is some friction in the design between the signatures and the parameter and return value lists. I'll re-use your example with the final signature cut off:

image

The `Optional` for the `errors` parameter is a bit confusing, as we wouldn't actually want the user to pass `None` here: they should either pass nothing or a list. While this difference is relatively theoretical (it's  unlikely the hidden default will ever become something other than `None`), in other situations the hidden default might be a private type that we don't want to appear in the public API docs.

The `Optional` in the return type only applies to the 2-argument version of the function. In the 1-argument version, the function raises an exception instead. Therefore, the "raises" clause is exclusive to the 1-argument version of the function and that is something we currently cannot express in a clear way.

I'm wondering if it might be clearer to show the entire function twice, filtered by overloaded version, so I created a mock-up:

image

While this does take up more space in the output, many of the individual information items are easier to understand since the text no longer needs to make the distinction between an `errors` list being passed or not.

To be able to produce this output, pydoctor would have to parse docstrings on the overloaded function definitions:
```py
@overload
def parse(text: str) -> Element:
    """
    Return a DOM tree encoding the contents of an epytext string.
    Fatal errors will generate exceptions, non-fatal errors will be ignored.

    @return: a DOM tree encoding the contents of an epytext string.
    @raise ParseError: If a fatal error is encountered while parsing.
    """

@overload
def parse(text: str, errors: List[ParseError]) -> Optional[Element]:
    """
    Return a DOM tree encoding the contents of an epytext string.
    Any errors generated during parsing will be stored in C{errors}.

        will be stored.
    @return: a DOM tree encoding the contents of an epytext string,
        or C{None} if fatal errors were encountered.
    """

def parse(text: str, errors: Optional[List[ParseError]] = None) -> Optional[Element]:
    """
    @param text: The epytext string to parse.
    """
```

tristanlatr:
Hi @mthuurne,

I also thought it was better to list the whole function documentation as many time as overloads but I realize it was taking too much space for so little information. Plus the per definition docstring are not something widely use right ? So this is why I propose to include the annotation in the signature. also it would make the signature closer to the code. 

> For functions with a lot of parameters, the annotations might actually make it less concise, since the signature won't fit onto a single line then.

The signatures can already take multiple lines and I think it's OK if they do. I think it would be beneficial to add the annotation in the signatures. But I agree to skip the actual function definition when it's overloaded. 

>  provide a tool/mode for converting formal type documentation to type annotations.

Regarding the type fields processing, I think that I've implemented this feature in #358 with the new option --process-types. 

> I think that undocumented parameters standing out a bit is useful to remind authors to document them.

I disagree. sometimes developers just don't want to document parameters (only have a description). In this case seeing "undocumented" everywhere is a lot of noise. 

> I'm wondering if it might be clearer to show the entire function twice

This might makes the pages very long for not a lot of extra information. The `overload` mechanism is only related to annotations, so we should allow this usage without showing "undocumented" everywhere. 
I agree to display the whole function twice or more, but only if the overloaded definitions do have docstrings associated. Otherwise I would go with a more concise approach that simply list the signatures one after another. Regarding the types aside parameters we could compute a string like "nothing or List[ParseError]" instead of the implementation type.

As an example here is how an overloaded function shows with sphinx-autodoc:

```
@overload
def hey(s: bytes) -> bytes: ...
@overload
def hey(s: str) -> str: ...
def hey(s: Union[str, bytes]) -> Union[bytes, str]:
  """
  Say Hey.

  :param s: a string
  :Returns: Something.
  """
  pass
```

by default 2021-07-11 at 1 51 29 PM

glyph:
> > I think that undocumented parameters standing out a bit is useful to remind authors to document them.
> 
> I disagree. sometimes developers just don't want to document parameters (only have a description). In this case seeing "undocumented" everywhere is a lot of noise.

I agree with both of you a little bit :)

@mthuurne - I do think it's important to have some kind of motivation for projects, like Twisted, which wish to have an "all public functions and classes must have documentation coverage for all parameters and attributes"; this is a good motivation.

@tristanlatr - I agree that just spewing "undocumented" all over the place just makes the docs uglier and doesn't actually achieve that goal.

I am inclined to remove our habit of putting "undocumented" everywhere, and just present whatever docs we have (even if it's just a type, or the parameter's existence) and focus efforts on adding a ratchetable / failable lint step that requires documentation on all components and start enforcing that ratchet in CI (ideally with as much help as possible from pydoctor's artifacts itself).  Give the code reviewer the ability to easily notice undocumented things and request clearer docs for them.

(Really, if we could quickly and easily see a *diff* of changed documentation inline, that's the holy grail; if we could do that, then the `undocumented`'s could actually serve a purpose. But I imagine that doing this properly would be ... extremely challenging, at best.)

mthuurne:
> > > I think that undocumented parameters standing out a bit is useful to remind authors to document them.
> > 
> > 
> > I disagree. sometimes developers just don't want to document parameters (only have a description). In this case seeing "undocumented" everywhere is a lot of noise.
> 
> I agree with both of you a little bit :)
> 
> @mthuurne - I do think it's important to have some kind of motivation for projects, like Twisted, which wish to have an "all public functions and classes must have documentation coverage for all parameters and attributes"; this is a good motivation.
> 
> @tristanlatr - I agree that just spewing "undocumented" all over the place just makes the docs uglier and doesn't actually achieve that goal.
> 
> I am inclined to remove our habit of putting "undocumented" everywhere, and just present whatever docs we have (even if it's just a type, or the parameter's existence) and focus efforts on adding a ratchetable / failable lint step that requires documentation on all components and start enforcing that ratchet in CI (ideally with as much help as possible from pydoctor's artifacts itself). Give the code reviewer the ability to easily notice undocumented things and request clearer docs for them.

We could add an option to to warn about undocumented parameters (and for also classes, functions etc.) when pydoctor runs, rather than in the HTML output. We already have an option to return a non-zero exit code when there are warnings, so this could be used to fail CI runs.

> (Really, if we could quickly and easily see a _diff_ of changed documentation inline, that's the holy grail; if we could do that, then the `undocumented`'s could actually serve a purpose. But I imagine that doing this properly would be ... extremely challenging, at best.)

It would require some extra tooling, but I don't think it would be overly difficult. Inside pydoctor, we already know the definition AST node to which a docstring belongs. We take the line number for source links from this, for example. Instead of a line number, we'd need a line ranges, since definitions can span multiple lines (more often when using type annotations) and some definitions require more than one range (overloaded functions, settable properties). Then those ranges could be logged to JSON.

At the CI side, we'd have to know which line ranges are touched by a PR. I assume there are already tools for this, but if none of them are open source then it's not that hard to write a parser for diff files. I once wrote [a Python version of `patch`](https://github.com/openMSX/openMSX/blob/master/build/patch.py) without support for fuzzy matching (as fuzzy matching is actually hard to implement but not needed if you have clean diffs).

If there is an overlap between the definition line ranges of something that is (partially) undocumented and the set of lines changed in a commit, that would be flagged as a regression. It wouldn't  be 100% accurate, but in the worst case we'd force someone to document something that is at least related to the thing they were changing.

I'm not volunteering though, as I'm already way behind reviewing Tristan's PRs.

glyph:
> At the CI side, we'd have to know which line ranges are touched by a PR. I assume there are already tools for this, but if none of them are open source then it's not that hard to write a parser for diff files. I once wrote [a Python version of `patch`](https://github.com/openMSX/openMSX/blob/master/build/patch.py) without support for fuzzy matching (as fuzzy matching is actually hard to implement but not needed if you have clean diffs).

If you want to file a ticket and you think it's really tractable, being able to see e.g. an HTML preview of changed documentation inline on a PR would be *incredible* for encouraging a culture of proper documentation review.  Regarding "which line ranges are touched by a PR", I haven't dived into the implementation, but http://diff-cover.readthedocs.io/en/latest/README.html does this for coverage, and I think can deal with other tools that emit line range outputs as well.  That might be a good place to start.

mthuurne:
> > At the CI side, we'd have to know which line ranges are touched by a PR. I assume there are already tools for this, but if none of them are open source then it's not that hard to write a parser for diff files. I once wrote [a Python version of `patch`](https://github.com/openMSX/openMSX/blob/master/build/patch.py) without support for fuzzy matching (as fuzzy matching is actually hard to implement but not needed if you have clean diffs).
> 
> If you want to file a ticket and you think it's really tractable, being able to see e.g. an HTML preview of changed documentation inline on a PR would be _incredible_ for encouraging a culture of proper documentation review. Regarding "which line ranges are touched by a PR", I haven't dived into the implementation, but http://diff-cover.readthedocs.io/en/latest/README.html does this for coverage, and I think can deal with other tools that emit line range outputs as well. That might be a good place to start.

Ah, I thought you only wanted to filter the warnings about undocumented parameters and display those next to the _source diff_.

Diffing the _generated HTML_ is a step beyond that. Displaying the "before" and "after" could be done by rendering both and looking up the fully qualified name in the Intersphinx inventory, then cut the node with the right ID from the right HTML file. Doing a visual diff of HTML could be tricky, as you'd have to figure out how to not break the original layout and styling while embedding HTML inside HTML.

mthuurne:
> I also thought it was better to list the whole function documentation as many time as overloads but I realize it was taking too much space for so little information.

It does take up a lot of space, but what matters is whether it is slower or faster to find the information that the user is looking for. In the specific case of the Epydoc `parse()` function, I think that the split documentation is easier to understand because it lacks the "if called like this, it behaves like this, but if called like that, it behaves like that" conditions.

What I don't know is how representative that function is for overloading in general. I'm not a big fan of one function/class doing two different things depending on arguments, Twisted's `DeferredList` is another example of such a design that would be better off just split in two, with two separate names instead of overloading.

One potential problem we'd face with per-definition docstrings is that if we ever turn parameters into `Documentables` (to be able to link to them), we would have more than one output location for the same fully qualified name. However, linking to the first overloaded definition that includes the parameter is probably good enough in practice.

> Plus the per definition docstring are not something widely use right ? So this is why I propose to include the annotation in the signature. also it would make the signature closer to the code.

I don't recall seeing per-definition docstrings used, but if there is no tool support for it yet, that's not surprising. The question is whether they would be useful if they were supported.

I'm not sure that having the signature look similar to the code is necessarily a good thing. With type annotations, signatures can get quite complex and having them pulled apart per parameter and return type can improve readability.

> The signatures can already take multiple lines and I think it's OK if they do. I think it would be beneficial to add the annotation in the signatures.

A tool like Black will reformat multi-line function signatures to have a single parameter per line, which looks a lot like our parameters table.

> But I agree to skip the actual function definition when it's overloaded.

OK.

> > provide a tool/mode for converting formal type documentation to type annotations.
> 
> Regarding the type fields processing, I think that I've implemented this feature in #358 with the new option --process-types.

Yes, in theory all we'd need to add to that is an option to write out type stubs, which can then be applied to the source with [retype](https://pypi.org/project/retype/).

> > I think that undocumented parameters standing out a bit is useful to remind authors to document them.
> 
> I disagree. sometimes developers just don't want to document parameters (only have a description). In this case seeing "undocumented" everywhere is a lot of noise.

I guess it depends on whether those parameters will ever be documented. If they are, seeing "undocumented" serves as a reminder, but if they are not, it is indeed noise.

Would it be useful to have a convention to mark a parameter as deliberately undocumented? For example, by having the description be an empty string or for example `/`. Or would that still be too much overhead?

> > I'm wondering if it might be clearer to show the entire function twice
> 
> This might makes the pages very long for not a lot of extra information. The `overload` mechanism is only related to annotations, so we should allow this usage without showing "undocumented" everywhere.

It's not only about annotations: in the `parse()` example, whether an exception can be raised or not depends on the arguments passed, while Python doesn't have annotations for exceptions.

I think the discussion about how we handle undocumented parameters is useful, but is mostly separate from how we present overloaded functions. So I'd like that to be addressed in a separate PR.

> I agree to display the whole function twice or more, but only if the overloaded definitions do have docstrings associated.

That would allow the user to pick which presentation is better on a per-function basis, which sounds useful if we can't find a single presentation that is good for all overloaded functions. The down side is of course that we'd have to maintain both implementations.

> Otherwise I would go with a more concise approach that simply list the signatures one after another. Regarding the types aside parameters we could compute a string like "nothing or List[ParseError]" instead of the implementation type (or just let it empty like the example bellow).

I think that if we already put the more accurate per-variant types in the signature, we'd just create confusion by listing combined types elsewhere.

> As an example here is how an overloaded function shows with sphinx-autodoc:
> 
> by default 2021-07-11 at 1 51 29 PM

That looks nice, but I'm always a bit cautious about whether something that looks good on a synthetic example also applies to real code. This example doesn't actually require overloading: it could use `AnyStr` instead of `Union[bytes, str]`. Also it's missing information, such as specifying which encoding is used when `bytes` are used to represent a string. Since encoding doesn't apply to `str`, a per-variant docstring might be useful here.

I dug up a few examples from real code:
```py
@overload
def atomicWrite(path: Path, mode: Literal['w'], fsync: bool = True, **kwargs: Any) -> TextIOWrapper:
   ...

@overload
def atomicWrite(path: Path, mode: Literal['wb'], fsync: bool = True, **kwargs: Any) -> FileIO:
   ...

@contextmanager # type: ignore[misc]
def atomicWrite(path: Path, mode: str, fsync: bool = True, **kwargs: Any) -> Iterator[IO[Any]]:
    '''A context manager to write a file in such a way that in an event of
    abnormal program termination either an old version of the file remains,
    or a new one, but not something inbetween.

    (snip)

    @param path: The file path to open.
    @param mode: The mode in which the file will be opened;
        only modes C{w} and C{wb} are supported.
    @param fsync: Can be set to C{False} to not force changes to be committed
        to long-term storage; this is faster but destroys the atomicity
        guarantee. Used to speed up unit testing.
    @param kwargs: Arguments passed to the builtin L{open()} function.
    '''
```
This one could probably be handled just fine with multiple typed signatures. The signature is wider than 80 columns though, so depending on the output layout it may or may not need wrapping.

```py
@overload
def get(self, instance: object, methodName: str) -> Callable:
    ...

@overload
def get(self, instance: object, methodName: str, default: T) -> Union[Callable, T]:
    ...

def get(self, instance: object, methodName: str, default: object = _nodefault) -> object:
    """
    Retrieve an exposed method with the given name from the given instance.

    @raise UnexposedMethodError: Raised if C{default} is not specified and
    there is no exposed method with the given name.

    @return: A callable object for the named method assigned to the given
    instance.
    """
```
This is another case where the `raise` clause differs per overloaded variant. While this example isn't as complex as `parse()`, this pattern of using overloading to handle a default value seems to be quite common: 2 out of 3 uses of `@overload` in Twisted follow this pattern, while I found a few in SoftFab as well.

---

Looking at these examples, using multiple signatures is maybe not ideal but at least good enough in most cases. So I'd be fine with merging a single documentation entry solution first and treating separate entries per `@overload` definition as a separate feature.

tristanlatr:
Thanks for all this feeback. Probably we should open a few tickets to summarize what's been said into actionable features. 
I'm not quite sure what action should be taken now regarding the overload support for the short-term, though. 

> I'd be fine with merging a single documentation entry solution

- This PR adds the type annotations to ALL signatures, when present in the code. 
   Example: https://pydoctor--417.org.readthedocs.build/en/417/api/pydoctor.epydoc2stan.html#extract_fields
   Should I revert this behaviour ? Let's note that the links are not automatically handled by the `_ValueFormatter`, until #402 is merged. 
   If yes, what should be done instead ? Only display the annotations for overloaded functions ?

> Would it be useful to have a convention to mark a parameter as deliberately undocumented? For example, by having the description be an empty string or for example /. Or would that still be too much overhead?

I think that would be too much overhead for handling something that you don't want to bother documenting.

TODOS: 

- omit the final variant (the one without the @overload decoration in the source)

mthuurne:
> > I'd be fine with merging a single documentation entry solution
> 
> * This PR adds the type annotations to ALL signatures, when present in the code.
>       Example: https://pydoctor--417.org.readthedocs.build/en/417/api/pydoctor.epydoc2stan.html#extract_fields
>       Should I revert this behaviour ? Let's note that the links are not automatically handled by the `_ValueFormatter`, until [Constant values #402](https://github.com/twisted/pydoctor/pull/402) is merged.
>       If yes, what should be done instead ? Only display the annotations for overloaded functions ?

It would be more consistent to move all annotations into the signature, also for non-overloaded functions, but we'd have to figure out a new way to present informal (non-annotation) type documentation then. Maybe a line below the parameter description instead of a separate column?

On the other hand, overloaded functions are relatively rare, so maybe it's not something that should dictate how we present all other functions. If overloaded functions didn't exist, which presentation would you prefer: annotations in the signature or in the parameters table?

Maybe #402 can be split into one PR that upgrades `_AnnotationFormatter` and another PR that adds the formatting of values? That would simplify review and the first PR could be merged earlier.

> TODOS:
> 
> * omit the final variant (the one without the `@overload` decoration in the source)
> * remove the final variant annotations from the parameters and returns table, precise types are added in the overload definitions. Which are shown in the signature already.

Yes to both.

> * show `@overload` decorators on top of signatures ?

I added it for the two-entries approach, as a hint that there is another entry for the same function. I think we need something that makes it clear that a function is an overloaded; in the case of a one-entry presentation that is implicit by there being multiple signatures, but for a two-entry presentation we need some explicit clue.

That something doesn't need to literally be `@overload` though. I picked that because we also show `@property`, but I'm not convinced that is the most elegant way of showing that something is a property. I think it would be better if properties were formatted more like variables are, since for the end user it shouldn't matter much whether an attribute is implemented as a variable or as a property.

tristanlatr:
>  If overloaded functions didn't exist, which presentation would you prefer: annotations in the signature or in the parameters table?

My personal take on this question is that I'de rather always have the annotations in the signatures. The informal annotation should be displayed aside the variable names, probably in parenthesis. 

If we look at sphinx, it's also the behaviour that has been adopted. This is an example if a documentation generated by sphinx: https://jira.readthedocs.io/en/master/api.html?highlight=search_issues#jira.client.JIRA.search_issues

Here, having the duplication of informations is not that bad. So I would propose something similar. 

This means that we can simplify the part of the code the merges the annotations from AST and from the type fields.

tristanlatr:
Regarding #402, I will submit a PR that adds only the PyvalColorizer and update _ValueFormatter and _AnnotationFormatter related code, but in any case, #386 needs to be merged first. 

The PR #386 is really a stepping stone for many features and it would be great to have it reviewed and merged :)

tristanlatr:
To summarize what's currently on the table, here is a list of the propositions: 

Propositions that applies to all functions

- Always have the type annotations from AST with the signature (parameters and return type).
- Keep (and improve) the part of the code the merges the annotations from AST and from the informal type fields and keep display that information aside the parameters names when they are documented OR that the type is set from informal type fields, otherwise show nothing (stop listing the parameters as "undocumented" when they only have AST annotation, this makes the behaviour more consistent because we don't list undocumented param that don't have AST annotation). 

Propositions that applies to overloads function

- Have a single documentation entry solution that simply list several signatures, maybe with `@overload` keyword on each. 
- Show the overloads signatures only. Omit the final variant signature (the one without the @overload decoration in the source)
- In the parameters table, only show annotation (parameters and returns) from the informal type fields aside parameters names. Precise types are added in the overload definitions

Thanks for the feedback.

pydoctor/epydoc2stan.py
```diff
@@ -376,31 +376,36 @@ def __init__(self, obj: model.Documentable):
 
         self.parameter_descs: List[FieldDesc] = []
         self.return_desc: Optional[FieldDesc] = None
-        self.raise_descs: List[RaisesDesc] = []
+        self.raise_descs: List[FieldDesc] = []
```
tristanlatr, line 379: This is an error.

cretz:
Can this be revisited? I would be willing to help get this over the finish line with the propositions @tristanlatr has listed. If there is anything I can do to help it along, let me know. If desired, I can make a fresh PR w/ a forked branch w/ the changes herein applied on top of master to fix conflicts.

tristanlatr:
Hello @cretz, 

Thanks a lot for you interest in pydoctor, I'm totally willing to hand this feature over you.

I'll have a look at the diff on this PR and add more comments because there are couple of stuff that I wouldn't have done this way now.

I believe starting from a clean branch from master would be the best if you like to contribute. I would suggest to fist create PR that includes the annotation in signatures and then another PR which takles the overload decorator.

As you can see in the [contributing docs](https://pydoctor.readthedocs.io/en/latest/contrib.html#review-process-and-requirements), anyone can review the PRs other contributors submit. So if you like it, you can add some comments/approve some of my PRs too ;)

Thanks a lot,

Talk to later,

pydoctor/astbuilder.py
```diff
@@ -765,46 +765,69 @@ def _handleFunctionDef(self,
...
             if is_classmethod:
                 attr.report(f'{attr.fullName()} is both property and classmethod')
             if is_staticmethod:
                 attr.report(f'{attr.fullName()} is both property and staticmethod')
             return
-
-        func = self.builder.pushFunction(func_name, lineno)
+        
+        # Check if it's a new func
+        func: model.Function = parent.contents.get(func_name) # type:ignore
```
tristanlatr, line 820: We should not need type ignore here.

pydoctor/astbuilder.py
```diff
@@ -765,46 +765,69 @@ def _handleFunctionDef(self,
...
+        if func is not None and isinstance(func, model.Function) and any(
+            'overload' in (node2dottedname(deco)or()) for
```
tristanlatr, line 822: Meaning this code here.

We should probably use node2fullname() and check whether it uses the « typing.overload ». Instead of just overload.

pydoctor/astbuilder.py
```diff
@@ -919,12 +959,13 @@ def _annotation_from_attrib(self,
 
     def _annotations_from_function(
             self, func: Union[ast.AsyncFunctionDef, ast.FunctionDef]
-            ) -> Mapping[str, Optional[ast.expr]]:
+            ) -> Tuple[Mapping[str, Optional[ast.expr]], Optional[ast.expr]]:
         """Get annotations from a function definition.
         @param func: The function definition's AST.
-        @return: Mapping from argument name to annotation.
-            The name C{return} is used for the return type.
```
tristanlatr, line 926: We can probably keep the older way of storing the return annotation, since parameters can’t be called « return », it’s probably better not to change this needlessly.

this will also make the some code more simple.

tristanlatr:
Replaced by PR #595.

PR #419, opened 2021-07-07
Enable break-word for parameter names and function names

tristanlatr:
This small adjustment to the CSS makes long words (namely long parameter names or function names) take multiple lines (without hyphen) when there is no space left. 
The current behaviour is annoying because long names overlaps with documentation.

Example of annoyance: [configargparse.ArgumentParser](https://tristanlatr.github.io/ConfigArgParse/configargparse.ArgumentParser.html#__init__)

### DIFF
--- a/docs/epytext_demo/demo_epytext_module.py
+++ b/docs/epytext_demo/demo_epytext_module.py
@@ -5,7 +5,7 @@ Most part of this documentation is using Python type hinting.
 """
 
 from abc import ABC
-from typing import AnyStr, Generator, Union
+from typing import AnyStr, Dict, Generator, List, Union
 from somelib import SomeInterface
 import zope.interface
 import zope.schema
@@ -39,7 +39,7 @@ def demo_typing_arguments(name: str, size: bytes) -> bool:
 
 def demo_long_function_and_parameter_names__this_indeed_very_long(
         this_is_a_very_long_parameter_name_don_t_do_that_at_home_kids: str, 
-        what__another_super_super_long_name__ho_no: Generator[Union[list[AnyStr], dict[str, AnyStr]], None, None]) -> bool:
+        what__another_super_super_long_name__ho_no: Generator[Union[List[AnyStr], Dict[str, AnyStr]], None, None]) -> bool:
     """
     Long names and annotations should display on several lines when they don't fit in a single line. 
     """
--- a/docs/epytext_demo/demo_epytext_module.py
+++ b/docs/epytext_demo/demo_epytext_module.py
@@ -38,7 +38,7 @@ def demo_typing_arguments(name: str, size: bytes) -> bool:
     return True
 
 def demo_long_function_and_parameter_names__this_indeed_very_long(
-        this_is_a_very_long_parameter_name_don_t_do_that_at_home_kids: str, 
+        this_is_a_very_long_parameter_name_aahh: str, 
         what__another_super_super_long_name__ho_no: Generator[Union[List[AnyStr], Dict[str, AnyStr]], None, None]) -> bool:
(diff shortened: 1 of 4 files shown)


adiroiban (changes requested):
Thanks. 

As a minor comment, consider updating the example project to include a long parameter name so that we can observe that this is ok.

I am not a CSS expert so I can't tell the side effects of this change.

Do we need to enable it in the "body". Maybe it should be enabled only for `.fieldArgContainer` 

As as major comment, I think that word-break should only be applied to `.fieldArgContainer`

tristanlatr:
Thanks for the quick review @adiroiban 

> As as major comment, I think that word-break should only be applied to `.fieldArgContainer`

There is also the function names that could overlap with the documentation in the summary table. So at least it should be enable for `.fieldArgContainer` and also for `#splitTables > table tr td:nth-child(2)`. This is why I choose to apply it to the whole body regardless, but I'll revert that.

tristanlatr:
> it should be enable for .fieldArgContainer and also for #splitTables > table tr td:nth-child(2)

It's now only enabled for the above selectors. 

Not sure where the mypy error comes from, though. It's probably a recent update in the `types-docutils` package.

adiroiban (approved):
Sorry for the delay.

I have no idea how to best fix this.

I think that this can be merged and if  later we found other ways to fix the long lines we can have a separate PR.

Thanks!

tristanlatr:
We should be good :)

tristanlatr:
@adiroiban should I merge this PR ?

Thanks

adiroiban:
yes. I think that you can merge it.

as a general rule, if a PR is not reviewed in more than 1 week, I think that is safe to merge it.

as otherwise the project is blocked.

tristanlatr:
> as a general rule, if a PR is not reviewed in more than 1 week, I think that is safe to merge it.

Hum,  they are PR opened for months now. I’m particularly thinking of #386 . I don’t want to merge something that did not reach a certain consensus before, but at the same time I’m getting tired of waiting an approval, cause we can always do better. I’m sorry if not all my PRs are as clean as this one. I’m not a professional developer, I just enjoy making pydoctor better and more recognized in the python documentation ecosystem. Anyway, I’m not sure what you exactly meant by « safe to merge it », because if I don’t get an approval, it’s not socially safe (at least) to merge it.

adiroiban:
You can ping me for "emergency review" .

But if we don't merge your request, the other way is for you to fork pydoctor ... and I don't think that will help anybody.

I am in a similar situation with Twisted in which my PR were not reviewed due to lack of interest for those features...but I had to deliver those features to the customers, so I ended up with a fork.

It makes my life harder when syncing with Twisted,  but there are also many other small bugfixed and improvements that are not in Twisted trunk.

adiroiban:
If something break and people start complaining, then those people should dedicate more time doing code review for pydoctor.

So @tristanlatr  don't worry about the code quality or PR quality. I am very grateful that you are contributing! Thanks again!

tristanlatr:
> So @tristanlatr don't worry about the code quality or PR quality.

Not sure our camarade @mthuurne feels the same way. 

> the other way is for you to fork pydoctor

I really don’t want to do that, I’m not doing this for clients. What I’m doing, nevertheless, is that I’m externalizing the ast builder and the python language model (see #430).

tristanlatr:
> You can ping me for "emergency review" .

I’m a bit doubtful regarding the concept of emergency in this context, honestly. Nevertheless, some PRs are blocking the développement of others, namely #386. So if you have time to give your opinion regarding the status of this PR , it would be very appreciated.

Issue #93, opened 2015-06-06
Generate JSON of native docstrings


ericholscher:
It would be awesome to be able to generate a JSON representation of the raw docstrings.

Something along the lines of epyparse: https://github.com/averagehuman/python-epyparse

mwhudson:
I think this make sense, on the face of it. Can you give a couple of more explicit use cases though? I'm lacking in imagination this evening :-)

ericholscher:
The main use case is to pydoctor as more of a library to pull docstrings
and structure out, and render it in your own output format. Having a
representation that is output agnostic is useful for lots of things,
including testing, portability, and downstream tooling.

On Sun, Jun 7, 2015 at 3:01 AM, Michael Hudson-Doyle <
notifications@github.com> wrote:

> I think this make sense, on the face of it. Can you give a couple of more
> explicit use cases though? I'm lacking in imagination this evening :-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/twisted/pydoctor/issues/93#issuecomment-109731743>.
>



-- 
Eric Holscher
Maker of the internet residing in Portland, Oregon
http://ericholscher.com

ericholscher:
The main use case I have for it, is to be able to include parse-only Python
docs in a Sphinx project. I want to output RST directly, not use pydoctor's
HTML output, and then have an intersphinx inventory.

On Sun, Jun 7, 2015 at 8:22 AM, Eric Holscher <eric@ericholscher.com> wrote:

> The main use case is to pydoctor as more of a library to pull docstrings
> and structure out, and render it in your own output format. Having a
> representation that is output agnostic is useful for lots of things,
> including testing, portability, and downstream tooling.
>
> On Sun, Jun 7, 2015 at 3:01 AM, Michael Hudson-Doyle <
> notifications@github.com> wrote:
>
>> I think this make sense, on the face of it. Can you give a couple of more
>> explicit use cases though? I'm lacking in imagination this evening :-)
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/twisted/pydoctor/issues/93#issuecomment-109731743>.
>>
>
>
>
> --
> Eric Holscher
> Maker of the internet residing in Portland, Oregon
> http://ericholscher.com
>



-- 
Eric Holscher
Maker of the internet residing in Portland, Oregon
http://ericholscher.com

mwhudson:
[...]

PR #424, opened 2021-08-11
Cleanup astbuilder and add support for nested modules

tristanlatr:
Create the `SignatureBuilder` and the `BaseBuilder` classes.

This PR does some cleanup in order to generalize part of the ASTBuilder logic such that we can subclass it to build from and to JSON data. See #93.


tristanlatr:
This is probably not needed if we are using [pydocspec](https://tristanlatr.github.io/pydocspec/pydocspec.html) in the future.

tristanlatr:
I’m reopening this PR since I believe not everything is to be trashed here, especially the signature builder class.

PR #425, opened 2021-08-14
Documentable.setDocstring now accepts a string type.

tristanlatr:
When passing a `str`: the `Documentable.docstring_lineno` attribute is set to the same values as `Documentable. linenumber` , instead of calculating the real line number. 
This is needed in order to load the documentable data from JSON, where the raw docstring is stored as astring. 

See #93.

### DIFF
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -146,23 +146,27 @@ class Documentable:
         #       requires a boatload of changes.
         self.contents: Dict[str, Any] = {}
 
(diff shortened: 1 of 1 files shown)


tristanlatr:
Having a optional `docstring_lineno ` parameter would be better I suppose. Maybe that's a case where we should use `@overload`.

tristanlatr:
I'll close this PR because

PR #426, opened 2021-08-20
Docutils typeshed are a bit more complete now.

tristanlatr:
We don't need this type ignore comment anymore. 

Issue https://github.com/python/typeshed/issues/5667 is fixed.


adiroiban (approved):
Thanks. All good!

Issue #420, opened 2021-07-09
Getting organize for next release 


tristanlatr:
Hi, 

With the new templating system in place, it's probably time to do a release. 

But before, I think the following PRs should be merged:

- #419 because it's a really easy fix for a real problem
- #389 because it introduces improvements to the templating system, this way we can already use the [pydoctor-theme-readthedocs](https://github.com/tristanlatr/pydoctor-theme-readthedocs) as a custom template. 

Other PR can probably wait until another release. 

Tell what you think,

Thanks

tristanlatr:
I think we should make a release now. I’ll create a branch 21.9.x and tag it.

Do I need any approval for such process ?

Thanks.

adiroiban:
looks like we have a few release docs here https://github.com/twisted/pydoctor/blob/master/docs/source/contrib.rst#releasing-and-publishing-a-new-package

It contains "First commit the version update to master and wait for tests to pass."

I would prefer to have a separate branch on which the version is updated on on which we check the tests are green.

Once the tests are green, create a tag/release based on the code from the branch.

if all is ok, the version of the branch can be updated to `.dev0` and the branch can be merged in master.

But yes @tristanlatr  feel free to create the release branch. I can review it.

Thanks!

adiroiban:
the release is triggered from here https://github.com/twisted/pydoctor/blob/master/.github/workflows/unit.yaml#L83

any time a tag is created

tristanlatr:
Hi @adiroiban ,

So if I understand correctly, first I create a release branch, then I update the version number to « 21.9.0 », then I merge this branch to master ( with your approval ), then I tag this version with the tag « 21.9.0 », this will trigger the release process on pypi.

I do not understand why you want a dev0 version, though ?

adiroiban:
yes. That is what is documented right now.

But I will suggest using this process:

* create a branch: name it release-21.9.0
* On the branch update the version and release notes and release date
* Create a PR for that branch and get an approval
* Create a tag based on the HEAD of the release branch
* This will trigger the release
* Update the code on the branch and have version 21.9.0.post0 (or dev0)
* Merge the branch

In this way, version 21.9.0 only existed for a brief period of time...and the main branch will have 21.9.0.

Right now, you will check the code and it says  version 21.2.2, but the code from master is not 21.2.2

https://github.com/twisted/pydoctor/blob/master/setup.cfg#L3

adiroiban:
[...]

PR #432, opened 2021-09-26
Update release docs.

tristanlatr:
Following the process detailed in https://github.com/twisted/pydoctor/issues/420#issuecomment-926529563

### DIFF
--- a/docs/source/contrib.rst
+++ b/docs/source/contrib.rst
@@ -45,6 +45,7 @@ The following process ensures correct version management:
         git push --tags
 
  - Update the version on the branch and append ``.dev0`` to the current version number. In this way, stable versions only exist for a brief period of time (if someone tries to do a ``pip install`` from the git source, they will get a ``.dev0`` version instead of a misleading stable version number.
+ - Update the README file and add an empty placeholder for unreleased changes.
  - Merge the branch
 
 Author Design Notes
--- a/docs/source/contrib.rst
+++ b/docs/source/contrib.rst
@@ -32,17 +32,20 @@ environment dedicated to that extra test.
 Releasing and publishing a new package
 --------------------------------------
 
-Publishing to PyPI is done via a GitHub Actions workflow that is triggered when a tag is pushed.
+Publishing to PyPI is done via a GitHub Actions workflow that is triggered when a tag is pushed. Version is configured in the ``setup.cfg``. 
 
-First commit the version update to master and wait for tests to pass.
+The following process ensures correct version management: 
 
-You can push a tag by creating a `GitHub release <https://github.com/twisted/pydoctor/releases/new>`_.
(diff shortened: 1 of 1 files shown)


docs/source/contrib.rst
```diff
@@ -32,17 +32,20 @@ environment dedicated to that extra test.
...
 --------------------------------------
 
-Publishing to PyPI is done via a GitHub Actions workflow that is triggered when a tag is pushed.
+Publishing to PyPI is done via a GitHub Actions workflow that is triggered when a tag is pushed. Version is configured in the ``setup.cfg``. 
 
-First commit the version update to master and wait for tests to pass.
+The following process ensures correct version management: 
 
-You can push a tag by creating a `GitHub release <https://github.com/twisted/pydoctor/releases/new>`_.
+ - Create a branch: name it by the name of the new major ``pydoctor`` version, i.e. ``21.9.x``, re-use that same branch for bug-fixes.
```
adiroiban, line 39:
```suggestion
 - Create a branch: name it by the name of the new major ``pydoctor`` version, i.e. ``release-21.9.x``, re-use that same branch for bug-fixes.
```

If you want to reuse the branch and not delete it, it's best to not have it using the same convention as tag... to help with autocompletion.

I would say that is best to just name the branch `release-2.9.0` and delete the branch after the merge.

----------

when a bugfix is required you can create branch `release-21.9.1` based on the existing tag.

docs/source/contrib.rst
```diff
@@ -32,17 +32,20 @@ environment dedicated to that extra test.
...
+ - Create a tag based on the ``HEAD`` of the release branch, name it by the full version number of the ``pydoctor`` version, i.e. ``21.9.1``, this will trigger the release. For instance::
 
-You can also create a tag on local branch then push it::
-
-        git tag 1.2.3
+        git tag 21.9.1
         git push --tags
 
+ - Update the version on the branch and append ``.dev0`` to the current version number. In this way, stable versions only exist for a brief period of time (if someone tries to do a ``pip install`` from the git source, they will get a ``.dev0`` version instead of a misleading stable version number.
+ - Merge the branch
```
adiroiban, line 48:
```suggestion
 - Update the version on the branch and append ``.dev0`` to the current version number. In this way, stable versions only exist for a brief period of time (if someone tries to do a ``pip install`` from the git source, they will get a ``.dev0`` version instead of a misleading stable version number.
 - Update the README file and add an empty placeholder for unreleased changes.
 - Merge the branch
```

adiroiban (approved):
Thanks for the release and this PR.

All good. Only minor comments.

Thanks

docs/source/contrib.rst
tristanlatr: I would prefer continue using the same branch name convention as the precedent versions. Actually I think it’s best to use the .x convention such that one can install a specific major version (with the latest fixes) with pip by selecting the branch name.

adiroiban: ok. np.

PR #433, opened 2021-09-27
Avoid using shell script inside pydoctor argument in example.

tristanlatr:
This is not required and might confuse users.

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -27,7 +27,7 @@ The result looks like `this <api/pydoctor.html>`_.
         --html-viewsource-base=https://github.com/twisted/pydoctor/tree/20.7.2 \
         --make-html \
         --html-output=docs/api \
-        --project-base-dir="$(pwd)" \
+        --project-base-dir="." \
         --docformat=epytext \
         --intersphinx=https://docs.python.org/3/objects.inv \
         ./pydoctor


tristanlatr:
This is a no brainer review, if you want to give it go @graingert, it would be very appreciated. 

Thanks

docs/source/quickstart.rst
```diff
@@ -27,7 +27,7 @@ The result looks like `this <api/pydoctor.html>`_.
         --html-viewsource-base=https://github.com/twisted/pydoctor/tree/20.7.2 \
         --make-html \
         --html-output=docs/api \
-        --project-base-dir="$(pwd)" \
+        --project-base-dir="." \
```
adiroiban, line 30: Does it work with relative paths ?

```suggestion
        --project-base-dir="/home/johnd/projectx/" \
```

adiroiban (approved):
Thanks.

docs/source/quickstart.rst
tristanlatr: Yes it does work with relative paths.

adiroiban: perfect. Feel free to merge it :)

In the future, I guess that `--project-base-dir` can default to current working directory so that we don't have to explicitly use it.

tristanlatr: See #441

PR #439, opened 2021-10-03
configure pytest strict mode

### LINKED ISSUE #438: Run tests with python `PYTHONWARNINGS=error`
No warnings should be triggered while running the tests. This would avoid issues like #435.

There is still currently one `UserWarning` being triggered: 

```
driver.py:473: UserWarning: Writer 'InMemoryWriter' does not support HTML template customization with --template-dir.
```

We should first change the tests such that this warning get triggered inside a `pytest.warns` clause such that it does not get listed in the warnings summary and we assert it triggers correctly. 

Then, we can enable errors on warnings, see [PYTHONWARNINGS docs](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNINGS).

graingert:
Fixes #438

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,8 +70,10 @@ doc_files = README.txt
 [tool:pytest]
 testpaths = pydoctor/test
 python_files = test_*.py
-addopts = --doctest-glob='*.doctest'
+addopts = --doctest-glob='*.doctest' --strict-config --strict-markers
 doctest_optionflags = ELLIPSIS IGNORE_EXCEPTION_DETAIL
+xfail_strict = true
+filterwarnings = error
 
 [upload]
 sign = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -76,7 +76,6 @@ xfail_strict = true
 filterwarnings =
     error
     ignore:Writer 'InMemoryWriter' does not support HTML template customization with --template-dir\.:UserWarning:pydoctor.driver
-    ignore:Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated since Python 3\.3, and in 3\.(9|10) it will stop working:DeprecationWarning: pydoctor.templatewriter.util
 
 [upload]
 sign = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -76,7 +76,7 @@ xfail_strict = true
 filterwarnings =
     error
     ignore:Writer 'InMemoryWriter' does not support HTML template customization with --template-dir\.:UserWarning:pydoctor.driver
-    ignore:Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated since Python 3\.3, and in 3\.10 it will stop working:DeprecationWarning: pydoctor.templatewriter.util
+    ignore:Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated since Python 3\.3, and in 3\.(9|10) it will stop working:DeprecationWarning: pydoctor.templatewriter.util
 
 [upload]
 sign = True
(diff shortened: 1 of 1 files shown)


graingert:
@tristanlatr wow only the collections.abc warning and the UserWarning you mentioned - that's impressive for not checking the warnings!

setup.cfg
```diff
@@ -70,8 +70,10 @@ doc_files = README.txt
 [tool:pytest]
 testpaths = pydoctor/test
 python_files = test_*.py
-addopts = --doctest-glob='*.doctest'
+addopts = --doctest-glob='*.doctest' --strict-config --strict-markers
 doctest_optionflags = ELLIPSIS IGNORE_EXCEPTION_DETAIL
+xfail_strict = true
+filterwarnings = error
```
graingert, line 76:
```suggestion
filterwarnings =
    error
    ignore:Writer 'InMemoryWriter' does not support HTML template customization with --template-dir\.:UserWarning:pydoctor.driver
    ignore:Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated since Python 3\.3, and in 3\.10 it will stop working:DeprecationWarning: pydoctor.templatewriter.util
```

setup.cfg
```diff
@@ -70,8 +70,13 @@ doc_files = README.txt
...
 testpaths = pydoctor/test
 python_files = test_*.py
-addopts = --doctest-glob='*.doctest'
+addopts = --doctest-glob='*.doctest' --strict-config --strict-markers
 doctest_optionflags = ELLIPSIS IGNORE_EXCEPTION_DETAIL
+xfail_strict = true
+filterwarnings =
+    error
+    ignore:Writer 'InMemoryWriter' does not support HTML template customization with --template-dir\.:UserWarning:pydoctor.driver
+    ignore:Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated since Python 3\.3, and in 3\.10 it will stop working:DeprecationWarning: pydoctor.templatewriter.util
```
graingert, line 79:
```suggestion
    ignore:Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated since Python 3\.3, and in 3\.(9|10) it will stop working:DeprecationWarning: pydoctor.templatewriter.util
```

tristanlatr: This warning is now fixed in master.

setup.cfg
```diff
@@ -70,8 +70,13 @@ doc_files = README.txt
...
 testpaths = pydoctor/test
 python_files = test_*.py
-addopts = --doctest-glob='*.doctest'
+addopts = --doctest-glob='*.doctest' --strict-config --strict-markers
 doctest_optionflags = ELLIPSIS IGNORE_EXCEPTION_DETAIL
+xfail_strict = true
+filterwarnings =
+    error
+    ignore:Writer 'InMemoryWriter' does not support HTML template customization with --template-dir\.:UserWarning:pydoctor.driver
+    ignore:Using or importing the ABCs from 'collections' instead of from 'collections\.abc' is deprecated since Python 3\.3, and in 3\.(9|10) it will stop working:DeprecationWarning: pydoctor.templatewriter.util
```
graingert, line 79:
```suggestion
```

tristanlatr:
> that's impressive for not checking the warnings!

Thanks for your kind words, @graingert. 

The PR looks very good. 

Regarding the `UserWarning`, I thing we should simply change the `InMemoryWriter`. The interface ([`IWriter`](https://pydoctor.readthedocs.io/en/latest/api/pydoctor.templatewriter.IWriter.html)) stipulate that the init method should take the argument `template_lookup`, this warning is designed for those who have a custom (older) version of the writer and uses it with the option `--html-writer`.

tristanlatr (approved):

PR #440, opened 2021-10-03
Fix `InMemoryWriter` to match `IWriter` interface and remove unnecessary warning

tristanlatr:
Avoids UserWarning to be triggered while testing. 

We should add a test for this code block nevertheless, or completely remove the catch block if we decide that's not worth it to keep. 

This warning come from the fact that we can pass a custom HTML writer with the option `--html-writer`, the thing is that the interface of such writer has changed, we don't pass the build directory as `str` anymore, bu rather as `Path`, so in any case, older customized HTML writer will be broken, If the customization overrides the `__init__` method. 

So I think it's probably OK to remove the `TypeError` catching altogether and remove the `@overload` from the `IWriter.__init__` definition.

### DIFF
--- a/pydoctor/driver.py
+++ b/pydoctor/driver.py
@@ -463,15 +463,9 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
 
             build_directory = Path(options.htmloutput)
 
-            try:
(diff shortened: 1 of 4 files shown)


adiroiban (approved):
awesome. Thanks!

PR #454, opened 2021-11-24
Fix broken link

tristanlatr:


docs/source/docformat/epytext.rst
```diff
@@ -17,7 +17,7 @@ All markup should work except the indexed terms ``X{}`` tag, which has been remo
 Fields
 ------
 
-See `fields section <codedoc.rst#fields>`__.
+See `fields section <../codedoc.html#fields>`__.
```
adiroiban, line 20: Looks good. Thanks.

But maybe is best to add a dedicated Sphinx directive to reference this via https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#ref-role

```suggestion
See :ref:`fields section <codedoc-fields>`.
```
and add this inside codedoc.rst before the fields header.

```
.. _codedoc-fields:
```

In this way, even if we move the section, the links are automatically updated.

adiroiban (approved):
all good. Thanks.
Only a minor comment about using sphinx internal links

docs/source/docformat/epytext.rst
tristanlatr: Thanks!

PR #457, opened 2022-01-02
Read the docs theme

tristanlatr:
This PR adds the `readthedocs` theme. 

Demo: https://pydoctor--457.org.readthedocs.build/en/457/docformat/epytext/epytext_demo.html


docs/source/conf.py
```diff
@@ -108,6 +108,15 @@
         '--project-url=../index.html',
         f'{_pydoctor_root}/pydoctor',
         ] + _common_args,
+    'rtd_theme_api': [
```
adiroiban, line 111: very minor comment.

can we just use the epytext example to showcase the rtd theme?

so that we don't need yet another example.

The idea is that we have the real API docs for pydoctor itself using the standard theme and we have epydoc_demo with RTD theme.

pydoctor/themes/readthedocs/fonts/book.svg
```diff
@@ -0,0 +1,2 @@
+<?xml version="1.0" encoding="utf-8"?>
```
adiroiban, line 1: it this really a font.. or just a svg ?

Maybe use https://fontawesome.com/ or something similar for icons

adiroiban (commented):
just a minor comment. this looks great.

Looking forward to see this code released.

pydoctor/themes/readthedocs/fonts/book.svg
tristanlatr: It's not a font, And i wanted only this SVG, not all the other icons from https://fontawesome.com/. So this is the way I choose to do it.

adiroiban: sure. no.

maybe have the file located into `pydoctor/themes/readthedocs/images/` or something similar :)

pydoctor/themes/readthedocs/readthedocstheme.css
```diff
@@ -0,0 +1,730 @@
...
+#splitTables > table {
+    -webkit-border-horizontal-spacing: 0px;
+    -webkit-border-vertical-spacing: 0px;
+    border: 1px solid rgb(210, 220, 216);
+    border-bottom: 0;
+    vertical-align: middle;
+}
+
+#splitTables > table td {
+    border-bottom: 1px solid rgb(210, 220, 216);
```
not-my-profile, line 241: I don't really like that border in combination with the `border-spacing` of the table. It looks like each cell has a small shadow.

image

glyph:
@tristanlatr Just a bump here in the hopes of getting the conflicts resolved :)

pydoctor/themes/readthedocs/readthedocstheme.css
```diff
@@ -0,0 +1,739 @@
...
+    .sidebarcontainer{
+        display: none;
+    }
+    
+}
+
+/* Search results */
+#search-results-container {
+    position: relative;
+}
```
tristanlatr, line 691:
```suggestion
#search-results-container {
    position: fixed;
    top: 10px;
    left: 300px;
}
```

docs/source/conf.py
tristanlatr: I've done what you suggested.

pydoctor/themes/readthedocs/fonts/book.svg
tristanlatr: We use fonts for the other theme as well, and I think it's a pretty common name for such a folder.

pydoctor/themes/readthedocs/readthedocstheme.css
tristanlatr: Fixed, Thanks

tristanlatr:
Hi @glyph,

We're ready to look at this PR, look forward to see the Twisted docs like that :)

Please let me know if this can be merged by approving this PR.

Thanks,

glyph (commented):
The one thing I'm unclear on here is licensing. It seems like this is pretty substantially a derivative work of https://github.com/readthedocs/sphinx_rtd_theme but I don't see any attribution or inclusion of Dave Snider, Read the Docs, Inc., or for that matter a copy of the Apache license or a mention of Christian Robertson https://fonts.google.com/specimen/Roboto+Slab#about (which the rtd theme seems to have left out as well).

We don't have a lawyer "on staff" as it were but I'd like to see at least a quick check-in from someone associated with RTD first.  (Perhaps @vanl is feeling charitable today but beyond a mention I wouldn't want to bother him.)

Assuming we can get that squared away, I'll approve this.

adiroiban:
Thanks Glyph.

Just for my general knowledge.

Should we worry about licensing ?

Is the world in which we are leaving that crazy to find someone to sue a project like pydoctor ?!

Would a judge in any country consider such a case?

--------

Now that you mention, yes. we can update the LICENCE to mention the RTD MIT licence.
and the font licence.

I don't think there is any malicious intent in this PR.

tristanlatr:
I'll add mentions to the LICENSE file pronto!

Let's note that we don't have a license mention for bootstrap CSS neither, which is used by pydoctor since 2015 ^^

glyph:
> Should we worry about licensing ?

Yes, always. Software engineering is 90% being worried about licensing 5% getting into fights on mailing lists, 1% coding, and 1% rounding errors

> Is the world in which we are leaving that crazy to find someone to sue a project like pydoctor ?!

"Pydoctor" is barely even an entity you can sue. The issue is that we want people to be able to *use* Pydoctor (and Twisted, and…) and if they work for even a medium-sized tech company, people are always looking for ways to sue those. So screwing up copyrights, attributions, licenses etc is one way that you might make it possible for somebody to sue your users.

Consider a situation where we have engineer X who works for BigCo Y, X starts using Pydoctor at work.  Y's major rival Z figures this out by scraping github repos, Z files another lawsuit against Y, X gets fired, everybody has a real bad time.

Is this one, specific copyright issue a problem right now? No. Do I expect that any of these copyright holders will be either litigious or bought out by some bad actor? Also no. But it's like security; the place where the attackers get you is the place you forgot to defend, so you've gotta be diligent everywhere just in case.

> Would a judge in any country consider such a case?

Of course.

> I don't think there is any malicious intent in this PR.

Malicious intent isn't really relevant, at least in US law, where I think the only place that matters is calculation of per-infringement damages.

tristanlatr:
Let's be diligent, I'll add the full Apache License as well.

glyph:
> Let's be diligent, I'll add the full Apache License as well.

If it's just on the fonts, let's make it clear that it only applies to those, I don't want to add the complexity of several different licenses applying to the software as a whole.  Also, the fonts are *not* distributable under the MIT license, so there would be some confusion there.

glyph:
> Let's note that we don't have a license mention for bootstrap CSS neither, which is used by pydoctor since 2015 ^^

Oops, thanks for spotting that.

tristanlatr:
Maybe @stsewd could chip in and reassure us about the legitimacy of this theme.

Thanks,

stsewd:
> Maybe @stsewd could chip in and reassure us about the legitimacy of this theme.

do you mean about the licensing? I can't be much help there, but I can ask the team about it.

tristanlatr:
Yes I meant about licensing @stsewd, thanks a lot for checking.

tristanlatr:
TODO: fix code blocks style for smaller displays.

BBAC1FC5-945D-4304-899D-C46BBD08B77A

stsewd:
> or for that matter a copy of the Apache license or a mention of Christian Robertson https://fonts.google.com/specimen/Roboto+Slab#about (which the rtd theme seems to have left out as well).

Looks like we do https://github.com/readthedocs/sphinx_rtd_theme/blob/master/Apache-License-2.0.txt

tristanlatr:
Hi @ericholscher @stsewd, thanks a lot for checking. I understand that we should be good license wise.

All credits can be found in the license file (I just liked the Apache license) https://github.com/twisted/pydoctor/blob/c3620f9c7e09e70af422a1e24e06911c29e52dba/LICENSE.txt#L77

I’ll fix the minor issues still present and merge this PR, With approval of @glyph.

Thanks a lot,

glyph (approved):
✅

pydoctor/themes/readthedocs/readthedocstheme.css
```diff
@@ -0,0 +1,769 @@
...
+    display: block;
+    margin: 0 0 10px;
+    word-break: break-word;
+    word-wrap: break-word;
+}
+pre code {
+    padding: 0;
+    font-size: inherit;
+    color: inherit;
+    white-space: pre-wrap;
```
tristanlatr, line 351:
```suggestion
    white-space: pre-wrap;
}
pre code {
    padding: 0;
    font-size: inherit;
    color: inherit;
```

tristanlatr (commented):
Couple of minor stuff, after I'll merge it and release a new version of pydoctor :)

PR #458, opened 2022-01-10
Include the sre_parse36 module in pyodoctor.epydoc

tristanlatr:
It seems that extra PyPi dependencies that do not already exist in the Debian archive is generating some trouble (because they need to be added just for pydoctor) and that's probably making pydoctor more slow to update for users that uses `apt` to install pydoctor. 

Issue initially reported here: https://github.com/tristanlatr/sre_parse36/issues/1

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,6 +3,7 @@ branch = True
 omit =
     pydoctor/sphinx_ext/*
     pydoctor/test/*
+    pydoctor/epydoc/sre_parse36.py
 source =
     pydoctor
 
--- a/README.rst
+++ b/README.rst
@@ -73,6 +73,10 @@ What's New?
 in development
 ^^^^^^^^^^^^^^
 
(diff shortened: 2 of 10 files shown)


adiroiban (approved):
All good. Thanks.

One question. I this PR also for the 21.12.1 release.
 
If so, I think that "in development" section from README can be removed, tag and release and than add it back.

Thanks!

tristanlatr:
Ok thanks i'll remove the place holder. 
Also I should find a way to exlclude `pydoctor/epydoc/sre_parse36.py` from the coverage stats.

adiroiban:
Files can be excluded from coverage reporting via `.coveragerc` file . See an example here https://github.com/twisted/pydoctor/blob/master/.coveragerc#L5

tristanlatr:
Thanks @adiroiban , we’re good to merge that branch now

PR #460, opened 2022-01-18
Update codedoc.rst

iosonofabio:
`__all__` should be unmutable, so a tuple is better than a list here. Stumbled upon this while improving docs for `python-igraph`...


tristanlatr:
Thanks @iosonofabio !

tristanlatr (approved):

PR #467, opened 2022-02-02
Insert zero-width spaces for better line wrapping

### LINKED ISSUE #466: Ugly line wrapping of API member names
```python
class Foo:
    def some_methods_have_longer_names(self):
        pass
```

results in the following:

image

because of the following CSS:

https://github.com/twisted/pydoctor/blob/996e03a6042c386e2aa00c248e746f8cc506b81c/pydoctor/themes/base/apidocs.css#L247-L249

I think `word-break: break-word;` should be removed ... breaking words in the middle looks horrible.

not-my-profile:
For the following code:

    class Foo:
        def some_methods_have_longer_names_that_break(self):
            pass

        def someMethodsHaveLongerNamesThatBreak(self):
            pass

Pydoctor previously generated a class page with a table
that Firefox ended up rendering as:

    Method   some_methods_have_longe    Undocumented
             r_names_that_break

    Method   someMethodsHaveLongerNa    Undocumented
             mesThatBreak

This commit makes Pydoctor detect snake_case and camelCase and insert
zero-width spaces accordingly so that the table becomes:

    Method   some_methods_have          Undocumented
             _longer_names_that_break

    Method   someMethodsHaveLonger      Undocumented
             NamesThatBreak

Fixes #466.

### DIFF
--- a/pydoctor/templatewriter/pages/table.py
+++ b/pydoctor/templatewriter/pages/table.py
@@ -47,7 +47,7 @@ class TableRow(Element):
     @renderer
     def name(self, request: object, tag: Tag) -> Tag:
         return tag.clear()(tags.code(
-            epydoc2stan.taglink(self.child, self.ob.url, self.child.name)
+            epydoc2stan.taglink(self.child, self.ob.url, epydoc2stan.insert_break_points(self.child.name))
             ))
 
     @renderer
--- a/README.rst
+++ b/README.rst
@@ -72,6 +72,7 @@ What's New?
 
 in development
 ^^^^^^^^^^^^^^
+* Smarter line wrapping in summary and parameters tables.
 * Fix introspection of functions comming from C-extensions.
 
 pydoctor 21.12.1
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -9,6 +9,7 @@ from typing import (
 )
 import ast
 import itertools
+import re
 
 import attr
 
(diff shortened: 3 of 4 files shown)


tristanlatr:
Thanks a lot ! It looks great.

Would it be possible to quickly add unit tests to the function _insert_break_points()?

Probably this function should also be used to present the parameter names in the parameter table?

not-my-profile:
You're welcome. I updated the PR to add unit tests and also insert break points for parameter names.

tristanlatr:
Thanks a lot, looks perfect.

Could you try to put the new function inside the  `epydoc2stan` module? I don’t think it will create circular imports.

Thanks again

tristanlatr (approved):

pydoctor/templatewriter/pages/table.py
```diff
@@ -6,6 +6,7 @@
 from pydoctor import epydoc2stan
 from pydoctor.model import Documentable, Function
 from pydoctor.templatewriter import TemplateElement, util
+from pydoctor.epydoc2stan import insert_break_points
```
tristanlatr, line 9:
```suggestion
```

pydoctor/templatewriter/pages/table.py
```diff
@@ -47,7 +48,7 @@ def kind(self, request: object, tag: Tag) -> Tag:
     @renderer
     def name(self, request: object, tag: Tag) -> Tag:
         return tag.clear()(tags.code(
-            epydoc2stan.taglink(self.child, self.ob.url, self.child.name)
+            epydoc2stan.taglink(self.child, self.ob.url, insert_break_points(self.child.name))
```
tristanlatr, line 51:
```suggestion
            epydoc2stan.taglink(self.child, self.ob.url, epydoc2stan.insert_break_points(self.child.name))
```

tristanlatr (commented):
Suggesting minor changes because epydoc2stan module is already imported in table.py

tristanlatr:
Thanks!

PR #468, opened 2022-02-03
Make intersphinx links blue

### LINKED ISSUE #464: Having different link color for internal and external object references
Links are good, but we are reaching a point where we can see a lot of red in some pages.

Maybe all links to `@property` don't need to be in red... 

I suggest that we use the red color for links that points to an internal object, otherwise use a more regular color like black or blue.

tristanlatr:
Make intersphinx links blue. 

We now have a clear indication on whether the linked object is in the project or not.

Example parameters values and annotations: https://pydoctor--468.org.readthedocs.build/en/468/api/pydoctor.epydoc.markup.html#get_parser_by_name

Example base classes: https://pydoctor--468.org.readthedocs.build/en/468/api/pydoctor.test.test_sphinx.ClosingBytesIO.html

Fixes #464

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -72,6 +72,7 @@ What's New?
 
 in development
 ^^^^^^^^^^^^^^
+* Make intersphinx links blue instead of red. Keep red color for internal links only.
 * Smarter line wrapping in summary and parameters tables.
 * Fix introspection of functions comming from C-extensions.
 
--- a/pydoctor/test/test_epydoc2stan.py
+++ b/pydoctor/test/test_epydoc2stan.py
@@ -390,7 +390,7 @@ def test_func_raise_linked() -> None:
         """
     ''', modname='test')
     html = docstring2html(mod.contents['f']).split('\n')
-    assert '<a href="test.SpanishInquisition.html">SpanishInquisition</a>' in html
+    assert '<a href="test.SpanishInquisition.html" class="internal-link">SpanishInquisition</a>' in html
 
 
 def test_func_raise_missing_exception_type(capsys: CapSys) -> None:
@@ -1175,7 +1175,7 @@ def test_constant_values_rst(capsys: CapSys) -> None:
     expected = ('<table class="valueTable"><tr class="fieldStart">'
                 '<td class="fieldName">Value</td></tr><tr><td>'
                 '<pre class="constant-value"><code>(<wbr></wbr>'
-                '<a href="pack.mod1.html#f">f</a>)</code></pre></td></tr></table>')
+                '<a href="pack.mod1.html#f" class="internal-link">f</a>)</code></pre></td></tr></table>')
     
     attr = mod.contents['CONST']
     assert isinstance(attr, model.Attribute)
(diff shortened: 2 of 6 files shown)


README.rst
```diff
@@ -72,6 +72,7 @@ What's New?
 
 in development
 ^^^^^^^^^^^^^^
+* Make intersphinx links blue instead of red. Keep red color for internal links only.
```
adiroiban, line 75: I think that the important change is that you can now, as a theme designer, you can differentiate between internal and external documentation links. 

```suggestion
* Intersphinx links have now dedicated markup.
  With the default theme,
  this allows to have the external intershinx links blue while the internal links are red.
```

pydoctor/themes/base/apidocs.css
```diff
@@ -371,6 +371,12 @@ code > a, #childList > div .functionHeader a {
     color:#c7254e;
     background-color:#f9f2f4;
 }
+
+/* Intersphinx links are not red, but simply blue */
+a.intersphinx-link {
+    color:#306998!important;
```
adiroiban, line 377: very minor suggestion

at least on my computer, with the grey background, this blue has little contrast and saturation. and looks like dark grey

```suggestion
    color:##06f!important;
```

README.rst
adiroiban, line 75: I think that the important change is that you can now, as a theme designer, you can differentiate between internal and external documentation links. 

```suggestion
* Intersphinx links have now dedicated markup.
  With the default theme,
  this allows to have the external intershinx links blue while the internal links are red.
```

adiroiban (approved):
Looks good. Thanks.

Only minor comments.

I think that it would help to also have markup for the internal apidoc links.

In this way you can have 3 types of links

* internal apidoc links
* external intersphinx links
* generic external links to non-documentation pages

tristanlatr:
> I think that it would help to also have markup for the internal apidoc links. In this way you can have 3 types of links

We already have three types of css for links, but it’s implicit. The rules applied are the following: links are blue normal by default, links inside code tags are red, except if it’s an intersphinx link. I agree having another special css class for internal links would be good. I’ll try to add that to this PR.

tristanlatr:
Please tell me if the contrast is Ok. 
I did not went with your color choice, i found it just too much intense blue.

adiroiban (approved):
Thanks. All good.

the color and contrast is a matter of taste and monitor capabilities :)

This can be merged. Thanks again for the update!

pydoctor/test/test_epydoc2stan.py
```diff
@@ -390,7 +390,7 @@ def f():
         """
     ''', modname='test')
     html = docstring2html(mod.contents['f']).split('\n')
-    assert '<a href="test.SpanishInquisition.html" title="test.SpanishInquisition">SpanishInquisition</a>' in html
+    assert '<a href="test.SpanishInquisition.html" title="test.SpanishInquisition" class="internal-link">SpanishInquisition</a>' in html
```
tristanlatr, line 393:
```suggestion
    assert '<a href="test.SpanishInquisition.html" class="internal-link" title="test.SpanishInquisition">SpanishInquisition</a>' in html
```

PR #471, opened 2022-02-03
Skip `if __name__ == '__main__'` blocks

### LINKED ISSUE #470: Ignore everything declared in `if __name__ == '__main__'` blocks
For the following module:

```python
if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('x', type=int)
    args = parser.parse_args()
```

image

which is just wrong. These variables are not part of the API ... they cannot be imported (see the [Python documentation](https://docs.python.org/3/library/__main__.html#name-main)). Everything declared in `if __name__ == '__main__':` blocks should be excluded from the documentation.

not-my-profile:
Fixes #470.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,6 +73,7 @@ What's New?
 in development
 ^^^^^^^^^^^^^^
 * Smarter line wrapping in summary and parameters tables.
+* Any code inside of ``if __name__ == '__main__'`` is now excluded from the documentation.
 * Fix introspection of functions comming from C-extensions.
 
 pydoctor 21.12.1
--- a/pydoctor/astutils.py
+++ b/pydoctor/astutils.py
@@ -2,6 +2,7 @@
 Various bits of reusable code related to L{ast.AST} node processing.
 """
 
+import sys
 from typing import Optional, List
 from inspect import BoundArguments, Signature
 import ast
@@ -34,3 +35,24 @@ def bind_args(sig: Signature, call: ast.Call) -> BoundArguments:
         if kw.arg is not None
         }
     return sig.bind(*call.args, **kwargs)
+
+
+if sys.version_info[:2] >= (3, 8):
+    # Since Python 3.8 "foo" is parsed as ast.Constant.
+    def _is_str_constant(expr: ast.expr, s: str) -> bool:
+        return isinstance(expr, ast.Constant) and expr.value == s
+else:
+    # Before Python 3.8 "foo" was parsed as ast.Str.
+    def _is_str_constant(expr: ast.expr, s: str) -> bool:
+        return isinstance(expr, ast.Str) and expr.s == s
+
+def is__name__equals__main__(cmp: ast.Compare) -> bool:
+    """
+    Returns whether or not the given L{ast.Compare} is equal to C{__name__ == '__main__'}.
+    """
+    return isinstance(cmp.left, ast.Name) \
+    and cmp.left.id == '__name__' \
+    and len(cmp.ops) == 1 \
+    and isinstance(cmp.ops[0], ast.Eq) \
+    and len(cmp.comparators) == 1 \
+    and _is_str_constant(cmp.comparators[0], '__main__')
--- a/pydoctor/test/test_astbuilder.py
+++ b/pydoctor/test/test_astbuilder.py
@@ -1988,3 +1988,25 @@ def test_constant_override_do_not_warns_when_defined_in_module_docstring(systemc
     assert ast.literal_eval(attr.value) == 99
     captured = capsys.readouterr().out
     assert not captured
+
+@systemcls_param
+def test__name__equals__main__is_skipped(systemcls: Type[model.System]) -> None:
+    """
+    Code inside of C{if __name__ == '__main__'} should be skipped.
+    """
+    mod = fromText('''
+    foo = True
+
+    if __name__ == '__main__':
+        var = True
+
+        def fun():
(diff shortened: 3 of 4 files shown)


pydoctor/test/test_astbuilder.py
```diff
@@ -1988,3 +1988,16 @@ def test_constant_override_do_not_warns_when_defined_in_module_docstring(systemc
     assert ast.literal_eval(attr.value) == 99
     captured = capsys.readouterr().out
     assert not captured
+
+@systemcls_param
+def test__name__equals__main__is_skipped(systemcls: Type[model.System]) -> None:
+    mod = fromText('''
```
adiroiban, line 1994: I think that it would be nice to have a docstring decribing the purpose and scope of this test.
When you end up with hundred of tests is hard to tell which tests is for which use case/functionality.

```suggestion
def test__name__equals__main__is_skipped(systemcls: Type[model.System]) -> None:
    """
    The code included inside the C{__main__} block is excluded from the API docs.
    Any outside code is part of the API.
    """
    mod = fromText('''
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1988,3 +1988,16 @@ def test_constant_override_do_not_warns_when_defined_in_module_docstring(systemc
...
+    if __name__ == '__main__':
+        foo = True
```
adiroiban, line 1996: Maybe add a minimal external API to shows that the presence of `if __main__` only ignores the code inside the conditional block and does not ignores the whole module.

```suggestion
    PUBLIC_API = 'some value'
    if __name__ == '__main__':
        foo = True
```

README.rst
```diff
@@ -73,6 +73,7 @@ What's New?
 in development
 ^^^^^^^^^^^^^^
 * Smarter line wrapping in summary and parameters tables.
+* Fix that variables in ``if __name__ == '__main__'`` blocks were documented.
```
adiroiban, line 76: very minor comment

```suggestion
* Any code inside ``if __name__ == '__main__'`` is now excluded from the API documentation.
```

adiroiban:
Changes looks good. I left a few minor comments.
Let me know when this is ready for a full review.
Many thanks!

not-my-profile:
I made the changes you suggested but I think the CI is stuck.

adiroiban:
I think this is just a bug in CI reporting.

I see the run here https://github.com/twisted/pydoctor/actions/runs/1788711484

It looks good to me

@tristanlatr  can you take a quick look at this and merge if you are happy ? Thanks

adiroiban (approved):
Looks good. Thanks!

pydoctor/astutils.py
```diff
@@ -34,3 +35,26 @@ def bind_args(sig: Signature, call: ast.Call) -> BoundArguments:
         if kw.arg is not None
         }
     return sig.bind(*call.args, **kwargs)
+
+
+if sys.version_info[:2] >= (3, 8):
```
tristanlatr, line 40: You could use `literal_eval` to avoid a checking the python version, I think.

tristanlatr (commented):
Just a few questions.

Thanks for all your energy @not-my-profile .

pydoctor/astutils.py
```diff
@@ -34,3 +35,26 @@ def bind_args(sig: Signature, call: ast.Call) -> BoundArguments:
...
+    # Since Python 3.8 "foo" is parsed as ast.Constant.
+    def _is_str_constant(expr: ast.expr, s: str) -> bool:
+        return isinstance(expr, ast.Constant) \
+        and expr.value == '__main__'
+else:
+    # Before Python 3.8 "foo" was parsed as ast.Str.
+    def _is_str_constant(expr: ast.expr, s: str) -> bool:
+        return isinstance(expr, ast.Str) \
+        and expr.s == '__main__'
```
tristanlatr, line 49: Also it seem that the current implementation ignores the parameter `s`.

```suggestion
def _is_str_constant(expr: ast.expr, s: str) -> bool:
    try:
        value = ast.literal_eval(expr)
    except Exception:
        # literal_eval can raise a variety of exceptions...
        return False
    else:
        return value==s
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1988,3 +1988,25 @@ def test_constant_override_do_not_warns_when_defined_in_module_docstring(systemc
...
+
+@systemcls_param
+def test__name__equals__main__is_skipped(systemcls: Type[model.System]) -> None:
+    """
+    Code inside of C{if __name__ == '__main__'} should be skipped.
+    """
+    mod = fromText('''
+    foo = True
+
+    if __name__ == '__main__':
```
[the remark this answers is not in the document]
not-my-profile: No it does not. Neither does switching the order (`"__main__" == __name__`). But to be honest I don't think it should work for those cases since both are unidiomatic.

pydoctor/astutils.py
not-my-profile: I think I prefer what I have currently since `ast.literal_eval` can raise so many exceptions. And we don't actually need to evaluate any expressions IMO.

not-my-profile:
I fixed the ignored `s` parameter and answered your questions.

tristanlatr:
Thanks! 

Looks OK to me, I’ll merge this. 

Talk to you taker.

PR #475, opened 2022-02-04
Fix variables named like modules not being documented

### LINKED ISSUE #474: Variables named like module are not documented
For a file `example.py` with the content:

```python
example = True
```

image

(note the absence of the `example` variable)

not-my-profile:
Fixes #474.


tristanlatr (approved):
Looks good, but merge conflicts needs to be resolved

pydoctor/astbuilder.py
```diff
@@ -518,7 +518,7 @@ def _handleModuleVar(self,
         parent = self.builder.current
         obj = parent.resolveName(target)
         
-        if obj is None:
+        if obj is None or isinstance(obj, model.Module):
```
tristanlatr, line 521:
```suggestion
        if obj is None or obj.fullName() != f'{parent.fullName()}.{target}':
```

Maybe this check is more generic, what do you think @adiroiban ?

not-my-profile: I find that check harder to understand, so if we were to use it I think it would need to be explained in a comment.

tristanlatr: Ok, now that I think about it, this is probably not the best to handle this situation. 
I'm not sure exactly why `resolveName` is used in this context. I think we should avoid using `resolveName` from the builder since it expect all object in the tree to be already created in order to work correctly. 
I've edited this suggestion.

pydoctor/astbuilder.py
```diff
@@ -527,7 +527,7 @@ def _handleModuleVar(self,
         parent = self.builder.current
         obj = parent.resolveName(target)
```
tristanlatr, line 528:
```suggestion
        obj = parent.contents.get(target)
```

tristanlatr:
Looks great for me. 

I’ll give a few days for other maintainers to review if they want and I’ll merge this PR next week.

Thanks @not-my-profile

not-my-profile:
I clarified the commit & PR that this is about variables named like the **current** module. Because variable names still conflict with submodule names (as they should).

PR #479, opened 2022-02-04
Compact module index

### LINKED ISSUE #465: Collapsing modules with many submodules in the Module Index
Some Python modules have hundreds of submodules (e.g. `pygments.lexers` or `searx.engines`). If all of these are displayed on the Module Index the index becomes hard to use ... whatever comes alphabetically after the module containing the many submodules is only reachable after much scrolling and it's unnecessarily difficult to get an overview of the overall module structure (which I'd argue is the main purpose of the Module Index).

So I think it would be a good idea to automatically omit submodules from the Module Index for modules that contain more than a specific amount of modules (maybe 50 by default?). If deemed necessary that amount could be made configurable. Ideally the Module Index would still indicate something like "(200 submodules)" to not give the wrong impression that the module is empty.

What do you think about this?

not-my-profile:
| Before | After |
|--|--|
| image | image

Closes #465.


pydoctor/templatewriter/summary.py
```diff
@@ -26,13 +26,26 @@ def moduleSummary(module: model.Module, page_url: str) -> Tag:
...
     def fullName(obj: model.Documentable) -> str:
         return obj.fullName()
+
+    if len(contents) > 50 and not any(any(s.submodules()) for s in contents):
+        # If there are more than 50 modules and no submodule has
+        # further submodules we use a more compact presentation.
+        pass
+        div = tags.div(class_='submodules')
+        for m in sorted(contents, key=fullName):
+            div(tags.code(tags.a(m.name, href=m.url)))
```
tristanlatr, line 42: Please use epydoc2stan.taglink to create links to objects in the system.

not-my-profile: Why? Of the three `if`s in epydoc2stan.taglink none will ever apply:

* `if not o.isVisible`: not the case because we filter modules for visibility before
* `if label is None:` we would need to specify the label since we don't want the full name
* `if url.startswith(page_url + '#'):` also will never apply

So using it seems quite pointless to me.

tristanlatr: Because we want to centralize the link generation at only one place such that we can add custom css for internal vs external links without too much hazard. See #468

pydoctor/templatewriter/summary.py
```diff
@@ -26,13 +26,26 @@ def moduleSummary(module: model.Module, page_url: str) -> Tag:
...
         return r
-    ul = tags.ul()
+
     def fullName(obj: model.Documentable) -> str:
         return obj.fullName()
+
+    if len(contents) > 50 and not any(any(s.submodules()) for s in contents):
+        # If there are more than 50 modules and no submodule has
+        # further submodules we use a more compact presentation.
+        pass
```
tristanlatr, line 39:
```suggestion
```

tristanlatr (changes requested):
I’m not sure that we absolutely need css customization for this corner case.
Couldn’t we do with a simple list item? 

Also, I know it’s a pain, but it would be good to have a test for that new corner case. 

About the model refractor, I think it’s good to have this new method. I’de like to read @adiroiban opinion on this.

pydoctor/templatewriter/summary.py
```diff
@@ -26,13 +26,26 @@ def moduleSummary(module: model.Module, page_url: str) -> Tag:
...
-                if isinstance(m, model.Module) and m.isVisible]
+    contents = list(module.submodules())
     if not contents:
         return r
-    ul = tags.ul()
+
     def fullName(obj: model.Documentable) -> str:
         return obj.fullName()
+
+    if len(contents) > 50 and not any(any(s.submodules()) for s in contents):
```
tristanlatr, line 36: What if there is 50 sub modules and one package? 
I think the 50 du modules should still be grouped together while the packages would be extended.

Also packages are instances of model.Package, so maybe it could work with an `isinstance` check instead ?

pydoctor/templatewriter/summary.py
```diff
@@ -26,13 +26,26 @@ def moduleSummary(module: model.Module, page_url: str) -> Tag:
...
+        # If there are more than 50 modules and no submodule has
+        # further submodules we use a more compact presentation.
+        pass
+        div = tags.div(class_='submodules')
```
tristanlatr, line 40: Maybe we could read a text like: « 50 submodules: <the list>»

not-my-profile: I am afraid I don't know what you mean?

tristanlatr: I'm suggesting to show a text saying how many modules are listed under a single list item, such that reader have a better view of the tree. Also cannot misunderstand the entry as being only one module.

pydoctor/templatewriter/summary.py
not-my-profile: I believe that having more than 50 submodules and one package is really uncommon. I don't think I have ever seen that in a Python package, have you? If not I'd rather keep it simple.

not-my-profile:
> I’m not sure that we absolutely need css customization for this corner case.
Couldn’t we do with a simple list item?

I changed it to a list item. Though I still have two lines of CSS:

* `list-style: none` to disable the list bullet (because one bullet on the module index represents one module, using one bullet for multiple modules would give the wrong impression)
* `line-height: 1.8em` because otherwise the text is very vertically bunched together which makes it hard to read

> Also, I know it’s a pain, but it would be good to have a test for that new corner case.

I added a test case.

pydoctor/templatewriter/summary.py
not-my-profile: Now that the submodules are displayed in a list item the vertical bar is displayed which IMO really helps with indicating that these are modules. So I don't think adding a label is necessary.

pydoctor/templatewriter/summary.py
tristanlatr: Very well

pydoctor/themes/base/apidocs.css
```diff
@@ -114,6 +114,11 @@ li {
     padding-bottom: 5px;
 }
 
+.submodules {
```
tristanlatr, line 117:
```suggestion
/*special css class used only in the modules index when there is more than 50 modules on the same package*/
.submodules {
```

not-my-profile: I updated the CSS selector to `#summaryTree .compact-modules` which is more descriptive. Is that descriptive enough or should I still add a comment? I certainly think we should avoid mentioning the exact number in the CSS, it's an implementation detail that is subject to change.

tristanlatr (approved):
Looks good for me! Thanks

pydoctor/templatewriter/summary.py
```diff
@@ -26,15 +26,31 @@ def moduleSummary(module: model.Module, page_url: str) -> Tag:
...
+    if len(contents) > 50 and not any(any(s.submodules()) for s in contents):
+        # If there are more than 50 modules and no submodule has
+        # further submodules we use a more compact presentation.
+        li = tags.li(class_='compact-modules')
+        for m in sorted(contents, key=fullName):
+            span = tags.span()
+            span(tags.code(epydoc2stan.taglink(m, m.url, label=m.name)))
+            span(', ')
+            if m.isPrivate:
+                span(class_='private')
```
not-my-profile, line 47: I updated the PR to add the `private` class for private modules, which is necessary because otherwise they're always shown.

tristanlatr: Thanks, I did not see that bug. It looks like the whole span will be hidden though. Where only one tag link and the commas should be private I think.

not-my-profile: Yes it it's intentional that the whole span is hidden because otherwise there would be ,,,,,, left over.

tristanlatr: Sorry it’s my fault I though is was one span for the whole thing.

pydoctor/templatewriter/summary.py
```diff
@@ -26,15 +26,32 @@ def moduleSummary(module: model.Module, page_url: str) -> Tag:
...
+            li(span)
+        # remove the last trailing comma
+        li.children[-1].children.pop() # type: ignore
+        ul(li)
+    else:
+        for m in sorted(contents, key=fullName):
```
tristanlatr, line 53:
```suggestion
        for m in sorted(contents, key=objects_order):
```

pydoctor/test/test_templatewriter.py
```diff
@@ -469,3 +469,28 @@ def func():
...
+    ul = moduleSummary(top, '').children[-1]
+    assert ul.tagName == 'ul'       # type: ignore
+    assert len(ul.children) == 50   # type: ignore
+
+    # the 51th module triggers the compact summary
+    fromText('', parent_name='top', modname='yet_another_sub', system=system)
+
+    ul = moduleSummary(top, '').children[-1]
+    assert ul.tagName == 'ul'       # type: ignore
+    assert len(ul.children) == 1    # type: ignore
```
tristanlatr, line 489:
```suggestion
    fromText('', parent_name='top', modname='_yet_another_sub', system=system)

    ul = moduleSummary(top, '').children[-1]
    assert ul.tagName == 'ul'       # type: ignore
    assert len(ul.children) == 1    # type: ignore
    
    # test that the last module is private
    assert 'private' in ul.children[0].children[-1].attributes['class'] # type: ignore
```

pydoctor/templatewriter/summary.py
```diff
@@ -26,15 +26,32 @@ def moduleSummary(module: model.Module, page_url: str) -> Tag:
...
+    if len(contents) > 50 and not any(any(s.submodules()) for s in contents):
+        # If there are more than 50 modules and no submodule has
+        # further submodules we use a more compact presentation.
+        li = tags.li(class_='compact-modules')
+        for m in sorted(contents, key=fullName):
+            span = tags.span()
+            span(tags.code(epydoc2stan.taglink(m, m.url, label=m.name)))
+            span(', ')
+            if m.isPrivate:
+                span(class_='private')
```
tristanlatr, line 47: Added a suggestion in the test case to cover this branch.

adiroiban:
@tristanlatr if you guys are happy with the changes, feel free to merge merge it.

I have very little interest into pydoctor and generic API doc tools.
I always prefer to read the source code and most api docs is part of the source code.

I know that API doc tools are important.

I am helping with this project just because Twisted is using it and I care about Twisted... and there was no one else reviewing the PR...so I tried to help and unblock the project.

Many thanks to both of you for you help with pydoctor!

tristanlatr:
@not-my-profile are you willing to address the changes? I don't really care about using `objects_order` but a test case for the private modules would be good. 

Thanks

README.rst
```diff
@@ -108,6 +108,7 @@ pydoctor 22.2.0
   this allows to have the external intershinx links blue while the internal links are red.
 * Smarter line wrapping in summary and parameters tables.
 * Any code inside of ``if __name__ == '__main__'`` is now excluded from the documentation.
+* The module index now uses a more compact presentation for modules with more than 50 submodules and no subsubmodules.
```
tristanlatr, line 111: I'll move this to the right place.

tristanlatr (approved):

PR #481, opened 2022-02-06
Display name instead of full name in Module Index

not-my-profile:
| before | after |
|--|--|
|image | image

I think it makes the index more readable (especially for modules deeper down the hierarchy).


tristanlatr (approved):
Looks good for me. Just missing a change log entry.

not-my-profile:
I added a change log entry. I also added a commit to set the full name as a `title` attribute so browsers show it as a tooltip on hover.

tristanlatr (approved):
Looks perfect, thanks a lot.

Issue #430, opened 2021-09-02
Big refactoring of the model and astbuilder


tristanlatr:
Hello, 

I'm working on a huge refactoring of the model and the AST builder. 

I've roughly proposed the changes here https://github.com/twisted/pydoctor/issues/295#issuecomment-903052019, I'll try to go in more details here, and hopefully get a consensus around this and move forward. 

The core of my proposition is the following:
- Have 2 layers of model objects, One layer that will be the "true" structure of the API, represented and built with [pydocspec](https://github.com/tristanlatr/pydocspec), I'm porting the astbuilder to it, but it might take some time still. It is more "true" than the second model (`Documentable`), because `pydocspec` will not reparent objets when visiting import statements, it simply creates a special `Indirection` object (instead of using a dict to keep track of imports). Object that can later be followed by the `ApiObject.resolve_name()` method. (Note that I've basically adapted all `Documentable` features related to python language processing to `pydocspec`) 
- Once this model is created, we build the `Documentable` objects from that, improve the re-exporting system and create a new class doing the bridge between the `pydocspec` word and the `Documentable` word. Let's call it `DocumentableLookup` and it would be a `System` attribute (for the other way arround, simply `Documentable.spec:pydocspec.ApiObject` shoud be ok). It would have to be created at the time of the conversion `pydocspec`->`Documentable` , could have the following interface: 
```python
class DocumentableLookup:
    def add(ob: pydocspec.ApiObject, documentable: model.Documentable) -> None:
        ...
    def remove(ob: pydocspec.ApiObject, documentable: model.Documentable) -> None:
        ...
    def get(ob: pydocspec.ApiObject) -> Optional[model.Documentable]:
        ...
    def getall(ob: pydocspec.ApiObject) -> Optional[List[model.Documentable]]:
        ...
```
(we can probbly adapt the code of the [`ReverseMap`](https://github.com/NiklasRosenstein/docspec/blob/e7a2ad8cc2e07ad2f4689cedcb50a0f69dd4d37f/docspec/src/docspec/__init__.py#L246) to fit our needs.)
- This class will be used, for instance, by the method `Documentable.resolveName()` such that the lookup happens in the `pydocspec` word, but be translated in the `Documentable` word before returning the result. 

What are the motivations for such a massive change?:
- Firstly it comes from the motivation for fixing issue #295, 
[...]

PR #483, opened 2022-02-06
Add workaround for missing baseobjects

### LINKED ISSUE #482: There's a random chance that the superclass isn't linked
To reproduce this we need a package with three files:

`example/__init__.py`

```python
from . import bar
```

`example/foo.py`

```python
from . import xxx

class Foo(object):
	pass
```

`example/bar.py`

```python
from . import foo

class Bar(foo.Foo):
    pass
```

Running `pydoctor example` has a random chance that it fails to link the superclass:

| correct | bugged |
|--|--|
|image|image

The bug is difficult to debug since it doesn't always occur, running pydoctor with `PYTHONHASHSEED=0` or `PYTHONHASHSEED=1` does however seem to help with reproducing.

The underlying issue is that how pydoctor [currently handles cyclic imports is not reproducible](https://github.com/twisted/pydoctor/issues/138#issuecomment-730394192).

not-my-profile:
This is just a workaround that seems to work. If we fix our AST builder to be reproducible we can probably just remove this, but for now it should fix the missing base classes.

Closes #482.

### DIFF
--- a/pydoctor/driver.py
+++ b/pydoctor/driver.py
@@ -430,6 +430,15 @@ def main(args: Sequence[str] = sys.argv[1:]) -> int:
 
         system.process()
 
+        # Cyclic imports can cause Class.baseobjects to contain None values for
+        # base classes that actually exist, see issue #482. This is a simple workaround
+        # for this more complex problem.
+        classes = {cls.fullName(): cls for cls in system.objectsOfType(model.Class)}
+        for cls in classes.values():
+            for idx, base in enumerate(cls.bases):
+                if cls.baseobjects[idx] is None and base in classes:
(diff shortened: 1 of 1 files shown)


tristanlatr:
Not too sure about this workaround... In any case, it should come with a test.

You could put this code inside the post processing method of the system instead of the driver. It’s not currently the best place for that.

Also, I’m working on a another project that should replace our ast builder. See #430. So this issue will be properly fixed with this refactor hopefully. But it’s good to have it fixed like that too. Actually it’s perfect if we can have a solid test case.

talk to you later

not-my-profile:
Oh I think I discovered the root cause for the lack of reproducibility.

PR #484, opened 2022-02-06
Fix random module processing order

### LINKED ISSUE #429: Inconsistent behaviour in inherited members
There is a strange bug in pydoctor right now that makes subclasses sometimes fail to resolve their base classes names. 

This is probably happening because of the order the modules get processed?

I'm currently running into this issue for the `pydocspec` repo, See the script the builds the docs: https://github.com/tristanlatr/pydocspec/blob/main/apidocs.sh

Sometimes the informations do not show up:

by default 2021-08-23 at 11 32 44 PM

But at the next run of `apidocs.sh`, it does show up :

by default 2021-08-23 at 11 31 21 PM

When it works, all the objects works, but when it doesn't, none does.  

Do you have any insights @mthuurne @adiroiban ?

Thanks

### LINKED ISSUE #482: There's a random chance that the superclass isn't linked

not-my-profile:
Previously modules to be processed were stored in a set.
A set however is an unordered collection (and adding items
can completely change the order depending on the object hash).

Since the Module class did not implement `__hash__` every time
there was more than one module to be processed the order
in which they were processed was effectively random leading
to very hard to reproduce bugs.

Fixes #482 and fixes #429.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,6 +74,7 @@ in development
 ^^^^^^^^^^^^^^
 * Smarter line wrapping in summary and parameters tables.
 * Any code inside of ``if __name__ == '__main__'`` is now excluded from the documentation.
+* Fixed that modules were processed in a random order leading to several hard to reproduce bugs.
 * Fix introspection of functions comming from C-extensions.
 
 pydoctor 21.12.1
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -596,7 +596,7 @@ class System:
         self.verboselevel = 0
         self.needsnl = False
         self.once_msgs: Set[Tuple[str, str]] = set()
-        self.unprocessed_modules: Set[Module] = set()
+        self.unprocessed_modules: List[Module] = []
         self.module_count = 0
         self.processing_modules: List[str] = []
         self.buildtime = datetime.datetime.now()
@@ -789,7 +789,7 @@ class System:
         self.progress(
             "analyzeModule", len(self.allobjects),
             None, "modules and packages discovered")
-        self.unprocessed_modules.add(mod)
+        self.unprocessed_modules.append(mod)
         self.module_count += 1
         self.setSourceHref(mod, modpath)
         return mod
--- a/pydoctor/test/test_cyclic_imports_base_classes.py
+++ b/pydoctor/test/test_cyclic_imports_base_classes.py
@@ -1,0 +1,39 @@
+"""
+This test case is in its own file because it requires the
+PYTHONHASHSEED=0 environment variable. See issue #482.
+"""
+
+import os
+import subprocess
+import sys
+
+def test_cyclic_imports_base_classes() -> None:
+    if sys.platform == 'win32':
+        # Running this script with the following subprocess call fails on Windows
+        # with an ImportError that isn't actually related to what we want to test.
+        # So we just skip for Windows.
+        return
+
+    process = subprocess.Popen(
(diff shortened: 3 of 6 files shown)


tristanlatr:
I think we should still use something that behaves like a set. I think it was designed to be a set for a reason. What do you think @adiroiban ? This code is 15 years old at least : https://github.com/twisted/pydoctor/commit/60321c0f4ef2c5b6e137de0abb400c95fe6e50f3

Maybe we can use something like an `OrderedSet`. But we should avoid extra pypi dépendues when possible.

not-my-profile:
No using something that behaves like a set would not change anything. The only values that are ever added to `unprocessed_modules` are newly created Packages or Modules. Since `Package` and `Module` do not implement `__hash__` they will always have a different hash.

tristanlatr:
Ok, looks OK to me. Did you try if that changes anything with the Class base objects issue?

not-my-profile:
> Did you try if that changes anything with the Class base objects issue?

I think it fixes that bug.

> Do you think it would be possible to add a test?

I added a test case.

tristanlatr:
Are you sure this test fails without the fix ? Did you try ? Or there is still a random chance that the test passes ?

not-my-profile:
I thought I did but I missed something. I have now fixed the test case: it reliably fails without my fix.

README.rst
```diff
@@ -74,6 +74,7 @@ in development
 ^^^^^^^^^^^^^^
 * Smarter line wrapping in summary and parameters tables.
 * Any code inside of ``if __name__ == '__main__'`` is now excluded from the documentation.
+* Fixed that modules were processed in a random order leading to several hard to reproduce bugs.
```
adiroiban, line 77: Minor comment
I don't understand this release note.
The release note is designed for end users to know whether they should upgrade or not.

---------

Maybe something like this

* The internal processing of modules was updated to an algorithm used fixed order. This should help fix some of the bugs that previously seemed to happen for no apparent reason.

Or just keep it internal and don't have a release note.

pydoctor/model.py
```diff
@@ -789,7 +789,7 @@ def analyzeModule(self,
         self.progress(
             "analyzeModule", len(self.allobjects),
             None, "modules and packages discovered")
-        self.unprocessed_modules.add(mod)
+        self.unprocessed_modules.append(mod)
```
adiroiban, line 792: Now that we have a list, should this check for duplicate modules ?

At the same time, the de-duplicate can be done later when we start processing the unprocessed modules.

But I prefer to check for duplication at append time.

But maybe the way `analyzeModule` works could not lead to duplicate modules.

adiroiban (approved):
Looks good.

A list should be ok.

To be safe, I would check for duplicate members before calling append.
This should have a similar behavior to `set`.

PR #486, opened 2022-02-06
Add .gitattributes to classify bootstrap.min.css as binary

not-my-profile:
Stops bootstrap.min.css from polluting git grep results.

E.g. `git grep table`.

### DIFF
--- a/pydoctor/themes/classic/.gitattributes
+++ b/pydoctor/themes/classic/.gitattributes
@@ -1,0 +1,1 @@
+bootstrap.min.css binary


tristanlatr (approved):
Very well. Thanks.

adiroiban:
Thanks. Another option is to move it to git-lfs.

For my python projects using boostrap CSS I have created something similar to this package https://pypi.org/project/bootstrap4

is just that the package follows the boostrap versioning ... and it's a wheel

PR #487, opened 2022-02-07
If there is just a single root module, document it in index.html

not-my-profile:
I think most of the time `pydoctor` is just invoked with a single module. In that case it's nicer if `index.html` immediately shows the module documentation instead of just a page linking the module.

### DIFF
--- a/pydoctor/templatewriter/writer.py
+++ b/pydoctor/templatewriter/writer.py
@@ -90,6 +90,18 @@ class TemplateWriter(IWriter):
                 flattenToFile(fobj, page)
             system.msg('html', "took %fs"%(time.time() - T), wantsnl=False)
 
+        if len(system.root_names) == 1:
+            # If there is just a single root module it is written to index.html to produce nicer URLs.
+            # To not break old links we also create a symlink from the full module name to the index.html
+            # file. This is also good for consistency: every module is accessible by <full module name>.html
+            root_module_path = (self.build_directory / (list(system.root_names)[0] + '.html'))
+            try:
+                root_module_path.unlink()
+                # not using missing_ok=True because that was only added in Python 3.8 and we still support Python 3.6
+            except FileNotFoundError:
+                pass
+            root_module_path.symlink_to('index.html')
+
     def _writeDocsFor(self, ob: model.Documentable) -> None:
         if not ob.isVisible:
             return
@@ -97,7 +109,7 @@ class TemplateWriter(IWriter):
             if self.dry_run:
                 self.total_pages += 1
             else:
-                with self.build_directory.joinpath(f'{ob.fullName()}.html').open('wb') as fobj:
+                with self.build_directory.joinpath(ob.url.split('#')[0]).open('wb') as fobj:
                     self._writeDocsForOne(ob, fobj)
         for o in ob.contents.values():
             self._writeDocsFor(o)
--- a/pydoctor/templatewriter/writer.py
+++ b/pydoctor/templatewriter/writer.py
@@ -109,7 +109,7 @@ class TemplateWriter(IWriter):
             if self.dry_run:
                 self.total_pages += 1
             else:
-                with self.build_directory.joinpath(ob.url.split('#')[0]).open('wb') as fobj:
+                with self.build_directory.joinpath(ob.url).open('wb') as fobj:
                     self._writeDocsForOne(ob, fobj)
         for o in ob.contents.values():
             self._writeDocsFor(o)
--- a/pydoctor/templatewriter/writer.py
+++ b/pydoctor/templatewriter/writer.py
@@ -82,7 +82,7 @@ class TemplateWriter(IWriter):
 
     def writeSummaryPages(self, system: model.System) -> None:
         import time
-        for pclass in summary.summarypages:
+        for pclass in summary.summaryPages(system):
             system.msg('html', 'starting ' + pclass.__name__ + ' ...', nonl=True)
             T = time.time()
             page = pclass(system=system, template_lookup=self.template_lookup)
--- a/pydoctor/test/test_templatewriter.py
+++ b/pydoctor/test/test_templatewriter.py
@@ -6,7 +6,7 @@ import sys
 import tempfile
 import os
 from pathlib import Path, PurePath
-from pydoctor import model, templatewriter, stanutils
+from pydoctor import model, templatewriter, stanutils, __version__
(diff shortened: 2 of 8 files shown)


pydoctor/templatewriter/writer.py
```diff
@@ -97,7 +97,7 @@ def _writeDocsFor(self, ob: model.Documentable) -> None:
             if self.dry_run:
                 self.total_pages += 1
             else:
-                with self.build_directory.joinpath(f'{ob.fullName()}.html').open('wb') as fobj:
+                with self.build_directory.joinpath(ob.url.split('#')[0]).open('wb') as fobj:
```
tristanlatr, line 100: Why this change ?

tristanlatr:
@not-my-profile could you stop force pushing from the time we start reviewing your code? you can use regular merge instead of rebase and then we can squash everything together at the end. It’s a bit hard to follow your changes.

not-my-profile:
I don't actually want you to squash my commits. E.g. for this PR i intentionally split the refactor into a separate commit.

I think you should wait with reviewing until the CI succeeded ... if it fails I'll force push a fix.

tristanlatr:
> I don't actually want you to squash my commits. 

Ok.

Force pushing messes with the review. I’ll wait for the tests to be green to review. 

But if there are changes to be done after the first review, i’d like those changes not to be forced pushed. Such that I can review only the new changes.

tristanlatr:
Does it breaks links that were pointing to the single root ? Maybe, in case of single root, index.html should only be a redirect to the module page (url unchanged). What do you think ?

not-my-profile:
Right, I'll mark my PRs as drafts in the future until they're ready for review. Just FYI I always try to split up my changes into atomic commits to make them easier to review. So I recommend reviewing my changes commit by commit.

pydoctor/driver.py
```diff
@@ -489,6 +489,12 @@ def p(msg: str) -> None:
                 for fn in sorted(system.docstring_syntax_errors):
                     p('    '+fn)
 
+            if len(system.root_names) == 1:
+                # If there is just a single root module it is written to index.html to produce nicer URLs.
+                # To not break old links we also create a symlink from the full module name to the index.html
+                # file. This is also good for consistency: every module is accessible by <full module name>.html
+                (build_directory / (list(system.root_names)[0] + '.html')).symlink_to('index.html')
```
not-my-profile, line 496: I just added a check to create a symlink so that old links outside of pydoctor don't break.

pydoctor/templatewriter/writer.py
```diff
@@ -82,22 +82,30 @@ def writeIndividualFiles(self, obs: Iterable[model.Documentable]) -> None:
...
                 flattenToFile(fobj, page)
             system.msg('html', "took %fs"%(time.time() - T), wantsnl=False)
 
+        if len(system.root_names) == 1:
+            # If there is just a single root module it is written to index.html to produce nicer URLs.
+            # To not break old links we also create a symlink from the full module name to the index.html
+            # file. This is also good for consistency: every module is accessible by <full module name>.html
+            root_module_path = (self.build_directory / (list(system.root_names)[0] + '.html'))
+            root_module_path.unlink()
+            root_module_path.symlink_to('index.html')
```
not-my-profile, line 99: I just added a check to create a symlink so that old links outside of pydoctor don't break. I initially tried adding the check in driver but then the symlink creation fails because some tests use an `InMemoryWriter`. So having the check in the `TemplateWriter` class makes more sense.

not-my-profile:
Ok this is now ready for review :) Btw @tristanlatr did you get my mail?

pydoctor/templatewriter/writer.py
```diff
@@ -82,22 +82,34 @@ def writeIndividualFiles(self, obs: Iterable[model.Documentable]) -> None:
...
+
     def _writeDocsFor(self, ob: model.Documentable) -> None:
         if not ob.isVisible:
             return
         if ob.documentation_location is model.DocLocation.OWN_PAGE:
             if self.dry_run:
                 self.total_pages += 1
             else:
-                with self.build_directory.joinpath(f'{ob.fullName()}.html').open('wb') as fobj:
+                with self.build_directory.joinpath(ob.url.split('#')[0]).open('wb') as fobj:
```
tristanlatr, line 112: Not sure we absolutely need to split on `#`. Since `ob.documentation_location is model.DocLocation.OWN_PAGE`, `page_object` will return `self` and nothing will be added after the filename in the URL. 

I think the following is OK:

```suggestion
                # Documentable.page_object.url will alway be a plain filename.
                with self.build_directory.joinpath(ob.page_object.url).open('wb') as fobj:
```

tristanlatr (commented):
Only a minor comment. 

I really like this feature. Thanks

pydoctor/test/test_templatewriter.py
```diff
@@ -104,7 +104,7 @@ def test_basic_package(tmp_path: Path) -> None:
         if '#' in url:
             url = url[:url.find('#')]
         assert (tmp_path / url).is_file()
-    with open(tmp_path / 'basic.html', encoding='utf-8') as f:
+    with open(tmp_path / 'index.html', encoding='utf-8') as f:
```
tristanlatr, line 107: Isn’t this supposed to still work with “basic.html” ?

pydoctor/templatewriter/writer.py
not-my-profile: I think splitting here is cleaner, because otherwise there is an undocumented constraint on the `url` method. If you really want to avoid that split we'd need to document that in the `url` method that it should not set an anchor for "page objects".

tristanlatr: Yes, to mt it's pretty clear the that the url method that it should not set an anchor for "page objects".

pydoctor/test/test_templatewriter.py
not-my-profile: Yes good point. I made this change initially since I didn't create the symlinks initially. I have remove this change.

pydoctor/templatewriter/writer.py
not-my-profile: Right. I still prefer such assumptions to be documented. I feel like not depending on such behavior is even better.

tristanlatr: You could quickly add a note to the url property docstring saying that page object should always return a plain relative file name, without anchor.

not-my-profile: Done.

tristanlatr (approved):
Thanks

PR #489, opened 2022-02-09
List packages & modules together

### LINKED ISSUE #480: Stop separating packages and modules on package pages
Pydoctor currently shows packages before modules:

image

I find this really irritating. As an API user I don't really care about the distinction between packages and modules, I just want to find the module/package I need quickly and for that it would help if modules/packages were listed together alphabetically.

Within a package subpackages and submodules share the same namespace so I think displaying them together makes the most sense. For example `sphinx` and `sphinx_ext` shouldn't be 9 rows apart they should be next to each other.

What do you think about this? If we display them intermixed they should of course both have the same background color.

not-my-profile:
| before | after |
|--|--|
|| image

Closes #480.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,6 +79,7 @@ in development
 * Any code inside of ``if __name__ == '__main__'`` is now excluded from the documentation.
 * Fix variables named like the current module not being documented.
 * The Module Index now only shows module names instead of their full name. You can hover over a module link to see the full name.
+* Packages and modules are now listed together (they were separated previously).
 * Fix introspection of functions comming from C-extensions.
 
 pydoctor 21.12.1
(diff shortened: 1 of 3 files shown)


not-my-profile:
If a non-draft PR of mine passes the pipeline you can assume it's ready for review. So this is ready for review :)

tristanlatr:
What about keeping a different color? Would it be too confusing ?

not-my-profile:
Yes I'd find it confusing. Especially because the color difference is so subtle. And I don't think the distinction is important enough to pick colors with a bigger contrast.

adiroiban:
> What about keeping a different color? Would it be too confusing ?

I also think that extra color is too much... in the end packages are just modules with extra semantic.

tristanlatr:
Very well!

tristanlatr:
@not-my-profile are you willing to address the required changes? 
Having 100% coverage diff would be good.

not-my-profile:
Sure, you're welcome to pick it up. If I understand correctly you just want to keep the separate colors? Like Adi I'd prefer them to have the same color, but I don't really care that much about this.

tristanlatr:
replaced by #535

PR #490, opened 2022-02-09
Patch out blockquote font-size from bootstap.min.css

not-my-profile:
```python
"""
some text

    some blockquote

some text
"""

__docformat__ = 'restructuredtext'
```

| before | after |
|--|--|
| image | image

Bootstrap 3.x sets the font-size of blockquotes to 17.5px
which just looks ridiculously large compared to regular text.

This commit simply removes that font-size from bootstrap.min.css.

### DIFF
--- a/pydoctor/themes/base/apidocs.css
+++ b/pydoctor/themes/base/apidocs.css
@@ -390,6 +390,13 @@ a.internal-link {
     color: #971c3a;
 }
 
+/* Bootstart 3.x sets font-size to 17.5px which just
+   looks ridiculously large, so we unset it here.
+*/
+blockquote {
+    font-size: unset;
+}
+
 /*
 This defines the code style, it's black on light gray.
 It also overwrite the default values inherited from bootstrap min


adiroiban (commented):
Thanks for the update.

To make this manageable I think that bootstrap.min.css should be left as it is and update the custom pydoctor CSS code.

I think that the `bootstrap.min.css` should be called something like

`pydoctor/themes/classic/vendor/bootstrap-v3.3.4.min.css`

This should hint that the css file is vendored and the only allowed operation is replacing it with an upstream released version.

It's hard to check this diff.

and wasn't this file supposed to be binary and no diff?

not-my-profile:
Yes I agree that putting it in another css file would be more readable but to me it seems wrong to put a bootstrap-specific fix in `base/apidocs.css` since `base` doesn't depend on bootstrap. And I don't think we want to include yet another css file because it results in an additional HTTP request.

tristanlatr:
I suggest to add it this adjustment to `apidocs.css` with an appropriate comment. 
It’s already pretty coupled with bootstrap anyway.

adiroiban:
>  And I don't think we want to include yet another css file because it results in an additional HTTP request.

Do you run pydoctor on a high traffic site?

Should we really bother about that one extra request? Just asking.

There is one extra request, but after that, the file should be cached.

If we are worried about performance, we can create an helper to concatenate all the CSS or JS files at build time ... poor man's webpack :) 

But I think that messing with boostrap.min.css is not the right way (tm)

Later we might want to update boostrap and we might forget about this change.

And if you have a custom theme, I guess that you will not use bootsrat.css or any of the base css at all.

My understanding is that boostrap is part of the base/default pydoctor theme.

Is not like we have a clean/reset base theme and then we have another boostrap theme.

just my 2 cents :)

tristanlatr:
> And if you have a custom theme, I guess that you will not use bootsrat.css or any of the base css at all.

I think custom themes are likely to be built on top of the `base` theme, like read the docs theme. But still, 3 lines of css with appropriate comment on `apidocs.css` seems OK to me. 

We might want to reconsider this at some point, but then we would need to clearly separate what belongs to the `base` theme and what belongs to the `classic` theme. Currently the only difference is that the classic theme uses bootstrap, the additional css is the same as in the base theme.

tristanlatr (changes requested):
Please move the changes to apidocs.css with the appropriate comment.

not-my-profile:
Done.

tristanlatr (approved):
Thanks!

PR #491, opened 2022-02-11
Colorize directives ``.. code::`` and ``.. code-block::``

### LINKED ISSUE #413: Colorize code inside ``..code ::`` directives
The code currently only displayed as a literal block, but no syntax highlight is applied. 
I think we could re-use the doctest syntaxt highlight and apply it to the ``..code ::`` directive, too.

tristanlatr:
Fixes #413


not-my-profile:
Can't these directives also be used with non-Python code?

tristanlatr:
No not right now.

tristanlatr:
To include a code bloc without colorization, just use `::`. I’ll add that to the docs.

not-my-profile:
I meant that e.g. `.. code:: html` should not be colorized as Python.

tristanlatr:
Currently it will.

not-my-profile:
Yes we can add it later. What I meant was that this PR should check the argument of the directive and only highlight it as Python when there is no argument or the argument == "python". When the argument is sth else it should not be highlighted as Python.

tristanlatr:
That would require more work to adopt such behavior. I’m unsure if that should be part of this PR.

glyph (approved):
This all looks good to me, assuming CI is happy I think we can land it.

tristanlatr:
Thanks @glyph, I'll merge this PR and create a follow-up issue for other languages.

PR #493, opened 2022-02-15
Bugfix the colorizer string escape function

tristanlatr:
Pydoctor 21.12.1 crashes when building the python standard library docs because of a `UnicodeEncodeError` :/ 

My bad. Thanks @not-my-profile for the report.

### DIFF
--- a/.github/workflows/system.yaml
+++ b/.github/workflows/system.yaml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        tox_target: [twisted-apidoc, cpython-summary, cpython-apidocs]
+        tox_target: [twisted-apidoc, cpython-summary]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/system.yaml
+++ b/.github/workflows/system.yaml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        tox_target: [twisted-apidoc, cpython-summary]
+        tox_target: [twisted-apidoc, cpython-summary, cpython-apidocs]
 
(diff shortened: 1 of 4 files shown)


adiroiban (approved):
Looks good. Thanks!

tristanlatr:
Thanks @adiroiban, 

I think follow-up question about these types of caracters inside `re.compile` strings: 

Currently for the input `re.compile('surrogates:\udc80\udcff')` it colorzise something like `re.compile(r'surrogates:\udc80\udcff')` which is not the same string, but is equal to the `re.escape` version of the string. 

The first thing we migh think is: if it's not the same string, then it's wrong, but atually, the `SubPattern` dump of the two strings is exactly the same:

```python
>>> sre_parse.parse('surrogates:\udc80\udcff').dump()
LITERAL 115
LITERAL 117
LITERAL 114
LITERAL 114
LITERAL 111
LITERAL 103
LITERAL 97
LITERAL 116
LITERAL 101
LITERAL 115
LITERAL 58
LITERAL 56448
LITERAL 56575
>>> sre_parse.parse(r'surrogates:\udc80\udcff').dump()
LITERAL 115
LITERAL 117
LITERAL 114
LITERAL 114
LITERAL 111
LITERAL 103
LITERAL 97
LITERAL 116
LITERAL 101
LITERAL 115
LITERAL 58
LITERAL 56448
LITERAL 56575
```

This is why I think the current behavuour is OK, but I just wanted to hear your opinion and have that on record.

adiroiban:
I think is ok. As long as you are happy, I am happy :p Thanks for the merge!

PR #494, opened 2022-02-17
More system tests

tristanlatr:
This PR only adds more tests


tox.ini
```diff
@@ -67,27 +67,42 @@ commands =
...
 description = Build CPython API documentation
 
-extras = rst
-
 commands =
     sh -c "if [ ! -d {toxworkdir}/cpython ]; then \
         git clone --depth 1 https://github.com/python/cpython.git {toxworkdir}/cpython; \
         fi"
-    sh -c "cd {toxworkdir}/cpython && git pull"
+    sh -c "cd {toxworkdir}/cpython && git fetch --tags && git checkout v3.9.10"
```
tristanlatr, line 74:
```suggestion
    sh -c "cd {toxworkdir}/cpython && git pull"
```

adiroiban (changes requested):
Thanks for the update.

I see that the new tests are failing.

I am not -1 on this.

I think that we ended up with twisted system tests as the pydoctor own test suite was not good enough.

The problem with these system tests is that we will depend too much on external repos, especially if we run the tests on a continuously changing external repo.

If you want to have them, it's ok.

And we can see how this goes.

But first we should have green tests for this PR :)

tristanlatr:
Yes, system tests ensure we're not crashing on complex code/corner cases. If we had that before, the colorizer would not have been bugged in the `21.12` version (see https://github.com/twisted/pydoctor/pull/493).

tristanlatr:
This change adds 15mn before any PR can be merged, it's not a very big deal IMO.

tox.ini
```diff
@@ -66,28 +66,45 @@ commands =
...
+    pytest
 commands =
     sh -c "if [ ! -d {toxworkdir}/cpython ]; then \
         git clone --depth 1 https://github.com/python/cpython.git {toxworkdir}/cpython; \
         fi"
     sh -c "cd {toxworkdir}/cpython && git pull"
     touch {toxworkdir}/cpython/Lib/__init__.py
     rm -rf {toxworkdir}/cpython-output
-    pydoctor \
+    - pydoctor \
```
adiroiban, line 78: I don't remember what this does...but I hope that if pydoctor fails here, the whole tox env run also fails.

docs/tests/__init__.py
```diff
@@ -0,0 +1,8 @@
+from pathlib import Path
+import os 
+
+def get_toxworkdir_subdir(subdir:str) -> Path:
+    dir = Path(os.environ['TOX_WORK_DIR']).joinpath(subdir) \
+        if os.environ.get('TOX_WORK_DIR') else Path(os.getcwd()).joinpath(f'./.tox/{subdir}')
+    assert dir.exists(), f"Looks like {dir} not not exist!"
+    return dir
```
adiroiban, line 8:
```suggestion
    return dir

```

adiroiban (approved):
All good. Thanks. Only minor comments.

tox.ini
tristanlatr: Not it doesn't because pydoctor always returns a non-zero status code when building standard libraty docs. The actual checks are in the python code.

adiroiban: ok... I think that twisted system tests were added with the goal at reaching a point at which when running pydoctor on twisted no warnings or errors are generated...and in this way we can better check for any regressions.

if  you are happy with the current tests, I am also happy :)

tristanlatr: Yes it’s still working like that for twisted, it just fails for the standard library and python-igraph. But that’s ok since we have tests.

Issue #10, opened 2015-01-12
methods inherited via more than one class are documented more than once


mwhudson:
See for example http://twistedmatrix.com/documents/9.0.0/api/twisted.internet.interfaces.IReadWriteDescriptor.html

I have this vague memory that there is code that's supposed to suppress the repeated documentation, but it clearly isn't working.

---

Imported from Launchpad using lp2gh.
- date created: 2010-02-21T22:57:04Z
- owner: mwhudson
- the launchpad url was https://bugs.launchpad.net/bugs/525521

mthuurne:
Assuming this is the problem, it still occurs with the latest pydoctor:

> Inherited from IFileDescriptor (via IReadDescriptor):
> ...
> Inherited from IFileDescriptor (via IWriteDescriptor):

So `IFileDescriptor` is listed as a superclass twice as a result of a diamond-shaped inheritance. However, `ILoggingContext` is listed only once, despite it being reachable via the same two routes:

> Inherited from ILoggingContext (via IReadDescriptor, IFileDescriptor):

tristanlatr:
We should probably adapt our own version of the C3 algorithm that works with `model.Class` objects. 

A nice implementation can be found at https://github.com/pilosus/c3linear

PR #495, opened 2022-02-17
Resolve name check super class

tristanlatr:
This PR solves the same issue as #380, but it adds the feature inside the `expandName` logic, which is to me, the right place.

### DIFF
--- a/pydoctor/test/test_model.py
+++ b/pydoctor/test/test_model.py
@@ -347,3 +347,14 @@ def test_c_module_python_module_name_clash(capsys:CapSys) -> None:
     finally:
         # cleanup
         subprocess.getoutput(f'rm -f {package_path}/*.so')
+
+def test_resolve_name_subclass(capsys:CapSys) -> None:
+    m = fromText(
+        """
+        class B:
+            v=1
+        class C(B):
+            pass
+        """
+    )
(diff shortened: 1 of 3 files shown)


pydoctor/test/test_model.py
```diff
@@ -346,4 +346,15 @@ def test_c_module_python_module_name_clash(capsys:CapSys) -> None:
 
     finally:
         # cleanup
-        subprocess.getoutput(f'rm -f {package_path}/*.so')
\ No newline at end of file
+        subprocess.getoutput(f'rm -f {package_path}/*.so')
+
+def test_resolve_name_subclass(capsys:CapSys) -> None:
```
adiroiban, line 351: can we have a docstring that describe what is tested here?

Maybe also have a test for sub-sub class inheritance  ... or maybe is not needed as we have good enough tests for our `Class.find` method.

```suggestion
def test_resolve_name_subclass(capsys:CapSys) -> None:
    """
    C{Model.resolveName} knows about single inheritance.
    """
```

adiroiban (approved):
Looks good. Thanks!

tristanlatr:
I think we're good with `find()` are related methods tests. We don't get the MRO right anyway see https://github.com/twisted/pydoctor/issues/10

Issue #421, opened 2021-07-25
Use docutils transformer to resolve links


tristanlatr:
Using a custom [Transform](https://tristanlatr.github.io/apidocs/docutils/docutils.transforms.Transform.html), we could apply this transformation in-place to `ParsedDocstring` documents. 

The main advantage of this approach is to be able to clearly separate the xref linking step from the HTML rendering. 

Initial discussion here: https://github.com/twisted/pydoctor/pull/386#discussion_r671392572.

tristanlatr:
This means major refactoring of the docstrig linker in order to make it work with docutils nodes instead of stan.

Issue #497, opened 2022-02-17
Pydoctor is slow if there are many subclasses (caching for summary stan)


not-my-profile:
```
$ wget -O nodes.py https://sourceforge.net/p/docutils/code/HEAD/tree/tags/docutils-0.17.1/docutils/nodes.py?format=raw
$ time pydoctor nodes.py
...
pydoctor nodes.py  14.44s user 0.27s system 99% cpu 14.744 total
```

`nodes.py` contains 133 classes, generating documentation for it with pydoctor takes over 14 seconds on my machine.

It seems like Pydoctor is doing a bunch of work (parsing reST, resolving links in reST) for each subclass, e.g:

```python
class paragraph(General, TextElement): pass
class compound(General, Element): pass
class container(General, Element): pass
class bullet_list(Sequential, Element): pass
class enumerated_list(Sequential, Element): pass
class list_item(Part, Element): pass
class definition_list(Sequential, Element): pass
class definition_list_item(Part, Element): pass
class term(Part, TextElement): pass
class classifier(Part, TextElement): pass
class definition(Part, Element): pass
class field_list(Sequential, Element): pass
class field(Part, Element): pass
class field_name(Part, TextElement): pass
class field_body(Part, Element): pass
```

I think it would make sense to introduce some caching to avoid doing the same work multiple times.

tristanlatr:
I get your point. I think the solution would be to cache the results of the EpydocLinker. In order to do that, we need to store and associate one EpydocLinkers to one context object (maybe simply getLinker() method on all Documentables), implement caching on the linker and reuse the same linker every time we need to link within that context. One exception would be the summary generation, they would use a different Linker instance because of #478

tristanlatr:
I haven’t run benchmark to see if the linker is actually a bottleneck, but it definitely could be, because we’re doing a lot of funcky stuff to resolve links.

tristanlatr:
I think what's slow if the calls to `flatten` function.  And re-generating the summaries of each object all the time. I have some =thing reasy to fix this issue as well as fixes #478 by improving the linker by providing caching. 

After that, the summaries should also be cached IMO.

tristanlatr:
See discussion about docstring linker caching here: https://github.com/twisted/pydoctor/pull/498#issuecomment-1050516761

[...]

PR #498, opened 2022-02-18
Provides caching for the linker

### LINKED ISSUE #478: Links in docstring summary break when the summary is included on another page
```python
"""
Provides L{foo}.
"""

def foo():
    pass
```

image

On that page `foo` links to `#foo` but it should link to `bar.html#foo`.

It's worth noting that ReST links break even worse:

```python
"""
Implements `Foo`_

.. _Foo: http://example.com
"""
```

results in

image

with `` `Foo`_ `` linking to `#id1` for some reason.

tristanlatr:
This PR tries to solve #497, it partially succeeds but we also need to cache the whole Documentable summary stan.  

It does fixes #478


tristanlatr:
Note: It does not seem to influence the performance of pydoctor, but many warnings are now displayed only once and not many times! so that's good.

adiroiban:
Thanks for the update.

I am a bit busy these days. if @not-my-profile can review this, that would be great :)

pydoctor/model.py
```diff
@@ -373,6 +375,18 @@ def report(self, descr: str, section: str = 'parsing', lineno_offset: int = 0) -
...
+    def docstringlinker(self) -> 'epydoc2stan.DocstringLinker':
+        """
+        Returns an instance of L{epydoc2stan.DocstringLinker} suitable for resolving names
+        in the context of the object scope. 
+        """
+        if self._linker is not None:
+            return self._linker
+        from pydoctor.epydoc2stan import _CachedEpydocLinker
+        self._linker = _CachedEpydocLinker(self)
+        return self._linker
```
tristanlatr, line 388: The best would be to avoid this workaround cyclic imports. 
In order to do that, we should move all code related to the concrete docstring linkers in a new module: `docstringlinker.py`. I wonder if this change should happen in this PR though ?

tristanlatr:
@adiroiban, I don't think @not-my-profile wants to be involved in pydoctor development. He manifested interest into creating a new doc generator based on mypy internals. So I don't think we should wait for his review.

pydoctor/epydoc2stan.py
```diff
@@ -74,11 +77,23 @@ def taglink(o: model.Documentable, page_url: str, label: Optional["Flattenable"]
         ret(title=o.fullName())
     return ret
 
-
+# TODO: Move this code to it's own module to avoid cyclic imports.
```
adiroiban, line 80: Is this TODO for this PR, or is there a follow-up issue created to track its progress?

pydoctor/epydoc2stan.py
```diff
@@ -119,31 +134,46 @@ def look_for_intersphinx(self, name: str) -> Optional[str]:
         return self.obj.system.intersphinx.getLink(name)
 
     def link_to(self, identifier: str, label: "Flattenable") -> Tag:
+        # :Raises _EpydocLinker.LookupFailed: If the identifier cannot be resolved and self.strict is True.
```
adiroiban, line 137: Maybe have this as docstrig instead of comment?

pydoctor/epydoc2stan.py
```diff
@@ -119,31 +134,46 @@ def look_for_intersphinx(self, name: str) -> Optional[str]:
...
             return self._create_intersphinx_link(label, url=url)
 
-        return tags.transparent(label)
+        link = tags.transparent(label)
+        if self.strict:
+            raise self.LookupFailed(identifier, link=link)
+        return link
 
     def link_xref(self, target: str, label: "Flattenable", lineno: int) -> Tag:
+        # :Raises _EpydocLinker.LookupFailed: If the identifier cannot be resolved and self.strict is True.
```
adiroiban, line 156: Can we have it as docstring?

pydoctor/model.py
adiroiban: Better for a separate PR, but also add all this info in a code comment :)

pydoctor/test/test_epydoc2stan.py
```diff
@@ -86,16 +86,31 @@ def test_xref_link_not_found() -> None:
 
 def test_xref_link_same_page() -> None:
     """A linked name that is documented on the same page is linked using only
-    a fragment as the URL.
+    a fragment as the URL. But that does not happend in summaries. 
     """
-    mod = fromText('''
+    src='''
```
adiroiban, line 91: We might want to enable `black` to create a common style for the code formatting :)
```suggestion
    src = '''
```

adiroiban (approved):
Hi,Thanks for the update.

If there is no speedup, I am not sure if it's worth it.

Also by not having repeated errors, it might be ok when you run locally, but for CI, I prefer to see all the errors so that I can fix them in one run. 

So.. I am now +1 for this change, but you are the maintainer and if you want to have this code, I am ok with that.

Thanks again!

tristanlatr:
> Also by not having repeated errors, it might be ok when you run locally, but for CI, I prefer to see all the errors so that I can fix them in one run.

We are talking about the same errors again and again for the summary table generation, not different errors. Just run 

```
$ wget -O nodes.py https://sourceforge.net/p/docutils/code/HEAD/tree/tags/docutils-0.17.1/docutils/nodes.py?format=raw
$ time pydoctor nodes.py
```

And you'll see what I mean. See https://github.com/twisted/pydoctor/issues/497

pydoctor/epydoc2stan.py
tristanlatr: Another PR, It's pretty simple change.

tristanlatr: I'll create an issue.

pydoctor/epydoc2stan.py
tristanlatr: Will do

tristanlatr: Actually, I remember. I did not si that not to override the docstirng inherited by `ParsedDocstring`. 
Since this an implementation detail because we're always using the `_CachedEpydocLinker`, and I did not want to change the pretty well documented docstring that we already have for `link_to` and `link_xref`.

tristanlatr:
> If there is no speedup, I am not sure if it's worth it.

@adiroiban,

I've thought to give up this change and instead implement caching for summaries and docstring stan. But here's the deal: there is not way to simply ignore xref warnings when we call `parse_docstring`, so warnings are getting reported when we generate stan for the summary, so at least invalid links present in the fist three lines of docstrings will be reported twice even if we cache the docstring and summary stan (and even if we fix https://github.com/twisted/pydoctor/issues/86. Actually we'de need to have both https://github.com/twisted/pydoctor/issues/86 and https://github.com/twisted/pydoctor/issues/421 fixed and compute the summary document from a version with already resolved links in it). So maybe this patch is just the wrong way of dealing with this problem.

I'm unsure what's the best way forward. The class `_CachedEpydocLinker` is 200 lines of codes and the same in tests. But tests can be recycled anyway. Yes it's a bit of complexity (and multiple invalid link on the same line are reported only once, but it's better than reporting warnings X times the number of summaries we generate for that object. Also if you fix one link in one line in the docstring it's likely that you'll notice the other invalid link that is using the same target on the same line). What I mean is this code can be used until we have fixed https://github.com/twisted/pydoctor/issues/421, https://github.com/twisted/pydoctor/issues/86 and implemented docstring stan caching, then we can remove the `_CachedEpydocLinker` and simply replace all occurences with `_EpydocLinker`.

Please tell me what you think, ok I'm maintainer but I do want to fix issue this the right way.

Thanks.

tristanlatr:
Do you have comments @adiroiban ?
Otherwise we would proceed with the proposed plan.

adiroiban:
I am not -1 on this.
And reducing noise in the output is +1.
So I think that this can be merge.

Thanks!

----------

Sorry... I didn't had time to look into pydoctor for twisted usage and why CI doesn't fail an warnings.

tristanlatr:
Great, thanks!

PR #499, opened 2022-02-19
Release pydoctor 22.2.0

tristanlatr:
I think it's time for a new release. 

Now, we can safely advertise that we support introspection of c-extensions.


adiroiban (approved):
Go, go, go!

All good. Thanks!

PR #500, opened 2022-02-21
JS search as you type

tristanlatr:
This PR adds a search function that works locally with Javascript and searches as you type.

Do a test here: https://pydoctor--500.org.readthedocs.build/en/500/api/

This PR replaces https://github.com/twisted/pydoctor/pull/381


pydoctor/themes/base/pydoctor.js
```diff
@@ -22,8 +22,31 @@ function updatePrivate() {
...
+            if (hidden){
+                new_location = new_location + '?';
+            }
+            else{
+                new_location = new_location + '&';
+            }
+            new_location = new_location + search_params;
+        }
+
+        history.replaceState(null, '', new_location + document.location.hash);
```
tristanlatr, line 49: This change is not required

pydoctor/themes/base/nav.html
```diff
@@ -1,28 +1,99 @@
...
-          <a href="nameIndex.html">
-            Names
-          </a>
+        <a href="nameIndex.html">
+          Names
+        </a>
+
+        <div id="search-box-container">
+          <div class="input-group">
+            <input id="search-box" type="text" name="search-query" placeholder="Search..." aria-label="Search" minlength="2" class="form-control"/>
```
not-my-profile, line 27: We definitely want to set `autocomplete=off` because otherwise browsers may show their own autocompletion on top of the JS autocompletion.

not-my-profile:
Testing on the rtd link you posted, clicking inside the search box shows an empty suggestion box which feels really weird.

tristanlatr:
Thanks for your feedback @not-my-profile

pydoctor/themes/base/nav.html
tristanlatr: I think `type=search` disables that.

not-my-profile: No it does not ... at least not in Firefox, I still get the Firefox autocompletion.

tristanlatr: Is it very bad? Because right now, the ways it works it's not really autocompletion... For instance, if you type "pydoctor.tes" there will be no results... So maybe it makes sens to keep it that way?

But it will show up if we search for "pydoctor.tes*". So maybe there is something to do ?

Let's note that "pydoctor.test" gives one result whereas "pydoctor.test*" gives 592 results

not-my-profile: Yes it is very bad. It suggests searches I have performed on completely different sites.

tristanlatr: What do you mean? I don’t get it

tristanlatr: Hum, I get your point

adiroiban (approved):
Hi. Sorry for the delay.
This looks awesome.

I will not have time to review this in the next weeks.. Ukraine war is affecting us all :(

I did a very quick check and it looks good.
So I think that it can be merged.

Thanks again.

tristanlatr:
Thanks

PR #502, opened 2022-02-23
Bugfix release 22.2.1

### LINKED ISSUE #501: Crash when building numpy docs
Pydoctor crashes when building numpy docs. 

This must be a regression introduced by #488 :/ 

I'll work on a fix and release a bugfix version of pydoctor, as well as a system test for numpy docs

```
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/pydoctor/driver.py", line 431, in main
    system.process()
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/pydoctor/model.py", line 1024, in process
    self.processModule(mod)
  File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/pydoctor/model.py", line 1013, in processModule
    del self.unprocessed_modules[mod.fullName()]
KeyError: 'numpy.core.char'
```

tristanlatr:
Fixes #501 and publish a new release

### DIFF
--- a/pydoctor/test/test_packages.py
+++ b/pydoctor/test/test_packages.py
@@ -90,3 +90,16 @@ def test_package_module_name_clash() -> None:
     system = processPackage('package_module_name_clash')
     pack = system.allobjects['package_module_name_clash.pack']
     assert 'package' == pack.contents.popitem()[0]
+
+def test_reparented_module() -> None:
+    """
+    Reparented modules does not crash pydoctor, and are correctly reparented like any other objects.
+    """
+    system = processPackage('reparented_module')
+
+    mod = system.allobjects['reparented_module.module']
+    top = system.allobjects['reparented_module']
+
+    assert mod.fullName() == 'reparented_module.module'
+    assert top.resolveName('module') is top.contents['module']
+    assert top.resolveName('module.f') is mod.contents['f']
--- a/pydoctor/test/test_packages.py
+++ b/pydoctor/test/test_packages.py
@@ -106,7 +106,7 @@ def test_reparented_module() -> None:
     assert top.resolveName('module') is top.contents['module']
     assert top.resolveName('module.f') is mod.contents['f']
 
-    # The module is not in allobjects
+    # The module old name is not in allobjects
     assert 'reparented_module.mod' not in system.allobjects
     # But can still be resolved with it's old name
     assert top.resolveName('mod') is top.contents['module']
--- a/pydoctor/test/test_packages.py
+++ b/pydoctor/test/test_packages.py
@@ -105,3 +105,4 @@ def test_reparented_module() -> None:
     assert mod.fullName() == 'reparented_module.module'
     assert top.resolveName('module') is top.contents['module']
     assert top.resolveName('module.f') is mod.contents['f']
+    
--- a/pydoctor/test/test_packages.py
+++ b/pydoctor/test/test_packages.py
@@ -93,7 +93,9 @@ def test_package_module_name_clash() -> None:
 
 def test_reparented_module() -> None:
     """
-    Reparented modules does not crash pydoctor, and are correctly reparented like any other objects.
+    A module that is imported in a package as a different name and exported
+    in that package under the new name via C{__all__} is presented using the
+    new name.
     """
     system = processPackage('reparented_module')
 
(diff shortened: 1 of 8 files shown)


pydoctor/test/testpackages/reparented_module/mod.py
```diff
@@ -0,0 +1,2 @@
+def f():
+    pass
```
adiroiban, line 2:
```suggestion
"""
This is the "origin" module which for testing purpose is used from the C{reparented_module} package.
"""
def f():
    pass

```

pydoctor/test/testpackages/reparented_module/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import mod as module
+__all__=('module',)
```
adiroiban, line 2:
```suggestion
"""
Here the module C{mod} is made available under an alias name
that is explicitly advertised under the alias name. 
"""
from . import mod as module
__all__=('module',)

```

pydoctor/test/test_packages.py
```diff
@@ -89,4 +89,17 @@ def test_package_module_name_clash() -> None:
...
     system = processPackage('package_module_name_clash')
     pack = system.allobjects['package_module_name_clash.pack']
-    assert 'package' == pack.contents.popitem()[0]
\ No newline at end of file
+    assert 'package' == pack.contents.popitem()[0]
+
+def test_reparented_module() -> None:
+    """
+    Reparented modules does not crash pydoctor, and are correctly reparented like any other objects.
+    """
```
adiroiban, line 97: Minor comment.

Rather than talking about the crash, I think that is best to describe the
expected behaviour.

And maybe also assert that that original name is no longer availble?
Or reparented is just an alias and both the original and the new names are available?

```suggestion
    """
    A module that is imported in a package as a different name and exported
    in that package under the new name via C{__all__} is presented using the
    new name.
    """
```

tox.ini
```diff
@@ -82,6 +82,23 @@ commands =
         {toxworkdir}/cpython/Lib
     pytest -v docs/tests/test_standard_library_docs.py
 
+[testenv:numpy-apidocs]
+description = Build numpy API documentation
```
adiroiban, line 86: Maybe add a short info to introduce the purpose and scope of this test.

```suggestion
description = Build numpy API documentation. For now we don't check for any warnings or other errors. The only purpose of this test is to make sure pydoctor doesn't crash.
```

adiroiban (approved):
Looks good. Only minor comments.

I have not reviewed the logic for reparented, but as long as the test are ok , it should be good.

Thanks!

tox.ini
```diff
@@ -82,6 +82,23 @@ commands =
...
+        git clone --depth 1 https://github.com/numpy/numpy.git {toxworkdir}/numpy; \
+        fi"
+    sh -c "cd {toxworkdir}/numpy && git pull"
+    rm -rf {toxworkdir}/numpy-output
+    python3 -c "from pydoctor.driver import main; \
+        code = main(['--html-summary-pages', '--quiet', \
+        '--html-output={toxworkdir}/numpy-output', \
+        '{toxworkdir}/numpy/numpy']); \
+        assert code==2, 'pydoctor exited with code %s, expected code 2.'%code"
+        # Code 2 error means bad docstrings, which is OK for this test.
```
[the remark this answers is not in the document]
tristanlatr: > Maybe try to work with numpy upstream to not get bad docstring ?

Haha... imagine? me sending PR to all repositories that have bag docstrings... No. 

Here's what I'm working on: https://pydocbrowser.github.io/

tristanlatr:
Thanks @adiroiban for your review, alway very useful. 

I'll release a new version now.

adiroiban:
Nice work for https://pydocbrowser.github.io/

I first saw this centralized documentation model in golang https://pkg.go.dev/

No need to send upstream report for each package :)

I can suggest the inclusion of 

* https://github.com/twisted/twisted/
* https://github.com/twisted/klein
* https://github.com/twisted/treq
* https://github.com/twisted/ldaptor

tristanlatr:
@adiroiban I've added the packages, thanks

PR #506, opened 2022-02-24
Bugfix release 22.2.2

### LINKED ISSUE #505: reparenting processing should follow aliases
Just compare https://twistedmatrix.com/documents/21.2.0/api/twisted.web.template.html (valid) and https://docs.twistedmatrix.com/en/latest/api/twisted.web.template.html (latest, invalid). 

Bunch of important stuff is just omitted, unsure why yet.

We did not caught that before because our intersphinx link for twisted points to old API site that is using pydoctor 20.12.1, so our links resolved correctly. 

Edit: Maybe this regression has been introduced in version 21.2.0, actually

@adiroiban

tristanlatr:
Fixes #505 and releasing a new version.


pydoctor/test/testpackages/reparenting_follows_aliases/_myotherthing.py
```diff
@@ -0,0 +1 @@
+from ._mything import MyClass
```
adiroiban, line 1: Can you add a docstring to explain what is the difference between "myotherthing" and "mything"  ? :)

And empty newlines at the end of files :)

pydoctor/test/testpackages/reparenting_follows_aliases/main.py
```diff
@@ -0,0 +1,3 @@
+from ._myotherthing import MyClass
```
adiroiban, line 1: A docstring to explain what happens here and what was the intention to setup this test would help.

docs/tests/test_twisted_docs.py
```diff
@@ -23,3 +23,32 @@ def test_IPAddress_implementations() -> None:
     with open(BASE_DIR / 'twisted.internet.interfaces.IAddress.html') as stream:
         page = stream.read()
         assert all(impl in page for impl in show_up), page
+
+# Test for https://github.com/twisted/pydoctor/issues/505
+def test_web_template_api() -> None:
+    """
+    This test ensures all important members of the twisted.web.template
```
adiroiban, line 30: I am not a big fan of theses tests.
I guess that this should be fine, as Twisted API is kind of fixed and we will not see any big refactoring soon.

But rather than write these tests, I prefer to just run the system tests and make sure no warnings are raised and fail on any warning.

If there is an issue with Twisted upstream I can help to get a fix fast.

adiroiban (approved):
Many thanks.

I can see that these errors were reported on the 22.2.1 PR.
Is just that we run the Twisted system tests but we don't do any automatic checks.

https://github.com/twisted/pydoctor/runs/5306393732?check_suite_focus=true#step:6:18

This is the main reason why I wanted to run Twisted checks inside pydctor and fail on any warnings.

There is not much use in running Twisted systems test on pydoctor is the result is not checked...and you can expect developers to manually check the output for regression  :)

adiroiban:
22.2.2 :) this should be a good one :)

docs/tests/test_twisted_docs.py
tristanlatr: I think the main issue here is that twisted doesn’t not check for pydoctor warnings.

tristanlatr: > This is the main reason why I wanted to run Twisted checks inside pydctor and fail on any warnings.

Yes I thought this is what we did in `tox -e twisted-apidoc`

adiroiban:
> Yes I thought this is what we did in tox -e twisted-apidoc

Thanks for the release!

Let's update the pydoctor used by Twisted, see the test fail on Twisted... then fix Twisted documentation.

PR #509, opened 2022-03-03
More search tweaks

tristanlatr:
- Fix warning message displayed when a search resturns no results.
- Add ability to filter by kind.
- Stem name indentifiers, i.e. 'DocumentableKind' -> 'DocumentableKind Documentable Kind' for better indexing.
- Include break points in python names likes we do in table.
- Tweak boots for the stemed names to be lower than the actual name.
- Generate two indexes, search with the small index by default and search in the full index when user click "search in docstrings" button. Making the initial search much faster when we're dealing with a lot of items.
- Rename field fullName to qname because it’s shorter and without case
- Performance tweaks
- Provide hackable Js functions
- Move the javascript imports to `footer.html`, this means that Twisted should update it's `footer.html` custom template to use the search bar.
- Update version of HTML templates ``nav.html`` and ``footer.html``.


pydoctor/templatewriter/search.py
```diff
@@ -98,4 +114,12 @@ def get_ob_boost(ob: model.Documentable) -> int:
     with output_dir.joinpath('searchindex.json').open('w', encoding='utf-8') as fobj:
         fobj.write(serialized_index)
 
+def stem_identifier(_t: str) -> Iterator[str]:
+    yield _t
+    parts = epydoc2stan._split_indentifier_parts_on_case(_t)
+    for p in parts:
+        p = p.strip('_')
+        if p and p.lower() not in stop_word_filter.WORDS and p!=_t: 
+            yield p
```
tristanlatr, line 123:
```suggestion
def stem_identifier(identifier: str) -> Iterator[str]:
    yield identifier
    parts = epydoc2stan._split_indentifier_parts_on_case(identifier)
    for p in parts:
        p = p.strip('_')
        if p and p.lower() not in stop_word_filter.WORDS and p!=identifier: 
            yield p
```

pydoctor/test/test_epydoc2stan.py
```diff
@@ -1248,3 +1252,27 @@ def test_insert_break_points_snake_case() -> None:
...
     assert epydoc2stan.insert_break_points('__einÜberlangerName__') == '__ein\u200bÜberlanger\u200bName__'
+
+def test_insert_break_points_dotted_name() -> None:
+    assert epydoc2stan.insert_break_points('mod.__some_very_long_name__') == 'mod\u200b.__some\u200b_very\u200b_long\u200b_name__'
+    assert epydoc2stan.insert_break_points('_mod.__SOME_VERY_LONG_NAME__') == '_mod\u200b.__SOME\u200b_VERY\u200b_LONG\u200b_NAME__'
+    assert epydoc2stan.insert_break_points('pack.mod.__someVeryLongName__') == 'pack\u200b.mod\u200b.__some\u200bVery\u200bLong\u200bName__'
+    assert epydoc2stan.insert_break_points('pack._mod_.__einÜberlangerName__') == 'pack\u200b._mod_\u200b.__ein\u200bÜberlanger\u200bName__'
+
+def test_stem_identifier() -> None:
+    from pydoctor.templatewriter.search import stem_identifier
```
tristanlatr, line 1263: This import can go at the top of the file.

pydoctor/themes/base/search.js
```diff
@@ -1,5 +1,14 @@
...
 // It's included in every HTML file.
+// Ideas for improvments: 
+// - Include filtering buttons:
+//    - search only in the current module
+//    - have a query frontend that helps build complex queries
+// - Filter out results that have score > 0.001 by default and show them on demand.
+// - Should we change the default term presence to be MUST and not SHOULD ?
+//        -> Hack something like 'name index -value' -> '+name +index -value'
+//        ->      'name ?index -value' -> '+name index -value'
+// - Highlight can use https://github.com/bep/docuapi/blob/5bfdc7d366ef2de58dc4e52106ad474d06410907/assets/js/helpers/highlight.js#L1
```
tristanlatr, line 11: Indexing docstrings for projects like Twsited might not be best idea becaue of the overhead it creates at searching time. 

```suggestion
// - Highlight can use https://github.com/bep/docuapi/blob/5bfdc7d366ef2de58dc4e52106ad474d06410907/assets/js/helpers/highlight.js#L1
// - Since searching in docstrings can be long, i.e. a search in Twisted docs currently takes few seconds.
//    It would be good to be able to search in an index with names only first, and launch docstring search optionnaly.
```

tristanlatr (commented):
One major thing that would be good to have is the ability to not to index docstrings, and only names. That could be useful for very big projects.

tristanlatr:
As you can see here: https://pydocbrowser.github.io/twisted/latest/

The search is currently barely workable - it took about 10 seconds to show that there is not results for search "+type +mail +search" - with Twisted codebase :/ 
Generating an index with the names only will probably make this better. 

It would be good to have the ability to generate two types of indexes (and adjust JS search for it): 
- small: contains index for `fullName` and `name` only
- full: contains index for `fullName`, `name`, `docstring` and `kind`.

Maybe the kind can be added to the small index as well.

We have a couple of options regarding how to deal with this situation in terms of UX and implementation :

- Introduce a new option to use the small index (use full index by default).
- Introduce a new option to use the full index (use small index by default).
- Determine at what size a full index becomes unworkable and fallback to small in these cases (we could introduce an option to write the full index anyway).
- Generate both indexes, search with the small index by default and search in the full index when user click "Search in docstrings" button.

The first three options are relatively easy in terms of implementation. But the last one would be the best in terms of user experience, because we do not loose any of the features, it's just one click away and 80% of the time it will suffice.

I'll do some tests with the Twisted code, hosted on github pages (I was testing with Twisted locally, but turns out it's much more rapid locally (less than 2 secs in general wich is workable). So I was under a wrong impression regarding search speed).

adiroiban:
Well... I am not expecting for miracles from a JS client-side full-text search engine.


AFAIK, for docs.twistedmatrix.com the RTD server-side search is used.


I expect that most libraries are small, and it should be ok.
And having this client-side search feature makes it easy to host the docs via GiHub pages... no need for Read the docs... and pydoctor with Read the docs is a big of a hack via the extension.

The info button is nice.

The "search in docstring" buttons is not obvious ... the enable/disable state is not clear. (grey vs green) . An explicit radio botton or checkbox would help.
Also, maybe place it in the same line with "Search for "test" yielded 444 results."

tristanlatr:
> AFAIK, for docs.twistedmatrix.com the RTD server-side search is used.

Yes I know, but I wonder it we should actually do #356 or instead ofer an option from the Sphinx UI to "search in API docs" andthat would use our index under the hood. Whaty do you think ?

> The "search in docstring" buttons is not obvious ... the enable/disable state is not clear. (grey vs green) . An explicit radio botton or checkbox would help.
Also, maybe place it in the same line with "Search for "test" yielded 444 results."

I'll look into that.

Thanks for your feedback.

adiroiban:
In terms of search for pydoctor, we can have different deployment scenarios

1. pydoctor standalone
2. pydoctor integrated with Sphinx
3. pydoctor integrated with Sphinx integrated with Read The Docs

I think that the current lunr.js search is for case 1.

For 2, sphinx uses the searchindex.js file ... from what I can see

For 3 by reading the read the docs code, it looks like the  .fjson files are used for Sphinx and raw html for mkdocs

pydoctor/epydoc2stan.py
```diff
@@ -979,29 +979,57 @@ def format_constant_value(obj: model.Attribute) -> "Flattenable":
...
+    
+    if prefix:
+        text_parts[0] = prefix + text_parts[0]
+    if suffix:
+        text_parts[-1] = text_parts[-1] + suffix
 
-    return prefix + text_with_breaks + suffix
+    return text_parts
+
+# TODO: Use tags.wbr instead of zero-width spaces
```
tristanlatr, line 1024: This is for another PR.

tristanlatr:
I think we're ready for the final review. I did not put the "search in docstrings" inline with the text because I don't want the button to change position all the time.

adiroiban (approved):
Looks awesome.

I am happy with the search UI. Excellent functionality.

I don't have time for a full review, but I think that this is ready to merge.

We can always fix via another PR.

Thanks again!

tristanlatr:
Great!

PR #516, opened 2022-03-13
Release pydoctor 22.3.0

tristanlatr:
Alongside minor change: Update head.html classic theme template to use the same viewport.


adiroiban (approved):
All good. Thanks!

Issue #461, opened 2022-01-18
Command line option to show all private objects


iosonofabio:
Related to #50, #327, et al.

One solution to all those problems, if not perfect, would be to have a command-line option (which works also via the sphinx ext) to get rid of the "show/hide private" business and just show everything in the docs, private or otherwise.

Of course, folks might want to have a more granular control, but for many situations this might be a good enough compromise: it's better to show users a little too much than to miss key API objects...

Happy to make a PR if you think it's useful

tristanlatr:
Hi, @iosonofabio,

Would you like the private objects to keep their gray color - and order ? If yes, the right way of doing this is through a custom theme (what you are already doing in python-igraph). This will also be the way if you'de like to get rid of the "show/hide private" business. 

If you'de like private members to be considered public, we need new code. 
See for example of how to customize the `System` class: https://github.com/twisted/pydoctor/issues/388#issuecomment-827900981

See this discussion about adding new options/fields to control object privacy: https://github.com/twisted/pydoctor/issues/36 . This is a ticket that I want to see closed for a long time!

I think we're ready for adding one new option, something like: `--make-public=igraph._igraph` that would be repeatable and taking either the full name of an API object or a fnmatch pattern (which could be "*").

If you are interested into making a PR to add this new option, that would be greatly appreciated! 

We discussed (in issue #36) the option to add a docstring field, it would a bit harder on te implementation side, though. Also, there is another way of working around this: names re-exported by the `__all__` variables are public names in Python. To make the submodule `_igraph` public, just re-rexport it in `igraph` package in the `__all__` variable. We do not currently support that. An issue with that approach (that should be supported anyway) is that all public names must be added to the `__all__` variable unless python will not import your names when doing a wildcard import. 

The command line option seems like a good start to me, what do you say ?

iosonofabio:
Thanks for the quick reply. We're using `__all__` in `python-igraph` at the moment, but I'm thinking more of a catchall situation, where people have no time/resources for making a proper theme, etc.

Yes something like `--make-public` with `action='append'` would work, with perhaps a special `--make-public='*'` value that makes everything public. I'll try and find the time for a PR in the next days, thank you!

tristanlatr:
Thanks @iosonofabio,

The custom theme is really just:

```css
body.private-hidden .private {
    display: inherit!important;
}
```

[...]

PR #517, opened 2022-03-14
Add option --privacy

### LINKED ISSUE #36: pydoctor should be able to ignore 'test' directories
One common layout for python source trees is to put unit tests in a 'test' subdirectory of the code being tested. It would be nice to have a way to ignore these subdirectories when generating documentation.

epydoc has an --exclude option that can be used for this.

---

Imported from Launchpad using lp2gh.
- date created: 2011-05-26T04:36:26Z
- owner: duncf
- the launchpad url was https://bugs.launchpad.net/bugs/788472

tristanlatr:
Add option `--privacy` to set the privacy of specific objects when default rules doesn't fit the use case.

Finally fix #36 and #461.

Tackles part of #315. 

The help:
```
--privacy=<PRIVACY>:<PATTERN>
                        Set the privacy of specific objects when default rules
                        doesn't fit the use case. Format:
                        '<PRIVACY>:<PATTERN>', where <PRIVACY> can be one of
                        'PUBLIC', 'PRIVATE' or 'HIDDEN' (case insensitive),
                        and <PATTERN> is fnmatch-like pattern matching objects
                        fullName. Pattern added last have priority over a
                        pattern added before, but an exact match wins over a
                        fnmatch. Can be repeated.
```

The docs: https://pydoctor--517.org.readthedocs.build/en/517/customize.html#override-objects-privacy-show-hide


tristanlatr:
So a new import in test module made introspection test crash because of recursion error (while introspecting typing.List) :/
I don't think we should maintain the introspection code, but rather rely on something else. Astroid can build AST trees from live modules, so once we enter the new generation of pydoctor with astroid, we won't have to maintain our introspection code anymore. 

So to fix this I'll just do the easiest hack.

tristanlatr:
View example here: https://pydoctor--517.org.readthedocs.build/en/517/api/

The last readthedocs build failed because "Concurrency limit reached (2)", so that's not related with the changes.

tristanlatr:
I feel like in order to be really helpful we should use regexes instead of fnmatch, because there are cases where it could be difficult to work with. Like if we'd like to make all top-level names in a module public, but not stuff under the top level. This is highly impractical with the current implementation. But maybe it's good enough anyway. It satisfies the Twisted use case anyway.

adiroiban:
I don't know what to say about `glob` vs `regex`.

At work for an `end-user product` I have implemented a custom enhanced glob.

I like regex, but for end users I feel that regex is on the other extreme - super powerful and complex.

So, maybe just go with glob and later add some sort of git pattern matching, and further improvements.

And I would say that `--privacy-re` should only be implemented is someone asked for it.. and sends a PR :)

At work, for a progressive regex support I did something like this:
You have the `s/REGEX/FLAGS` format. It's not idea, but in practice I never
had a customer that had a glob expression starting with `regex/` and ending with something  that look like valid regex flags 

```
--privacy="public:regex/twisted.test.proto_helpers.*/i"
```

So I would say , keep it simple :)

---------

Not sure if it should be called `HIDDEN` or `IGNORED`.

I think that the documentation should be extend to document the "standard" privacy types/classes

* Public - Always and visible rendered in HTML
* Private - Rendered in HTML, but hidden via CSS by default
* hidden - No HTML markup is generated and no links are created.

-----------

Not sure if `--privacy=` is the best name for this functionality.

Maybe it can be implemented as a `--tag` that later can be used in a theme to implement custom behaviour.

----------

What I can image, is building the apidocs in 2 versions:

* One version with only public API - this used by the external users / customers
* One version with private API hidden - this is used by the internal development teams


----------

I think that this is on the right direction.

But before doing a full review, it would help to have the documentation, to make sure I understand the scope and target of these changes.

--------

I guess that we have 2 main use cases, and I think that these 2 examples can be part of the docs:

* Make everything public
* Keep normal convention for public and private api, but ignore/exclude/hide the test code

Thanks!

docs/source/customize.rst
```diff
@@ -32,6 +32,47 @@ HTML templates have their own versioning system and warnings will be triggered w
...
+  Basically excluded from API documentation. If a module/package/class is hidden, then all it's members are hidden as well.
+
+When the default rules regarding privacy doesn't fit the use case, you can use the following repeatable option::
+
+  --privacy=<PRIVACY>:<PATTERN>
+
+where ``<PRIVACY>`` can be one of ``PUBLIC``, ``PRIVATE`` or ``HIDDEN`` (case insensitive), and ``<PATTERN>`` is fnmatch-like 
+pattern matching objects fullName.
+
+Pricavy tweak examples
```
adiroiban, line 60:
```suggestion
Privacy tweak examples
```

docs/source/customize.rst
```diff
@@ -32,6 +32,47 @@ HTML templates have their own versioning system and warnings will be triggered w
...
+- ``PRIVATE`` (If object name starts with an underscore and it's not a dunder method): 
+  Rendered in HTML, but hidden via CSS by default.
+
+- ``PUBLIC`` (Everything else):
+  Always rendered and visible in HTML.
+
+- ``HIDDEN`` (Nothing by default):
+  Not rendered at all and no links can be created to hidden objects. 
+  Not present in the search index nor the intersphinx inventory.
+  Basically excluded from API documentation. If a module/package/class is hidden, then all it's members are hidden as well.
```
adiroiban, line 51: Very minor suggestion 

```suggestion
Pydoctor supports 3 types of privacy.
Below is the description of each type and the default association:

- ``PRIVATE`` By default for objects whose name starts with an underscore and are not a dunder method. 
  Rendered in HTML, but hidden via CSS by default.

- ``PUBLIC`` - By default everything else that is not private.
  Always rendered and visible in HTML.

- ``HIDDEN`` - Nothing is hidden by default.
  Not rendered at all and no links can be created to hidden objects. 
  Not present in the search index nor the intersphinx inventory.
  Basically excluded from API documentation. If a module/package/class is hidden, then all it's members are hidden as well.
```

docs/source/customize.rst
```diff
@@ -32,6 +32,47 @@ HTML templates have their own versioning system and warnings will be triggered w
...
+
+When the default rules regarding privacy doesn't fit the use case, you can use the following repeatable option::
```
adiroiban, line 53: Minor suggestion 

```suggestion
When the default rules regarding privacy doesn't fit your use case,
use the `privacy` command line option.
It can be used multiple times to define multiple privacy rules::
```

docs/source/customize.rst
```diff
@@ -32,6 +32,47 @@ HTML templates have their own versioning system and warnings will be triggered w
...
+  Makes everything under ``twisted.test`` hidden except ``twisted.test.proto_helpers``, which will be public.
+  
+- ``--privacy="PRIVATE:**.__*__" --privacy="PUBLIC:**.__init__"``
+  Makes all dunder methods private except ``__init__``.
+
+.. note:: The order of arguments matters. Pattern added last have priority over a pattern added before,
+  but an exact match wins over a fnmatch.
+
+.. note:: See :py:mod:`pydoctor.qnmatch` for more informations regarding the pattern syntax.
+
```
adiroiban, line 75: Not sure if this information should be defined as a note/side-note

It looks like important functional description docs.

adiroiban (approved):
Excellent work. Only minor comments.

Thanks!

tristanlatr:
I’ve added a custom match function and rather complete documentation. Tell me what you think.

> happy to keep it like --privacy . No need for tagging.

Not even sure what you are referring as tagging. The theming system is really only about templates files. Themes cannot carry custom code at this time, and probably never. 

But I agree the option name could be more explicit.

docs/source/customize.rst
```diff
@@ -32,6 +32,47 @@ HTML templates have their own versioning system and warnings will be triggered w
...
+- ``--privacy="PUBLIC:**"``
+  Makes everything public.
+
+- ``--privacy="HIDDEN:twisted.test.*" --privacy="PUBLIC:twisted.test.proto_helpers"``
+  Makes everything under ``twisted.test`` hidden except ``twisted.test.proto_helpers``, which will be public.
+  
+- ``--privacy="PRIVATE:**.__*__" --privacy="PUBLIC:**.__init__"``
+  Makes all dunder methods private except ``__init__``.
+
+.. note:: The order of arguments matters. Pattern added last have priority over a pattern added before,
```
tristanlatr, line 71:
```suggestion
.. important:: The order of arguments matters. Pattern added last have priority over a pattern added before,
```

docs/source/customize.rst
```diff
@@ -32,6 +32,50 @@ HTML templates have their own versioning system and warnings will be triggered w
...
+  Makes everything under ``twisted.test`` hidden except ``twisted.test.proto_helpers``, which will be public.
+  
+- ``--privacy="PRIVATE:**.__*__" --privacy="PUBLIC:**.__init__"``
+  Makes all dunder methods private except ``__init__``.
+
+.. important:: The order of arguments matters. Pattern added last have priority over a pattern added before,
+  but an exact match wins over a fnmatch.
+
+.. note:: See :py:mod:`pydoctor.qnmatch` for more informations regarding the pattern syntax.
+
```
tristanlatr, line 78:
```suggestion

.. note:: Quotes should be added around each rules to avoid actual shell expansions.
    Except if you pass arguments directly to pydoctor, like from Sphinx's ``conf.py``, in this case you must not quote patterns.

```

tristanlatr:
Merging this, thanks @adiroiban for your reviews

docs/source/customize.rst
```diff
@@ -32,6 +32,53 @@ HTML templates have their own versioning system and warnings will be triggered w
...
+.. note:: Quotation marks must be added around each rule to avoid shell expansions.
+    Unless the arguments are passed directly to pydoctor, like in Sphinx's ``conf.py``, in this case you should not quote the patterns.
```
tristanlatr, line 80:
```suggestion
    Unless the arguments are passed directly to pydoctor, like in Sphinx's ``conf.py``, in this case you should not quote the privacy rules.
```

docs/source/customize.rst
```diff
@@ -32,6 +32,53 @@ HTML templates have their own versioning system and warnings will be triggered w
...
+- ``--privacy="PRIVATE:**.__*__" --privacy="PUBLIC:**.__init__"``
+  Makes all dunder methods private except ``__init__``.
+
+.. important:: The order of arguments matters. Pattern added last have priority over a pattern added before,
+  but an exact match wins over a fnmatch.
+
+.. note:: See :py:mod:`pydoctor.qnmatch` for more informations regarding the pattern syntax.
+
+.. note:: Quotation marks must be added around each rule to avoid shell expansions.
+    Unless the arguments are passed directly to pydoctor, like in Sphinx's ``conf.py``, in this case you should not quote the privacy rules.
```
tristanlatr, line 80:
```suggestion
.. note:: Quotation marks should be added around each rule to avoid shell expansions.
    Unless the arguments are passed directly to pydoctor, like in Sphinx's ``conf.py``, in this case you must not quote the privacy rules.
```

PR #518, opened 2022-03-14
--docformat=plaintext now overrides __docformat__ module variable

### LINKED ISSUE #503: --docformat=plaintext should override __docformat__ module variable
In our docs we advise to use `--docformat=plaintext` to focus on eventual python code parsing errors, but docstrings are parsed anywat if a _`_docformat__` module variable if defined. 

The issue resides here: https://github.com/twisted/pydoctor/blob/bfafbd720604de4fd5ced3f659b311db0c47d86c/pydoctor/epydoc2stan.py#L31

tristanlatr:
Fixes #503

### DIFF
--- a/pydoctor/test/test_epydoc2stan.py
+++ b/pydoctor/test/test_epydoc2stan.py
@@ -1249,13 +1249,13 @@ def test_module_docformat(capsys: CapSys) -> None:
     """
 
     system = model.System()
-    system.options.docformat = 'plaintext'
+    system.options.docformat = 'epytext'
 
     mod = fromText('''
     """
-    Link to pydoctor: U{pydoctor <https://github.com/twisted/pydoctor>}.
+    Link to pydoctor: `pydoctor <https://github.com/twisted/pydoctor>`_.
     """
-    __docformat__ = "epytext"
+    __docformat__ = "google"
     ''', modname='test_epy', system=system)
 
     epytext_output = epydoc2stan.format_docstring(mod)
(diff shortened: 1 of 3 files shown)


pydoctor/test/test_epydoc2stan.py
```diff
@@ -1353,6 +1350,28 @@ def f(self, a: str, b: int):
 
     assert ''.join(docstring2html(B_f).splitlines()) == ''.join(docstring2html(A_f).splitlines())
 
+def test_cli_docformat_plaintext_overrides_module_docformat(capsys: CapSys) -> None:
+    """
+    Test if System.options.docformat overrides Module.docformat in the case of plaintext docformat only.
+    Issue https://github.com/twisted/pydoctor/issues/503.
```
adiroiban, line 1356:
```suggestion
    When System.options.docformat is set to C{plaintext} it
    overwrites any specific Module.docformat defined for a module.
    
    See https://github.com/twisted/pydoctor/issues/503 for the reason
    of this behavior.
```

see https://jml.io/pages/test-docstrings.html

adiroiban (approved):
Not sure what to say. I don't understand why we need #503 :)

If the plaintext overwrite is done only to help with python code parting error, this looks like a bad hack.

I think that the docs should be updated to recommend another method.

Now, if you think that this is needed, I am not -1 for merging it.

Thanks!

tristanlatr:
> If the plaintext overwrite is done only to help with python code parting error, this looks like a bad hack.

Yes. It's to be able to focus on python parsing warnings and not be overwhelmed by unknown target warnings.

What would you recommend to to instead ?

adiroiban (approved):
OK. Thanks.

I don't know what to say. Let's go  with this.

I don't have another suggestion... other than adding another explicit argument.

If someone really uses plaintext and only needs `google` format in a single module, this change will break the doc build.

Now, I don't know if that is a real use case.
I guess that we can leave it like this, and if later we get a report, we can add something like `--check-syntax` that will have a similar effect.

All good!

tristanlatr:
> If someone really uses plaintext and only needs google format in a single module, this change will break the doc build.

Yes I've though of that situation. And no it won't break, just the docstrings will be plaintext. 
But when one uses `--docformat=plaintext` I think they expect all docstrings to be plaintext, at least this is what I would expect. 
And the older behaviour can be still reached by placing `__docformat__="plaintext"` on the top-level package `__init__.py `file. 

But I really don't expect this would be used in real life.

> All good!

Thanks

PR #539, opened 2022-04-06
Always enable pytest verbose

tristanlatr:
I'm always editing the `tox.ini` to add the `-vv` flag because it seems it does not work from `tox` CLI.


adiroiban (approved):
All good!

PR #541, opened 2022-04-06
Compact module index

tristanlatr:
Replaces #479


tristanlatr:
Since you already approved these changes in the older PR @adiroiban, I'll merge it now. I've added the test I was talking about.

PR #542, opened 2022-04-06
Fix index page

### LINKED ISSUE #533: Fix index page
Looks like something is broken in the index page... I don't think we should see many "packages" one aside another.

https://lesfruitsdefendus.github.io/saskatoon-ng/

by default 2022-03-30 at 10 48 37 PM

tristanlatr:
Fix #533


adiroiban (approved):
Thanks. Nicely done. That colon in the test is precious :)

tristanlatr:
Thanks for the quick review @adiroiban.

PR #545, opened 2022-04-09
 Use tags.wbr instead of zero-width spaces in epydoc2stan.insert_break_points 

tristanlatr:
Fixes #510

PR #547, opened 2022-04-11
Fix hiding of <wbr> tags

tristanlatr:
Fis issue was introduced in #545  

Before
Capture d’écran, le 2022-04-11 à 15 55 16

After
Capture d’écran, le 2022-04-11 à 16 05 58


adiroiban (approved):
Thanks!

PR #548, opened 2022-04-12
Builder and tests refactor

tristanlatr:
Takles parts of the changes proposed in #430. More specifically:

> Separate the module building logic from the System in a new class
> Adjust fromText to use the builder interface
> Also adjust the code in driver.py

This has the benefit of removing the fromAST() function, which was used only the tests (replaced by code in addModuleString, which is more simple).

Other minor refactors:
- use list instead of dict for unprocessed_modules
- create System._remove(ob) because it's used at two places now
- Ensure that zopeinterface post processing can be run multiple times

I did not went with:

> Use the new builder interface directly instead of fromText when there are more than one module to analyze

because it felt too much for single PR.


adiroiban:
is this a draft or is ready for review?

tristanlatr:
I’ve marked it as draft because I wanted to add a few tests before mergin this. But please go ahead and review this PR. I’ll only add some tests.

pydoctor/astbuilder.py
```diff
@@ -1194,6 +1195,18 @@ def parseFile(self, path: Path) -> Optional[ast.Module]:
                 self.warning("cannot parse", str(path))
             self.ast_cache[path] = mod
             return mod
+    
+    def parseString(self, py_string:str) -> Optional[ast.Module]:
```
tristanlatr, line 1199: Since it’s only used in the test, I’d suggest we drop the caching for parseString.

pydoctor/driver.py
```diff
@@ -62,30 +62,12 @@ def get_system(options: model.Options) -> model.System:
...
-            system.addPackage(path, prependedpackage)
-        elif path.is_file():
-            system.msg('addModuleFromPath', f"adding module {path}")
-            system.addModuleFromPath(path, prependedpackage)
-        elif path.exists():
-            error(f"Source path is neither file nor directory: {path}")
-        else:
-            error(f"Source path does not exist: {path}")
-        added_paths.add(path)
+    builder = model.SystemBuilder(system)
```
tristanlatr, line 65:
```suggestion
    builder = system.systemBuilder(system)
```

pydoctor/astbuilder.py
```diff
@@ -1114,6 +1114,7 @@ def __init__(self, system: model.System):
         self.currentMod: Optional[model.Module] = None
         self._stack: List[model.Documentable] = []
         self.ast_cache: Dict[Path, Optional[ast.Module]] = {}
+        self.ast_cache_str: Dict[str, Optional[ast.Module]] = {}
```
tristanlatr, line 1117:
```suggestion
```

pydoctor/astbuilder.py
```diff
@@ -1194,6 +1195,18 @@ def parseFile(self, path: Path) -> Optional[ast.Module]:
...
+        try:
+            mod: Optional[ast.Module] = self.ast_cache_str[py_string]
+        except KeyError:
+            mod = None
+            try:
+                mod = _parse(py_string)
+            except (SyntaxError, ValueError):
+                self.warning("cannot parse string: ", py_string)
+            self.ast_cache_str[py_string] = mod
+        return mod
```
tristanlatr, line 1209:
```suggestion
        mod = None
        try:
            mod = _parse(py_string)
        except (SyntaxError, ValueError):
            self.warning("cannot parse string: ", py_string)
        return mod
```

adiroiban:
Sorry. I will not have time to review this, anytime soon.
If the tests are green, I think that that it can be merged.

Thanks for your work on pydoctor :)

tristanlatr:
I'll merge this PR then :)

PR #549, opened 2022-04-24
Release pydoctor version 22.5.0

tristanlatr:
Couple of cool stuff in this new release:

* Add Read The Docs theme, enable it with option ``--theme=readthedocs``.
* Add a sidebar. Configure it with options ``--sidebar-expand-depth`` and ``--sidebar-toc-depth``. Disable with ``--no-sidebar``. 
* Highlight the active function or attribute.
* Packages and modules are now listed together.
* Docstring summaries are now generated from docutils nodes:
  - fixes a bug in restructuredtext references in summary.
  - still display summary when the first paragraph is long instead of "No summary".
* The module index now uses a more compact presentation for modules with more than 50 submodules and no subsubmodules.
* Fix source links for code hosted on Bitbucket or SourceForge.
* The ``--html-viewsource-template`` option was added to allow for custom URL scheme when linking to the source code pages and lines.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,8 +73,8 @@ You can select a different format using the ``--docformat`` option or the ``__do
 What's New?
 ~~~~~~~~~~~
 
-in development
-^^^^^^^^^^^^^^
+pydoctor 22.5.0
+^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


adiroiban (approved):
Awesome. Thanks!

PR #551, opened 2022-04-26
Docstring Linker refactor

### LINKED ISSUE #507: Isolate docstring linker code in it's own module
https://github.com/twisted/pydoctor/pull/498 introduced a new property: `Documentable.docstringlinker`, but it also introduced a cyclic import inside the property definition. 

To avoid this, we should create a new module : `pydoctor.docstringlinker` and put the `_EpydocLinker` and the `_CachedEpydocLinker` inside it.

tristanlatr:
Fixes #507, see the commit messages for more descriptions.


pydoctor/utils.py
```diff
@@ -59,8 +63,11 @@ def resolve_path(path: str) -> Path:
     return Path(Path.cwd(), path).resolve()
 
 def parse_path(value: str, opt: str) -> Path:
-    """Parse a str path to a L{Path} object
+    """
+    Parse a str path to a L{Path} object
     using L{resolve_path()}.
+
+    Watch out, prints a message and SystemExits on error!
```
adiroiban, line 70: Maybe ... but I think that error should be rename to `panic` :)

```diff
- raise error(f"{opt}: invalid path, {ex}.")
+ error(f"{opt}: invalid path, {ex}.")
```

adiroiban (approved):
Sorry.. These days I am super busy and don't have much time for pydoctor.
But the change looks good.
Thanks!

pydoctor/utils.py
tristanlatr: Not sure to understand your diff here, there is no "raise error" in the file.
I'm OK renaming error() to panic().

adiroiban: there is at line 75 

https://github.com/twisted/pydoctor/pull/551/files#diff-ba8141603c14d5cbbbafb27976d7aa8cde126ac102147640ac1158c1f8d298c6R75

tristanlatr:
Thanks @adiroiban

PR #553, opened 2022-04-28
Fix tests and refactor tox.ini

### LINKED ISSUE #552: Tests fails because latest twisted drop support for python 3.6
We get the error: 

```
ERROR: Package 'twisted' requires a different Python: 3.6.15 not in '>=3.7.1'
```

I think we should test with the stable version of Twisted and have an additional test that uses the latest version. 

I'm not seeing any strong benefits to drop python 3.6 right now for pydoctor.

tristanlatr:
Refactor ``tox.ini`` to split commands ``test``, ``coverage`` and ``twisted-apidoc`` in 3 sections.

Do not install latest twisted package for testing pydoctor.

This fixes #552 

Place testing requirements inside `setup.cfg` to avoid duplicating data in `tox.ini`.

Test with latest Twisted version in the CI, if we're not running with python3.6.

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -46,7 +46,7 @@ jobs:
         tox -e test
 
     - name: Run unit tests with latest Twisted version
-      if: matrix.python-version != '3.6'
+      if: matrix.python-version != '3.6' && matrix.python-version != 'pypy-3.6'
       run: |
         tox -e test-latest-twisted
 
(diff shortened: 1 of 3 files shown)


adiroiban (approved):
Looks good. Thanks.

Happy to see the move of deps declaration into setup.cfg

py3.6 was removed from twisted as we are low on developers so we try to spend the available time support the more popular py versions.

PR #554, opened 2022-05-01
Support twisted.python.deprecated

### LINKED ISSUE #315: Integrate Twisted's customizations

tristanlatr:
Almost fixes issue #315 

It also includes a bit of refactor. 

by default 2022-05-01 at 11 10 00 AM

### DIFF
--- a/pydoctor/deprecate.py
+++ b/pydoctor/deprecate.py
@@ -1,0 +1,135 @@
+
+# Copyright (c) Twisted Matrix Laboratories.
+# Adjusted from file twisted/python/_pydoctor.py
+
+"""
+Support for L{twisted.python.deprecate}.
+"""
+
+import ast
+import inspect
+from numbers import Number
+from typing import Optional, Sequence, Union, TYPE_CHECKING, cast
+
+from pydoctor import astbuilder, model, zopeinterface, epydoc2stan, astutils
+
+from twisted.python.deprecate import deprecated, _getDeprecationWarningString
+from incremental import Version
+
+if TYPE_CHECKING:
+    import incremental
+
+def getDeprecated(self:model.Documentable, decorators:Sequence[ast.expr]) -> None:
+    """
+    With a list of decorators, and the object it is running on, set the
+    C{_deprecated_info} flag if any of the decorators are a Twisted deprecation
+    decorator.
+    """
+    for a in decorators:
+        if isinstance(a, ast.Call):
+            fn = astbuilder.node2fullname(a.func, self)
+
+            if fn in (
+                "twisted.python.deprecate.deprecated",
+                "twisted.python.deprecate.deprecatedProperty",
+            ):
+                try:
+                    text = deprecatedToUsefulText(self, self.name, a)
+                except Exception as e:
+                    # It's a reference or something that we can't figure out
+                    # from the AST.
+                    self.report(str(e), section='deprecation text')
+                else:
+                    # Add a warning with reStructuredText .. warning:: directive.
+                    parsed_info = epydoc2stan.parse_docstring(self,
+                        ".. warning:: {}".format(text), self, 
+                        markup='restructuredtext', 
+                        section='deprecation text')
+
+                    self.extra_info.append(parsed_info)
+
+
+class ModuleVisitor(zopeinterface.ZopeInterfaceModuleVisitor):
+    def visit_ClassDef(self, node:ast.ClassDef) -> None:
+        """
+        Called when a class definition is visited.
+        """
+        super().visit_ClassDef(node)
+        try:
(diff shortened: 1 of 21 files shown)


pydoctor/templatewriter/pages/__init__.py
```diff
@@ -196,14 +196,6 @@ def namespace(self, obj: model.Documentable) -> List[Union[Tag, str]]:
         parts.reverse()
         return parts
```
tristanlatr, line 198: This will make the Twisted system build pass with a warning instead of failing.
```suggestion
    @renderer
    def deprecated(self, request: object, tag: Tag) -> "Flattenable":
        import warnings
        warnings.warn("Renderer 'CommonPage.deprecated' is deprecated, the twisted's deprecation system is now supported by default.")
```

tristanlatr (commented):
Still todo: 
- Use the deprecated decorators in restructuredtext and epytext demo packages

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -195,15 +195,10 @@ def namespace(self, obj: model.Documentable) -> List[Union[Tag, str]]:
...
     @renderer
     def deprecated(self, request: object, tag: Tag) -> "Flattenable":
-        msg = self.ob._deprecated_info
-        if msg is None:
-            return ()
-        else:
-            return tags.div(msg, role="alert", class_="deprecationNotice alert alert-warning")
-
+        import warnings
+        warnings.warn("Renderer 'CommonPage.deprecated' is deprecated, the twisted's deprecation system is now supported by default.")
```
tristanlatr, line 201:
```suggestion
        warnings.warn("Renderer 'CommonPage.deprecated' is deprecated, the twisted's deprecation system is now supported by default.")
        return ''
```

pydoctor/test/test_astbuilder.py
```diff
@@ -2020,3 +2021,59 @@ def test_variable_named_like_current_module(systemcls: Type[model.System]) -> No
     example = True
     ''', systemcls=systemcls, modname="example")
     assert 'example' in mod.contents
+
+def test_twisted_python_deprecate(capsys: CapSys) -> None:
+
```
adiroiban, line 2026:
```suggestion
def test_twisted_python_deprecate(capsys: CapSys) -> None:
    It recognizes Twisted deprecation decorators and add the
    deprecation info as part of the docemntation.
```

adiroiban: Maybe also extend this test to include a deprecatedProperty decorator

adiroiban (approved):
Looks good. Much appreciated.

Maybe also include a deprecated property example as part of: `class DemoClass(ABC, _PrivateClass):`

pydoctor/deprecate.py
```diff
@@ -0,0 +1,154 @@
...
+                try:
+                    text = deprecatedToUsefulText(self, self.name, a)
+                except Exception as e:
+                    # It's a reference or something that we can't figure out
+                    # from the AST.
+                    self.report(str(e), section='deprecation text')
+                else:
+                    # Add a warning with reStructuredText .. warning:: directive.
+                    parsed_info = epydoc2stan.parse_docstring(self,
+                        ".. warning:: {}".format(text), self,
```
tristanlatr, line 45: Would be better to use the .. deprecated:: directive, since we support it.

PR #557, opened 2022-05-01
Turn model classes into protocols

tristanlatr:
WIP


tristanlatr:
These changes are probably not compatible with #576. Further checks must be done to see if it's worth it to merge this PR.

PR #566, opened 2022-05-08
Update codedoc.rst

Codeberg-AsGithubAlternative-buhtz:
I just improved the explanation of the fields.

### DIFF
--- a/docs/source/codedoc.rst
+++ b/docs/source/codedoc.rst
@@ -115,15 +115,15 @@ ReStructuredText fields are written with colons, like ``:field:`` or ``:field ar
 
 Here are the supported fields (written with ReStructuredText format, but same fields are supported with Epytext): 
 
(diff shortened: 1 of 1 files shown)


Codeberg-AsGithubAlternative-buhtz:
btw: I also tried to modify this line

`Here are the supported fields (written with ReStructuredText format, but same fields are supported with Epytext):`

that way that it links to the ReStructuredText and Epytext section of the documentation. But I am not aware how to create in-doc-links with GitHub markdown. ;) Beside the link feature it would highlight the two words for the reader and make it clearer that there is a difference between documentation formats.

For me it wasn't clear the first time because pydoctor/twisted is epydoc focused but you use ReStructredText examples in your documentation.

tristanlatr (approved):
Thanks, the changes looks good. 

I agree that some more efforts could be done to improve the documentation ;-)

PR #570, opened 2022-05-08
process-fields to process-types

### LINKED ISSUE #568: DOC: --process-fields does not exist
The documentation [here](https://github.com/twisted/pydoctor/blob/master/docs/source/codedoc.rst#type-fields) tells me about the commandline argument `--process-fields`. But looking into `--help` tells me it does not exist.

But there is a `--process-types`.

It does something but I am not sure if it is a synonym for `--process-fields`. I expected type names as hyperlinks. But they don't.

Codeberg-AsGithubAlternative-buhtz:
Fix #568


tristanlatr (approved):
Thanks!

PR #571, opened 2022-05-08
Update __init__.py

### LINKED ISSUE #567: Inline code in epytext_demo
https://github.com/twisted/pydoctor/blob/6e2febe63a8ed0f2e316009f34721dc1778661e7/docs/epytext_demo/__init__.py#L170

I'm not sure that is why I ask instead of creating an PR.

Codeberg-AsGithubAlternative-buhtz:
Fix #567

Please check carefully if this is correct.

### DIFF
--- a/docs/epytext_demo/__init__.py
+++ b/docs/epytext_demo/__init__.py
@@ -167,9 +167,10 @@ Escaping is used to write text that would otherwise be interpreted as epytext ma
 Escaped text has the form EE{lb}codeE{rb},
 where code is an escape code that specifies what character should be produced.
 
-If the escape code is a single character (other than '{' or '}'), then that character is produced.
-For example, to begin a paragraph with a dash (which would normally signal a list item), write 'E{-}'.
-In addition, two special escape codes are defined: 'E{lb}' produces a left curly brace ('{'); and 'E{rb}' produces a right curly brace ('}').
+If the escape code is a single character (other than `{` or `}`), then that character is produced.
+For example, to begin a paragraph with a dash (which would normally signal a list item), write `E{-}`.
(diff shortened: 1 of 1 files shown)


tristanlatr:
Hi @Codeberg-AsGithubAlternative-buhtz, the back tic quotation mark is used in restructuredtext to signify a reference.

Here we use epytext, so the format to make a inline code element is `C{code}`. So I’m afraid you got confused here and you changes will not render as you like.

Look:

ACA2BD2D-0AA0-4ADB-AF8E-5984B597039C

Codeberg-AsGithubAlternative-buhtz:
Hi @tristanlatr 
I didn't understood your first answer. I even don'T understand if there really is a problem or not. So I am out of here I would say.

I also have no way to test the doc output before creating a new or modifying this PR. I love pydoctor but I won't waste more time of my life setting up sphinx beside of it. So I don't see how to test how my modifications will look like in HTML output.

I assume also that I am confused about the original doc text.

Sorry for wasting your time with an ugly PR.

Codeberg-AsGithubAlternative-buhtz:
OK, we comin closer. I used `tox` to check the results. To speak in HTML/CSS: My goal was to make inline code like this

`<tt class="rst-docutils literal">E{code}</tt>`

And now after upgrading the docs I (assume) to better understand how escaping is done. In the previous version it wasn't clear (for me).

This is how my result (local) looks like now:
grafik

And just to make it more obvious I changed the background color:
grafik

Codeberg-AsGithubAlternative-buhtz:
I think I got it now. This is totaly wrong what I have done because I mixed Escaping with Inline-Code.

Codeberg-AsGithubAlternative-buhtz:
So no please take a look. Would be glad to merge this because I don't use epydoc style. ;)

PR #572, opened 2022-05-09
Add support for re-exporting names imported from a wildcard import

### LINKED ISSUE #565: Re-exporting not working when using wildcard import in __init__.py
This is about Re-exporting as described in the [documentation](https://pydoctor.readthedocs.io/en/latest/codedoc.html#re-exporting).

I have a function `mypackage._mypackage.foo` and a class `mypackage._mypackage.Bar`. My goal is that they appear in the documentation as `mypackage.foo` and `mypackage.Bar`. But currently they don't. Please see the attach minimal example including the generate HTML docu: [pydbug.zip](https://github.com/twisted/pydoctor/files/8645330/pydbug.zip).

### `__init__.py`
As described in the documentation I try to use the `__all__` variable.
```
from ._mypackage import *

__all__ = ['foo', 'Bar']

```
But it has no effect.

### Sidenote
- The package was installed via `pip3 install -e .`.
- I call `pydoctor` via my script `docs/hey_doc.py`. It simply extract some meta-data from the package and use it as comandline arguments.
- Using Debian 11 (current stable) with Python 3.9.2 and pydoctor 22.5.1.

### EDIT
In my example I do an `import *`.
In your documentation about Re-exporting you explicit import all elements.
So when I do `from ._mypackage import foo, Bar` it works. In that case there is also a message about it in the generator output.
```
moving 'mypackage._mypackage.foo' into 'mypackage'
moving 'mypackage._mypackage.Bar' into 'mypackage'

```
OK, now we know some more.

I see no choice to explict import all components. I wan' everything and all for all the time. I don't want to think about to import something explicit when I create a new function or class in `_mypackage.py`. I know that [PEP8](https://peps.python.org/pep-0008/#imports) tells to _avoid wildcard imports like from module import *_. But in that case it makes sense.

tristanlatr:
Fixes #565. I did not changed the implementation, only made some basic refactor and apply the same process for imported names from wildcard imports.

It does the job, but we still have no code to handle cases where some names are re-exported more than once.

### DIFF
--- a/pydoctor/test/test_astbuilder.py
+++ b/pydoctor/test/test_astbuilder.py
@@ -2015,3 +2015,35 @@ def test_package_name_clash(systemcls: Type[model.System]) -> None:
     builder.addModuleString('', 'sub2', parent_name='mod', is_package=True)
 
     assert isinstance(system.allobjects['mod.sub2'], model.Module)
+
+@systemcls_param
+def test_reexport_wildcard(systemcls: Type[model.System]) -> None:
+    """
+    """
+    system = systemcls()
+    builder = system.systemBuilder(system)
+    builder.addModuleString('''
+    from ._impl import *
+    from _impl2 import *
+    __all__ = ['f', 'g', 'h', 'i', 'j']
+    ''', modname='top', is_package=True)
+
+    builder.addModuleString('''
+    def f(): 
+        pass
+    def g():
+        pass
+    def h():
+        pass
+    ''', modname='_impl', parent_name='top')
+    
+    builder.addModuleString('''
+    class i: pass
+    class j: pass
+    ''', modname='_impl2')
+
+    builder.buildModules()
+
+    assert system.allobjects['top._impl'].resolveName('f') == system.allobjects['top'].contents['f']
+    assert system.allobjects['_impl2'].resolveName('i') == system.allobjects['top'].contents['i']
+    assert all(n in system.allobjects['top'].contents for n in  ['f', 'g', 'h', 'i', 'j'])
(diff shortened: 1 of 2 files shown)


tristanlatr:
@Codeberg-AsGithubAlternative-buhtz, could you install this version of pydoctor and try it out to see if it handle your exports correctly. FYI, there is no code to handle cases where some names are re-exported more than once, so better to stick with one re-export per name.

Codeberg-AsGithubAlternative-buhtz:
Thanks a lot. I will check that asp and give you feedback

Codeberg-AsGithubAlternative-buhtz:
I can confirm that it works. Great. Thanks a lot!

pydoctor/test/test_astbuilder.py
```diff
@@ -2015,3 +2015,35 @@ def test_package_name_clash(systemcls: Type[model.System]) -> None:
     builder.addModuleString('', 'sub2', parent_name='mod', is_package=True)
 
     assert isinstance(system.allobjects['mod.sub2'], model.Module)
+
+@systemcls_param
+def test_reexport_wildcard(systemcls: Type[model.System]) -> None:
+    """
+    """
```
adiroiban, line 2022: I tried the explain the scope and purpose of this test :)

```suggestion
def test_reexport_wildcard(systemcls: Type[model.System]) -> None:
    """
    If a target module,
    explicitly re-export via C{__all__} a set of names
    that were initially imported from a sub-module via a wildcard,
    those names are documented as part of the target module.
    """
```

adiroiban (approved):
Looks good and can be merged.

I think that it would be nice to have a similar re-export as part of the example documentation...so that we can have a hand-on experience.

but this should not block the merge of this PR.

Thanks again!

PR #573, opened 2022-05-09
Actually support and document option add-package

### LINKED ISSUE #563: SOURCEPATH (positional arguments) in project config files
I am aware that you write in the docs: _Currently, positional arguments cannot be configured with a file._

I couldn't find an Issue about that. Isn't there a future goal to improve that behavior?

I am quite new to pydoctor but from my point of view the config files are nearly useless if I also have to use the commandline with them. I'm using config files because I don't wan't to use command line anymore. I just want to type `pydocotor` in the projects root and `pydoctor` should figure out what to do by itself reading a config file.

Does `pydoctor` use the current working directory as `SOURCEPATH` by default? Maybe this is not a problem for people using a more traditional project folder layout. But I am using the `src`-layout. I have to do `pydoctor ./src/packagename` in my case.

tristanlatr:
Fixes #563.

### DIFF
--- a/docs/source/help.rst
+++ b/docs/source/help.rst
@@ -16,7 +16,7 @@ All arguments can also be set in a config file.
 
 Repeatable arguments must be defined as list.
 
-Positional arguments can be set with repeatable option ``add-package``.
+Positional arguments can be set with option ``add-package``.
 
 By convention, the config file resides on the root of your repository. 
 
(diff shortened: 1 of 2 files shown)


glyph (approved):
Looks great!

tristanlatr:
Thanks @glyph

PR #575, opened 2022-05-10
Fix restructuredtext line numbers

### LINKED ISSUE #237: Exact line numbers for references in reStructedText
Since #227, we report exact line numbers for bad cross references in epytext. However, for reStructedText I couldn't get exact line number information, since `node.line` was always `None` when `visit_title_reference()` was called for the reference node. Since there were more pressing issues to work on, I settled for reporting the starting line of the docstring instead.

I left a TODO comment in `_EpydocHTMLTranslator.visit_title_reference()` in `pydoctor/epydoc/markup/restructuredtext.py`.

tristanlatr:
Fixes #237

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,6 +75,7 @@ What's New?
 
 in development
 ^^^^^^^^^^^^^^
+* Fix line numbers in reStructuredText xref warnings.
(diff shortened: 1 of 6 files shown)


adiroiban (approved):
Looks good. Thanks!

PR #576, opened 2022-05-12
Extension system

tristanlatr:
Provide a better extension system for pydoctor. 

This new extension system breaks some compatibility with the old way of customizing pydoctor with option `--system-class` because we don't need to subclass the ASTBuilder to include custom AST visitor code. This is why the **Twisted build is crashing with this new version**. If everything works correctly, the Twisted customizations can be completely removed except the part that adds a link to the latest version of the document we're viewing if it's not the latest version. 

The extension system currently support 3 types of customizations:
- AST builder visitor extensions
- Mixin classes to be applied to `Documentable` subclasses
- Post processors to be run after all modules are built

It does not support support customizing renderers, so the code in `ZopeInterfaceClassPage` is not part of the extension system, but still built in with pydoctor. This point can probably be taken care of in a later PR.

The extension system rely on composition of multiple visitors into one. 
We do not not need to subclass the `System` class to define an extension, we just need to add a new module under `pydoctor.extensions` (to be packaged to pydoctor). Though, we currently need to define a system class if we want to add a custom extension.

It now much easier to add more extensions and properly isolate each feature in a different module. 

~~Add a new CLI argument: `--extension` to load a custom extension.~~ Since it's still possible to add extensions with a custom system like the following:
   ```python
   from pydoctor import model
   class MySystem(model.System):
       custom_extensions = ['mylib._pydoctor_extension'] # Load addional extensions
   ```
It think I wont add this new argument right now. Also because this change does not mean pydoctor have now a stable API, more changes will come down the road, so I'de suggest we keep the customization system rather "private".
- Document how to write an extension in the narrative documentation- stop linking to the Twisted customizations be cause it doesn't work this way anymore. See documentation here: https://pydoctor--576.org.readthedocs.build/en/576/customize.html#use-a-custom-system-class

PR #578, opened 2022-05-13
Type aliases

### LINKED ISSUE #472: Document type aliases
For the following code:

```python
from typing import Callable, Tuple

Parser = Callable[[str], Tuple[int, bytes, bytes]]
```

image

It would be nice if the variable value was documented for variables that are just `typing` aliases.

tristanlatr:
Fixes #472 

Marked as draft until #576 is merged.


pydoctor/astbuilder.py
```diff
@@ -1322,3 +1302,7 @@ def parseDocformat(node: ast.Assign, mod: model.Module) -> None:
     '__all__': parseAll,
     '__docformat__': parseDocformat
 }
+
+
+def setup_pydoctor_extension(r:extensions.ExtRegistrar) -> None:
+    r.register_astbuilder_visitors(TypeAliasVisitorExt)
```
tristanlatr, line 1308:
```suggestion
    r.register_astbuilder_visitor(TypeAliasVisitorExt)
```

pydoctor/astutils.py
```diff
@@ -155,3 +174,167 @@ def is_using_annotations(expr: Optional[ast.AST],
...
+        module = ctx.module
+        assert module is not None
+        module.report(f'syntax error in annotation: {ex}', lineno_offset=node.lineno)
+        return node
+    else:
+        assert isinstance(expr, ast.expr), expr
+        return expr
+
+class _AnnotationStringParser(ast.NodeTransformer):
+    """Implementation of L{ModuleVistor._unstring_annotation()}.
```
tristanlatr, line 195:
```suggestion
    """Implementation of L{unstring_annotation()}.
```

pydoctor/astutils.py
```diff
@@ -155,3 +174,167 @@ def is_using_annotations(expr: Optional[ast.AST],
...
+        "typing.OrderedDict",
+        "typing.Counter",
+        "typing.ChainMap",
+        "typing.Generator",
+        "typing.AsyncGenerator",
+        "typing.Type",
+        "typing.Pattern",
+        "typing.Match",
+        # Special forms
+        "typing.Final",
```
tristanlatr, line 287:
```suggestion
```

pydoctor/templatewriter/pages/attributechild.py
```diff
@@ -71,7 +71,7 @@ def functionBody(self, request: object, tag: Tag) -> "Flattenable":
 
     @renderer
     def constantValue(self, request: object, tag: Tag) -> "Flattenable":
-        if self.ob.kind is not DocumentableKind.CONSTANT or self.ob.value is None:
+        if self.ob.kind not in (DocumentableKind.CONSTANT, DocumentableKind.TYPING_VAR) or self.ob.value is None:
```
tristanlatr, line 74: Maybe we should have a system method determining which object we should show the value for, this would have the benefit to be hackable.

```suggestion
        if not self.ob.sytem._should_present_constant_value_of(self.ob) or self.ob.value is None:
```

pydoctor/model.py
```diff
@@ -693,6 +694,8 @@ def __init__(self, options: Optional['Options'] = None):
             self.extensions = list(extensions.get_extensions())
         assert isinstance(self.extensions, list)
         assert isinstance(self.custom_extensions, list)
+        # pydoctor.astbuilder includes some required extensions, so always add it.
```
tristanlatr, line 697: This design allows us to separate the logic for new features that only depends on the current ast builder flow to be separated in different classes.

pydoctor/test/test_astbuilder.py
```diff
@@ -1406,7 +1406,7 @@ def test_literal_string_annotation(annotation: str, expected: str) -> None:
     """Strings inside Literal annotations must not be recursively parsed."""
     stmt, = ast.parse(annotation).body
     assert isinstance(stmt, ast.Expr)
-    unstringed = astbuilder._AnnotationStringParser().visit(stmt.value)
+    unstringed = astutils._AnnotationStringParser().visit(stmt.value)
```
tristanlatr, line 1409: Using unstring_annotation() directly would be best I suppose.

tristanlatr:
Hi @not-my-profile, this addresses the issue issue you opened, could you have a look quickly to tell me what you think ? Thanks.

not-my-profile:
I just looked at the two example links you posted, those look good to me :)

tristanlatr:
Probably, for the sake of correctness, we should have two new kinds. One for type aliases and one for type variables.

tristanlatr:
> Probably, for the sake of correctness, we should have two new kinds. One for type aliases and one for type variables.

I’m not quite sure.. @glyph any opinion about that ?

glyph:
> > Probably, for the sake of correctness, we should have two new kinds. One for type aliases and one for type variables.
> 
> I’m not quite sure.. @glyph any opinion about that ?

I don't really know enough about pydoctor's internals to appreciate the significance of a "kind". If you want to give me a little context about the implications of the distinctionI'm happy to provide an opinion though

tristanlatr:
Even if undocumented, the concept of "Kinds" is a public concept of pydoctor, below you can see a screen shot with a couples of different kinds 

by default 2022-06-04 at 9 26 07 PM

You can find the enumeration here: https://pydoctor.readthedocs.io/en/latest/api/pydoctor.model.DocumentableKind.html

In the current PR, I only added one new kind: "Type", and my question was wether we should differentiate type alias and type vars with two kinds or not? 

Tell me what you think,

Thanks,

glyph:
Thanks for the explanation.

I'd say that all of Type, TypeVar, and TypeAlias are distinct kinds of things, and should be called out as such.

For example, `AnyStr` from the stdlib is a `TypeVar`, but the equivalent `Type` would have a completely different meaning.  Consider these two signatures:

```python3
def something(x: AnyStr, y: AnyStr) -> AnyStr:
def something(x: str | bytes, y: str | bytes) -> str | bytes:
```

The first means "takes either str or bytes and returns same" whereas the latter means "takes any combination of str or bytes and returns one of them based on its own criteria"

tristanlatr:
Ok, your probably right, thank. I'll introduce two new kinds: "Type Alias" and "Type Variable".

pydoctor/test/test_astbuilder.py
```diff
@@ -1971,3 +1971,39 @@ class j: pass
...
+        mylst = yourlst = list[str]
+        alist: TypeAlias = 'list[str]'
+        
+        notanalias = 'Callable[[str], Tuple[int, bytes, bytes]]'
+
+        class F:
+            from ext import what
+            L = _j = what.some = list[str]
+            def __init__(self):
+                self.P: TypeAlias = 'Callable[[str], Tuple[int, bytes, bytes]]'
```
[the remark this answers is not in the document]
glyph: I agree that this is unlikely to happen, but when it *does* happen, what is the feedback going to be like? At the very least there should be some docs about the limitations here.

But, this is not a blocker: we can do that separately. Having up-to-date docs is great, but the current behavior with type aliases is not an improvement over having it mostly work and edge-cases be undocumented.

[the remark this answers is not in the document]
glyph: Oh, good point. Well, nevermind then :)

glyph (approved):
lGTM

PR #579, opened 2022-05-13
Correctly compute class method resolution order

### LINKED ISSUE #10: methods inherited via more than one class are documented more than once

### LINKED ISSUE #528: Missing subclasses for ParsedDocstring
With version 22.3.0.dev0:

Capture d’écran, le 2022-03-22 à 15 17 51
https://pydoctor.readthedocs.io/en/latest/api/pydoctor.epydoc.markup.ParsedDocstring.html

We should see a lot more of subclasses here :/ 

Investigation needed to see when was the regression exactly (if it's a regression)

tristanlatr:
Fixes #10 and Fixes #528 and introduce a bit of refactor. 

TODO: create an issue to Display the MRO when it extend more that the default `object` type. A bit like Java doc does,

I'll create a follow-up issue to work on that.


pydoctor/model.py
```diff
@@ -501,7 +519,47 @@ def setup(self) -> None:
...
+            return [o for o in self._mro if isinstance(o, Class)]
+        else:
+            return self._mro
 
+    @property
+    def bases(self) -> List[str]:
+        return [self.parent.expandName(str_base)for str_base in self.rawbases]
+    
+    @property
+    def baseobjects(self) -> List[Optional['Class']]:
```
tristanlatr, line 557: This is not the right way to handle baseobjects attribute, we should compute it only once, in post processing. But there is probably code that relies on some parts of the name resolving which relied on this attribute. So maybe we’ll have to compte it twice once at the time of visiting the classdef and in post processing.

tristanlatr, line 557: This is not the right way to handle baseobjects attribute, we should compute it only once, in post processing. But there is probably code that relies on some parts of the name resolving which relied on this attribute. So maybe we’ll have to compte it twice once at the time of visiting the classdef and in post processing.

pydoctor/model.py
```diff
@@ -554,7 +636,7 @@ def docsources(self) -> Iterator[Documentable]:
         yield self
         if not isinstance(self.parent, Class):
             return
-        for b in self.parent.allbases(include_self=False):
+        for b in self.parent.mro():
```
tristanlatr, line 639:
```suggestion
        for b in self.parent.mro()[1:]:
```
A test is required for the inherited doc sources. My code is obviously wrong and the tests still passes :/

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -469,7 +465,7 @@ def objectExtras(self, ob: model.Documentable) -> List[Tag]:
         page_url = self.page_url
         name = ob.name
         r: List[Tag] = []
-        for b in self.ob.allbases(include_self=False):
+        for b in self.ob.mro():
```
tristanlatr, line 468:
```suggestion
        for b in self.ob.mro()[1:]:
```
same comment as above.

pydoctor/model.py
```diff
@@ -501,7 +549,51 @@ def setup(self) -> None:
...
+        if include_self is False:
+            _mro = _mro[1:]
+        return _mro
 
+    @property
+    def bases(self) -> List[str]:
+        return [self.parent.expandName(str_base) for str_base in self.rawbases]
+    
+    @property
+    def baseobjects(self) -> List[Optional['Class']]:
```
tristanlatr, line 593:
```suggestion
    def baseobjects(self) -> List[Optional['Class']]:
        """
        Base objects, L{None} value is inserted when the base class could not be found in the system.
        
        @note: This property is currently computed two times, a first time when we're visiting the ClassDef and initially creating the object. 
            It's computed another time in post-processing to try to resolve the names that could not be resolved the first place. This is needed when there are import cycles. 
            
            Meaning depending on the state of the system, this property can return either the initial objects or the final objects
        """
```

pydoctor/model.py
```diff
@@ -501,7 +549,51 @@ def setup(self) -> None:
...
+        return self._finalbaseobjects if \
+            self._finalbaseobjects is not None else self._initialbaseobjects
+    
     def allbases(self, include_self: bool = False) -> Iterator['Class']:
```
tristanlatr, line 597:
```suggestion
    def allbases(self, include_self: bool = False) -> Iterator['Class']:
       """
       Iterate on all base objects of this class and it's super classes. Doesn't comply with MRO.
       """
```

tristanlatr:
This is ready for review, @adiroiban, thanks :)

pydoctor/model.py
```diff
@@ -501,8 +549,66 @@ def setup(self) -> None:
...
+    @overload
+    def mro(self, include_external:'Literal[True]', include_self:bool=True) -> List[Union['Class', str]]:...
+    @overload
+    def mro(self, include_external:'Literal[False]'=False, include_self:bool=True) -> List['Class']:...
+    def mro(self, include_external:bool=False, include_self:bool=True) -> List[Union['Class', str]]: # type:ignore[misc]
+        """
+        Get the method resution order of this class. 
+
+        @note: The actual correct value is only set in post-processing, if L{mro()} is called
+            in the AST visitors, it will return the same as C{list(self.allbases(True))}.
```
tristanlatr, line 574:
```suggestion
            in the AST visitors, it will return the same as C{list(self.allbases(include_self))}.
```

pydoctor/model.py
```diff
@@ -501,8 +549,66 @@ def setup(self) -> None:
...
+        if include_self is False:
+            _mro = _mro[1:]
+        return _mro
+
+    @property
+    def bases(self) -> List[str]:
+        """
+        Fully qualified names of the bases of this class.
+        """
+        return [self.parent.expandName(str_base) for str_base in self.rawbases]
```
tristanlatr, line 593: This is not a cheap property to access. I believe we could compute it two times only like we do for the resolved bases. Or maybe we actually don't need this attribute at all maybe.

pydoctor/model.py
```diff
@@ -501,8 +549,66 @@ def setup(self) -> None:
...
+    
+    @property
+    def baseobjects(self) -> List[Optional['Class']]:
+        """
+        Base objects, L{None} value is inserted when the base class could not be found in the system.
+        
+        @note: This property is currently computed two times, a first time when we're visiting the ClassDef and initially creating the object. 
+            It's computed another time in post-processing to try to resolve the names that could not be resolved the first place. This is needed when there are import cycles.
```
tristanlatr, line 601:
```suggestion
            It's computed another time in post-processing to try to resolve the names that could not be resolved the first time. This is needed when there are import cycles. 
```

tristanlatr:
Should we display the MRO when it extend more that the default `object` type ? @adiroiban maybe we can find a way to make it display on the right of the screen like a side information on class pages. What do you think?

pydoctor/mro.py
```diff
@@ -0,0 +1,135 @@
...
+            if head and (head not in linearizations.tails):
+                result.append(head)
+                linearizations.remove(head)
+
+                # Once candidate is found, continue iteration
+                # from the first element of the list
+                break
+        else:
+            # Loop never broke, no linearization could possibly be found
+            raise ValueError('Cannot compute linearization')
```
tristanlatr, line 123:
```suggestion
            raise ValueError('Cannot compute linearization of the class inheritance hierarchy')
```

pydoctor/astbuilder.py
```diff
@@ -229,22 +229,19 @@ def visit_ClassDef(self, node: ast.ClassDef) -> None:
...
+            # TODO: Handle generics in MRO
+            str_base = '.'.join(node2dottedname(n) or [astor.to_source(n).strip()])
             rawbases.append(str_base)
-            full_name = parent.expandName(str_base)
-            bases.append(full_name)
-            baseobj = self.system.objForFullName(full_name)
-            if not isinstance(baseobj, model.Class):
+            expandbase = parent.expandName(str_base)
+            baseobj = self.system.objForFullName(expandbase)
+            if baseobj and not isinstance(baseobj, model.Class):
```
tristanlatr, line 241:
```suggestion
            if not isinstance(baseobj, model.Class):
```

pydoctor/astbuilder.py
```diff
@@ -229,22 +229,19 @@ def visit_ClassDef(self, node: ast.ClassDef) -> None:
...
-            full_name = parent.expandName(str_base)
-            bases.append(full_name)
-            baseobj = self.system.objForFullName(full_name)
+            expandbase = parent.expandName(str_base)
+            baseobj = self.system.objForFullName(expandbase)
             if not isinstance(baseobj, model.Class):
                 baseobj = None
-            baseobjects.append(baseobj)
+            initialbases.append(expandbase)
+            initialbaseobjects.append(cast('Optional[model.Class]', baseobj))
```
tristanlatr, line 244:
```suggestion
            initialbaseobjects.append(baseobj)
```

tristanlatr:
I think I’ll merge this PR in the coming days, I want to have this before we publish pydoctor 22.6.

tristanlatr:
The only issue that I see currently is with the C3 linearization algorithm. Here is what we can read from the [c3linear](https://github.com/pilosus/c3linear) project readme:

> I've tried to keep the code idiomatic and easy to follow at the expense of performance. The project is really took a couple of hours to complete. Although I didn't care about time complexity and didn't check correctness as thoroughly as I should, I think the project may help someone who wants to learn how Python's Method Resolution Order works.

I have absolutely no idea how to optimize this code. Instead I can try to adjust to algorithm from the standard library but it might be more complicated that excepted. 

I did not noticed any notable time increase while generating the standard library api (43mn), though #584 should help us be sure of that. 

So all in all I would say we can go ahead and merge it, but I’d like to have other opinions @glyph @adiroiban.

Thanks

glyph (approved):
This looks basically fine.  (Do we need to add Black to this project, btw? I notice the formatting looks a bit off in some places.)

tristanlatr:
Great I’ll merge this PR, thanks.

pydoctor/extensions/zopeinterface.py
```diff
@@ -182,15 +182,15 @@ def _handleZopeInterfaceAssignmentInModule(self,
...
             # TODO: Currently, this implementation will create a duplicate class 
             # with the same name as the attribute, overriding it.
             interface = self.visitor.builder.pushClass(target, lineno)
             assert isinstance(interface, ZopeInterfaceClass)
             interface.isinterface = True
             interface.implementedby_directly = []
-            interface.bases = []
-            interface.baseobjects = []
+            interface.rawbases = []
+            interface._initialbaseobjects = []
```
tristanlatr, line 193:
```suggestion
```

PR #585, opened 2022-05-20
Refactor the ASTBuilder to get rid of the currentAttr attribute.

tristanlatr:
Refactor the ASTBuilder to get rid of the currentAttr attribute.

This attribute was used to attach the right docstring node to the right Attribute object. Now it uses AST node navigation (with the .parent attribute) instead for fetching the docstring node for an ast.Assign.

This change might not be worth it, on the one hand it removes a attribute being mutated at different places in the code, but replaces this kind of "unsafe" state tracking (meaning not with pop() and push()) by some more verbose solution that involves adding the .parent attribute on all nodes.

The zopeinferface extension needed to be adjusted as well because it relied on the docstring assignment feature in an implicit way, now it's explicit what we're doing.

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -15,7 +15,10 @@ from typing import (
 import astor
 from pydoctor import epydoc2stan, model, node2stan
 from pydoctor.epydoc.markup._pyval_repr import colorize_inline_pyval
-from pydoctor.astutils import bind_args, node2dottedname, node2fullname, is__name__equals__main__, NodeVisitor
+from pydoctor.astutils import (bind_args, node2dottedname, node2fullname, is__name__equals__main__, 
+    get_assign_docstring_node, parentage_ast_tree, NodeVisitor)
+
+
 
 def parseFile(path: Path) -> ast.Module:
     """Parse the contents of a Python source file."""
@@ -24,9 +27,13 @@ def parseFile(path: Path) -> ast.Module:
     return _parse(src, filename=str(path))
 
 if sys.version_info >= (3,8):
-    _parse = partial(ast.parse, type_comments=True)
+    _ast_parse = partial(ast.parse, type_comments=True)
 else:
-    _parse = ast.parse
+    _ast_parse = ast.parse
+
+def _parse(source: Union[str, bytes], **kwargs:Any) -> ast.Module:
+    mod = _ast_parse(source, **kwargs)
+    return parentage_ast_tree(mod)
 
 
 def _maybeAttribute(cls: model.Class, name: str) -> bool:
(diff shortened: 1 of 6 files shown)


tristanlatr:
After a second though, I believe it’s worth it. 

Mainly because: 
- it’s now more explicit what happens in the zope interface extension.

- Narrowing down the ASTBuilder interface is a good thing since it’s used in customization mechanism.

- Also the parentage feature of the ast tree makes our builder more powerful.

pydoctor/astbuilder.py
```diff
@@ -752,25 +765,69 @@ def visit_Assign(self, node: ast.Assign) -> None:
...
+            self._handleAssignment(node.target, annotation, node.value, node.lineno)
+        except SkipInlineDocstring:
+            return
+        else:
+            self._handleInlineDocstrings(node, node.target)
+
+    def _reduceNameInContext(self, target:ast.expr) -> Tuple[List[str], model.Documentable]:
+        """
+        If the current context is a method, strip the C{'self.'} part of assignment names and return 
+            the right L{Class} context in which to use the new name. The new name maybe dotted.
```
tristanlatr, line 793:
```suggestion
        the right L{Class} context in which to use the new name. The new name maybe dotted.
```

tristanlatr, line 793:
```suggestion
        the right L{Class} context in which to use the new name. The new name maybe dotted.
```

pydoctor/astbuilder.py
```diff
@@ -40,6 +47,12 @@ def _maybeAttribute(cls: model.Class, name: str) -> bool:
     obj = cls.find(name)
     return obj is None or isinstance(obj, model.Attribute)
 
+class SkipInlineDocstring(Exception):
+    ...
+class InvalidName(Exception):
```
tristanlatr, line 52: This is a better name for this exception:

```suggestion
class NotAName(Exception):
```

pydoctor/astbuilder.py
```diff
@@ -752,25 +765,69 @@ def visit_Assign(self, node: ast.Assign) -> None:
...
+        parent = self.builder.current
+        if dottedname[0] == 'self':
+            dottedname = dottedname[1:]
+            maybe_parent = self._getClassFromMethodContext()
+            if not maybe_parent:
+                raise UsingSelfInUnknownContext()
+            parent = maybe_parent
+        return dottedname, parent
+
+    def _handleInlineDocstrings(self, assign:Union[ast.Assign, ast.AnnAssign], target:ast.expr) -> None:
```
tristanlatr, line 811: I should add a test case for the following usage:

```python
class C:
    def __init__(self, x):
        self.x = b = range(x)
        "doc of C.x"
```

glyph:
@tristanlatr sorry to punt on this, but can you address the conflicts and then re-request from @twisted/twisted-contributors rather than me personally?

PR #588, opened 2022-05-23
Add a test to avoid regression in the alias handling

tristanlatr:
This PR only adds a new test.

We can see that for Attributes, first analyzed assignent wins. But for the Classes, it's something else


pydoctor/test/test_astbuilder.py
```diff
@@ -2075,3 +2075,70 @@ class j: pass
     assert system.allobjects['top._impl'].resolveName('f') == system.allobjects['top'].contents['f']
     assert system.allobjects['_impl2'].resolveName('i') == system.allobjects['top'].contents['i']
     assert all(n in system.allobjects['top'].contents for n in  ['f', 'g', 'h', 'i', 'j'])
+
+@systemcls_param
+def test_module_level_attributes_and_aliases(systemcls: Type[model.System]) -> None:
+    """
+    Currently, the first analyzed assigment wins, basically. I believe further logic should be added
+    such that definitions in the orelse clause of the Try node is processed before the
```
tristanlatr, line 2083:
```suggestion
    such that definitions in the orelse clause (that we currently doesn't visit at all) of the Try node is processed before the 
```

tristanlatr:
Closing this in favour of #589 which includes this test.

adiroiban (approved):
All good. Thanks!

Issue #33, opened 2015-01-12
two classes with the same name confuse pydoctor


mwhudson:
If two classes are defined in a file with the same name (e.g. one defined in an 'except' block, and the other in an 'else' block), pydoctor appends a number to the end of the generated name which links to nothing.

Generated docs:
http://twistedmatrix.com/documents/10.2.0/api/twisted.web.client.html

Source:
http://twistedmatrix.com/trac/browser/tags/releases/twisted-10.2.0/twisted/web/client.py#L567

mwhudson:
(by mwhudson)
Sigh, yeah, that's not very good.  I'm surprised by the 404 tbh, but it's still not clear what one should do here.  Do you have any ideas? :-)

mwhudson:
(by lifeless)
write both classes to the html?

tristanlatr:
I think I'm having the same issue with a Module name and a Class name here: https://tristanlatr.github.io/python-Wappalyzer/Wappalyzer.html
Or maybe it's related to #17

tristanlatr:
Now, I'm actually running into this bug with class 'Arena' from https://github.com/python/cpython/blob/master/Lib/multiprocessing/heap.py

I was trying to convert cpython API output to Dash compatible file with doc2dash and i ran into: 

```
FileNotFoundError: [Errno 2] No such file or directory: '.../cpython-output.docset/Contents/Resources/Documents/Lib.multiprocessing.heap.Arena%200.html'
```

Now it looks like both classes do not appear in the module documentation. BUT rather they both appear in the name index. 
Which is the issue for `doc2dash` cause it's parsing the nameIndex.html file to extract informations. 

<img width="430" alt="by default 2020-11-28 at 7 32 51 PM" src="https://user-images.githubusercontent.com/19967168/100528957-888ed800-31b0-11eb-9552-3ec9ce47f387.png">

This is a really tricky issue, but I think both class documentation should end up in the same class page.

tristanlatr:
Here is another instance of this issue,

https://tristanlatr.github.io/python-igraph/igraph.summary%200.html

It's a little bit more sneaky since the is no other class named `summary` but there is a method in the `__init__.py` named `summary`

https://github.com/tristanlatr/python-igraph/blob/3b2313ea6918d3c641f1b7d6bd23572a485d56d2/src/igraph/__init__.py#L4788

tristanlatr:
(This issue is most probably a duplicate of #17)

Issue #17, opened 2015-01-12
pydoctor's handling of duplicate names needs a rethink


mwhudson:
Originally reported as Twisted issue 2621:

> There are several cases, some of which get quite tricky. This is not an exhaustive list, but some examples are:

def foo(): pass
def foo(): pass

class B: pass
class B(B): pass

if quux:
    def foo(): pass
else:
    def foo(): pass

tristanlatr:
Here is my proposition for tackle this issue:

For `System.allobjects` and for `Documentable.contents`, we adopt a version of `dict` that maintains a stack of duplicates objects, that we can request with special methods. Like this one: https://github.com/tristanlatr/pydocspec/blob/main/pydocspec/dupsafedict.py

- Duplicates classes/modules/package: they could be shown in the same page. It could be a bit weird, but it's better than the current behaviour. 
- For cases where a `Function` or `Attribute` overrides a `Module` or a `Class`, I don't think it would be an issue to show the documentation properly and not having to rename anything, actually.

The HTML writer should be adapted such that it can show many classes/modules on the same page and iterates over all objects, including duplicates.

PR #589, opened 2022-05-24
Visit try orelse finalbody and if orelse

tristanlatr:
Since the new NodeVisitor class, all the node are not visited as before. The older visitor relied on generic_visit() to be recursive. Where the provided generic_visit() is not recursive anymore, and moreover not called automatically when visiting unknown nodes! So what I'm saying here is that since https://github.com/twisted/pydoctor/pull/576 have been merged, we're not visiting the statements inside the 'orelse' field of Try and If nodes, same goes for 'finalbody' and 'handlers'.

This commit fixes that issue. The rationale is now the following: All statements in the 'orelse' block of IF nodes and statements in the except handlers of TRY nodes that would override a name already defined in the main 'body' (or TRY 'orelse' or 'finalbody') are ignored.

Meaning that in the context of the code below, 'ssl' would resolve to 'twisted.internet.ssl':

```python
try:
    from twisted.internet import ssl as _ssl
except ImportError:
    ssl = None
else:
    ssl = _ssl
```

### DIFF
--- a/pydoctor/astbuilder.py
+++ b/pydoctor/astbuilder.py
@@ -1,6 +1,7 @@
 """Convert ASTs into L{pydoctor.model.Documentable} instances."""
 
 import ast
+import contextlib
 import sys
 from attr import attrs, attrib
 from functools import partial
@@ -201,7 +202,39 @@ class ModuleVistor(NodeVisitor):
         self.builder = builder
         self.system = builder.system
         self.module = module
+        self._override_guard_state: Tuple[bool, model.Documentable, Sequence[str]] = \
+                                    (False, cast(model.Documentable, None), [])
+    
+    @contextlib.contextmanager
+    def override_guard(self) -> Iterator[None]:
+        """
+        Returns a context manager that will make the builder ignore any extraneous 
+        assigments to existing names within the same context.  
+        
+        @note: The list of existing names is genrated at the moment of
+            calling the function.
+        """
+        current = self.builder.current
+        ignore_override_init = self._override_guard_state
+        # we list names only once to ignore new names added inside the block,
+        # they should be overriden as usual.
+        self._override_guard_state = (True, current, self._list_names(current))
+        yield
+        self._override_guard_state = ignore_override_init
+    
+    def _list_names(self, ob: model.Documentable) -> List[str]:
+        """
+        List the names currently defined in a class/module.
+        """
+        names = list(ob.contents)
+        if isinstance(ob, model.CanContainImportsDocumentable):
+            names.extend(ob._localNameToFullName_map)
+        return names
 
+    def _name_in_override_guard(self, ob: model.Documentable, name:str) -> bool:
+        return self._override_guard_state[0] is True \
+            and self._override_guard_state[1] is ob \
+                and name in self._override_guard_state[2]
 
     def visit_If(self, node: ast.If) -> None:
         if isinstance(node.test, ast.Compare):
@@ -210,6 +243,28 @@ class ModuleVistor(NodeVisitor):
                 # whatever is declared in them cannot be imported
                 # and thus is not part of the API
                 raise self.SkipNode()
+    
+    def depart_If(self, node: ast.If) -> None:
+        # At this point the body of the Try node has already been visited
+        # Visit the 'orelse' block of the If node, with override guard
+        with self.override_guard():
+            for n in node.orelse:
+                self.walkabout(n)
+    
+    def depart_Try(self, node: ast.Try) -> None:
+        # At this point the body of the Try node has already been visited
+        # Visit the 'orelse' and 'finalbody' blocks of the Try node.
+        
+        for n in node.orelse:
+            self.walkabout(n)
+        for n in node.finalbody:
+            self.walkabout(n)
+        
+        # Visit the handlers with override guard
+        with self.override_guard():
+            for h in node.handlers:
+                for n in h.body:
+                    self.walkabout(n)
 
     def visit_Module(self, node: ast.Module) -> None:
         assert self.module.docstring is None
@@ -226,6 +281,9 @@ class ModuleVistor(NodeVisitor):
         # Ignore classes within functions.
         parent = self.builder.current
         if isinstance(parent, model.Function):
+            raise self.SkipNode()
+        # Ignore in override guard
+        if self._name_in_override_guard(parent, node.name):
             raise self.SkipNode()
 
         rawbases = []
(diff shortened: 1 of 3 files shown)


tristanlatr:
Tell me if you think the rationale is good @glyph @adiroiban. Also, this needs to be documented in the codedoc.rst file.

I think that one day we could introduce options to choose how to evaluate a specific name in a module, for instance `TYPE_CHECKING`, in some complex module, it could more easy fo pydoctor to visit the 'orelse' field only if the `If.test` can be evaluated to something we passed in as an option.

tristanlatr:
This PR is somewhat liked to #17 and #33 in the sense that a duplicate name defined in except handlers or a If else block will  not override the name previously defined.

So in the case of class 'Arena' from https://github.com/python/cpython/blob/d2ef66a10be1250b13c32fbf3c0f9a9d2d98b124/Lib/multiprocessing/heap.py#L31
Only the first class will be documented.

docs/source/codedoc.rst
```diff
@@ -318,3 +318,45 @@ The content of ``my_project/__init__.py`` includes::
     from .core._impl import MyClass
 
     __all__ = ("MyClass",)
+
+Branch priorities
+-----------------
+
+When pydoctor deals with try/except/else or if/else block, it makes sure that the names defined in 
+the "principal" branch do not get overriden by names defined in the except hanlders or ifs' else block.
```
glyph, line 326: This strikes me as a bit awkward.  I think you want to be saying something about the priority of the binding of the name in the *last* branch? Or the `else` branch?

pydoctor/astbuilder.py
```diff
@@ -194,14 +195,58 @@ def extract_final_subscript(annotation: ast.Subscript) -> ast.expr:
...
         self.builder = builder
         self.system = builder.system
         self.module = module
+        self._override_guard_state: Tuple[bool, model.Documentable, Sequence[str]] = \
+                                    (False, cast(model.Documentable, None), [])
+    
+    @contextlib.contextmanager
+    def override_guard(self) -> Iterator[None]:
+        """
+        Returns a context manager that will make the builder ignore any extraneous
```
glyph, line 222: A clearer explanation of the definition of "extraneous" would be interesting here.

glyph (approved):
LGTM with the caveat that the specific semantics of the priorities should be better explained in terms of specific order, rather than things not-getting-overridden which is a sort of a double negative.

pydoctor/astbuilder.py
tristanlatr: extraneous meaning assignments to name that already exist

docs/source/codedoc.rst
```diff
@@ -318,3 +318,45 @@ The content of ``my_project/__init__.py`` includes::
...
+Meaning that in the context of the code below, ``ssl`` would resolve to ``twisted.internet.ssl``:
+
+.. code:: python
+
+    try:
+        from twisted.internet import ssl as _ssl
+    except ImportError:
+        ssl = None
+    else:
+        ssl = _ssl
```
tristanlatr, line 337: Here the else branch have the priority over what defined in the except handler. Because it's the "principal" flow.

docs/source/codedoc.rst
tristanlatr: No it's not about whether the branch if last or not. It's only the  except hanlders and ifs' else blocks where we enable the override name guard.

tristanlatr: the `else` branch of `try:` blocks have the priority over the `except:` block. Similarly, the main body of `if:` blocks have priority over what's defined in the `else:` block. 

So the thing that might be confusing here if that `Try.orelse` and `If.orelse` are NOT handled with the same override priority. Because the main branches in case of a `if:` is the `body` uniquely and in case of a `try:` it's the `body`, the  `orelse` and the `finalbody`, in this order. So just saying `else:` block can be confusing.

pydoctor/astbuilder.py
```diff
@@ -194,14 +195,58 @@ def extract_final_subscript(annotation: ast.Subscript) -> ast.expr:
...
+        names = list(ob.contents)
+        if isinstance(ob, model.CanContainImportsDocumentable):
+            names.extend(ob._localNameToFullName_map)
+        return names
 
+    def _name_in_override_guard(self, ob: model.Documentable, name:str) -> bool:
+        """Should this name be ignored because it matches the override guard in the context of C{ob}?"""
+        return self._override_guard_state[0] is True \
+            and self._override_guard_state[1] is ob \
+                and name in self._override_guard_state[2]
```
tristanlatr, line 249: @glyph please see this method for the implementation of the override guard.

The override guard is a re-entrant context manager that ensures that follow-up assignments to names that already existed before in the same frame context are simply ignored.

docs/source/codedoc.rst
```diff
@@ -318,3 +318,45 @@ The content of ``my_project/__init__.py`` includes::
     from .core._impl import MyClass
 
     __all__ = ("MyClass",)
+
+Branch priorities
+-----------------
+
+When pydoctor deals with try/except/else or if/else block, it makes sure that the names defined in 
+the "principal" branch do not get overriden by names defined in the except hanlders or ifs' else block. 
+
```
tristanlatr, line 327:
```suggestion

.. note:: Technical note
   The :py:meth:`pydoctor.astbuilder. ModuleVistor.override_guard` is a re-entrant context manager that ensures that assignments to names that already existed before in the same context are simply ignored. 
```

pydoctor/astbuilder.py
tristanlatr, line 222:
```suggestion
        Returns a context manager that will make the builder ignore any new 
```

tristanlatr: We can read: "Returns a context manager that will make the builder ignore any new assignments"

tristanlatr:
I see this in the code:

```python
   def handleDuplicate(self, obj: Documentable) -> None:
        """
        This is called when we see two objects with the same
        .fullName(), for example::

            class C:
                if something:
                    def meth(self):
                        implementation 1
                else:
                    def meth(self):
                        implementation 2

        The default is that the second definition "wins".
        """
```

But clearly, that was never tested, so never officially supported. 
I'm wondering if we should stick to that or to what I'm proposing, which seems better IMO, but would be a breaking change. (Anyway it's broken right now and the CI is still green ^^')

glyph (approved):
I do not want to block on my quibbles about documentation. If a brief reflection does not yield a satisfactorily improved phrasing, let's deal with that later, since these semantics make more sense.

> I'm wondering if we should stick to that or to what I'm proposing, which seems better IMO, but would be a breaking change

If you wouldn't mind doing a few spot-checks to make sure that Twisted's generated docs don't have any major reliance on this pattern, I'd definitely just merge this; my suspicion would be that it will improve the docs in a bunch of places where we didn't realize we had broken references and not break anything elsewhere.

PR #590, opened 2022-05-25
Choose type checking value for a module

### LINKED ISSUE #174: Handling of `if TYPE_CHECKING:` blocks
It is sometimes necessary to have code in a module (or more rarely, in a class) that is only there for the benefit of a static code analyzer and is skipped at runtime. In particular, there is the constant `typing.TYPE_CHECKING` that type checkers can evaluate to `True`, while its runtime value is always `False`. This is supported by mypy and possibly other analyzers.

A common purpose of `if TYPE_CHECKING:` blocks is to contain imports that are only necessary for the purpose of type annotations. These might be circular imports at runtime. When these types are used to annotate arguments or return values, they are relevant to pydoctor as well: either to extract type information from annotations or because of `L{name}` tags inside the docstring, where pydoctor has to be able to look up the fully qualified version of `name`.

In other cases, `if TYPE_CHECKING:` blocks are used to perform trickery to make mypy accept code that is difficult to analyze. In these cases, pydoctor can be better off analysing the runtime version of the code instead. For details, read [the comments in this Klein PR](https://github.com/twisted/klein/pull/315).

A related issue is that pydoctor doesn't handle duplicate definitions particularly elegantly, see #33. For example, if both the `if` and `else` branch define the same name, one of them will have a name with `␣0` (space + zero) appended and that patched name not only shows up visibly in the generated documentation, it also breaks hyperlinks to the corresponding pages, since it not only escapes the link itself (for example, `klein.resource%200.html`), but it also escapes the file name: it generates a file named `klein.resource%200.html` instead of `klein.resource 0.html`.

On the one hand, it would be nice to do the right thing in as many cases as possible. On the other hand, perfect code analysis is impossible and trying too hard to handle complex code will consume a lot of effort and complicate pydoctor's code. So we have to find a balance somewhere that makes pydoctor perform in a simple and predictable way that we can communicate to the end user and that they can adapt their code to.

tristanlatr:
Fixes #174. 

Marked as draft until #589 is merged.

### DIFF
--- a/docs/source/codedoc.rst
+++ b/docs/source/codedoc.rst
@@ -319,6 +319,8 @@ The content of ``my_project/__init__.py`` includes::
 
     __all__ = ("MyClass",)
 
+.. _branch-priorities:
+
 Branch priorities
 -----------------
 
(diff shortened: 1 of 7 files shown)


tristanlatr:
Should this feature be a proper option or it should stay a system class variable to override ? It's likely that other projects will need the same kind of  adjustments as Klein, so maybe it should.

Issue #473, opened 2022-02-03
Display types in function signatures


not-my-profile:
For functions Pydoctor currently displays:

1. the function signature without types
2. the docstring
3. the parameters and their types / documentation
4. the return type and its documentation

I would prefer to see the full function signature with all types first because:

* I want to know the signature before I read about a function
* the docstring & parameter lists can become quite long, so I might have to look far down to see the return type (which arguably is the most important detail of a function)

To illustrate this with an example:

![image](https://user-images.githubusercontent.com/73739153/152324968-01e47693-e230-4e38-88fc-86e26670617e.png)

This would obviously require a rework of the functionDef as has already been suggested in #422. Though I don't think `<wbr>` are the way to go. If the signature is too long it should just be properly formatted from the get go.

tristanlatr:
See https://github.com/twisted/pydoctor/pull/417#issuecomment-888516326 about type annotations in signatures.

I think it’s a good idea to show them I the signature.

tristanlatr:
Let's talk about rendering the type annotation inside signatures, @cretz, here is what's my old code (#417) currently generates for a random function (not overloaded):

<img width="711" alt="by default 2022-05-26 at 7 42 44 PM" src="https://user-images.githubusercontent.com/19967168/170600273-dc38679e-7fb5-4ee1-9e86-0bd389cba48f.png">

As talked about with @glyph, I believe we should only see the following instead:

<img width="565" alt="by default 2022-05-26 at 7 46 59 PM" src="https://user-images.githubusercontent.com/19967168/170601592-807b3728-6caa-473e-8195-da30ac4f8744.png">
 
Here are notes on this issue:

Change the `FieldHandler` to show nothing when none of the parameters (annotated or not) are documented (stop listing the parameters or return value as "undocumented" when they only have AST annotation, this makes the behaviour more consistent because we don't list undocumented param that don't have AST annotation). In case there is only one parameter documented, then still show the whole table of parameters with "undocumented" when parameters miss docs, like we do for functions without annotations.

PR #595, opened 2022-06-02
Function overload support and param/return-sections-undocumented change

### LINKED ISSUE #406: Handle overloaded functions

cretz:
Handle `@overload` decorator and situation where all params/returns are undocumented. Specifically:

* Show signatures for `@overload` functions and add `@overload` decorator above them
* Do not show main function signature if overloads are present
* If all params are undocumented, do not show param table
* If return is undocumented, do not show return table
* Builds off of #417 (only minor changes). Fixes #406.

Example, for reST-format of:

```python
@overload
def demo_overload(s: str) -> str:
    ...

@overload
def demo_overload(s: bytes) -> bytes:
    ...

def demo_overload(s: Union[str, bytes]) -> Union[str, bytes]:
    """
    Overload signatures appear without the main signature and with ``@overload`` decorator.

    :param s: Some string or bytes param.
    :return: Some string or bytes result.
    """
```

Result:

image

Note how `(source)` is below the signatures instead of to the right as it would be w/ just one signature. Seems to be an acceptable tradeoff for now.

For undocumented reST-format of:

```python
def demo_undocumented(s: str) -> str:
    raise NotImplementedError
```

Result:

image

Notice the lack of parameter and return sections.

Here's an example of a really large overload from [some source in my company's repo](https://github.com/temporalio/sdk-python/blob/4657453dd9aaa67c92603439af92987c8df6dc11/temporalio/client.py#L200-L345):

image

Yes it can appear kind of like a large mess, but there aren't really good alternatives and it's my company's fault for using so many args. I don't think we'd want to wrap param signatures nor do I think we'd want to do a diff and only have the overloads show what's unique to them.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,6 +78,7 @@ in development
 * Add support for the ``@overload`` decorator.
 * Show type annotations in function's signatures.
 * If none of a function's parameters have documentation, do not render the parameter table.
+* Themes have been adjusted to render annotations more concisely.
 
 pydoctor 22.9.1
 ^^^^^^^^^^^^^^^


tristanlatr:
Hello @cretz,

I know the PR is marked as draft so you'll probably work on it again later, but I want to give a little heads-up about little stuff I saw that might be worth mentioning before going to far in the hacking: 

Since the object `FunctionOverload` is not an actual documentable type, I wouldn't add it to the factory mechanism, since we could tweak the function overloads by applying a mixin to the `Function` class. 

Also did you read this ticket about types in signatures ? https://github.com/twisted/pydoctor/issues/473

Thanks for you work on this already :) 

Talk to you later,

cretz:
@tristanlatr - Thanks! Yeah, I have it draft because I was (properly) afraid the CI checks here might not pass. I am fixing and then will open full review.

> Since the object FunctionOverload is not an actual documentable type, I wouldn't add it to the factory mechanism, since we could tweak the function overloads by applying a mixin to the Function class.

Ok, I can remove the factory part for that

> Also did you read this ticket about types in signatures ? https://github.com/twisted/pydoctor/issues/473

I am seeing it now. Seems like removing removing the entire parameters/return section if _all_ are "Undocumented", correct? But otherwise leave as is? Usually I would say this should be a separate PR, but I have also been really wanting this (I have big shortcut methods that are documented as just "see whatever"). I am adding now.

pydoctor/astbuilder.py
```diff
@@ -882,15 +899,20 @@ def add_arg(name: str, kind: Any, default: Optional[ast.expr]) -> None:
...
             func.report(f'{func.fullName()} has invalid parameters: {ex}')
             signature = Signature()
 
         func.signature = signature
-        func.annotations = self._annotations_from_function(node)
-    
+        func.annotations = annotations
+
+        if is_overload_func:
+            func.overloads.append(self.system.FunctionOverload(signature=func.signature))
```
tristanlatr, line 914: I think this implementation is OK, but could be more robust. 

The flow is just a bit weird because we're starting by creating the overloads (which makes sens because they are visited first) and incrementally append them to the `overloads` list of a `Function` object that is already created at the time we hit the last `FunctionDef` that contains the logic code of the function.

But what happens if a `@overload` decorator is encountered after the main declaration? I suppose the last signature overrides what what stored in `Fucntion.signature`. So maybe `Fucntion.signature` should be set only if the function is not an overloaded function? 

My understanding is that All `@overload` definitions must come before the implementation, and multiple implementations are not allowed. So when hitting an overloads function and the `Function.signature` is not `None` is means the `@overload` appeared after the main function def and we should warn about it, saying it's ignored.

Also what happens if a `@overload` function has a docstring? I'de suggest we leave a warning warning saying that docstrings on overloads function are ignored.

Tell me what you think

pydoctor/astbuilder.py
```diff
@@ -807,24 +807,29 @@ def _handleFunctionDef(self,
...
+                        is_property = True
+                    elif deco_name == ['classmethod']:
+                        is_classmethod = True
+                    elif deco_name == ['staticmethod']:
+                        is_staticmethod = True
+                    elif len(deco_name) >= 2 and deco_name[-1] in ('setter', 'deleter'):
+                        # Rename the setter/deleter, so it doesn't replace
+                        # the property object.
+                        func_name = '.'.join(deco_name[-2:])
+                if deco_name[-1] == 'overload':
```
tristanlatr, line 830:
```suggestion
                full_deco_name = parent.expandName('.'.join(deco_name))
                if full_deco_name == 'typing.overload':
```

This will check whether we use `typing.overload` and not just `something.overload`.

pydoctor/epydoc2stan.py
```diff
@@ -474,9 +475,11 @@ def resolve_types(self) -> None:
     def format(self) -> Tag:
         r: List[Tag] = []
 
-        r += format_desc_list('Parameters', self.parameter_descs)
-        if self.return_desc:
-            r += format_desc_list('Returns', [self.return_desc])
+        # Only include parameters and returns if any are documented
+        if any(p.body for p in self.parameter_descs) or (self.return_desc and self.return_desc.body):
```
tristanlatr, line 479: If I read correctly, if only the returns value is documented, all parameters will still display. Maybe 'Parameters' and 'Returns' should be independently shown ? Or maybe it makes send to keep it like this?

pydoctor/model.py
```diff
@@ -712,6 +722,9 @@ def Package(self) -> Type['Package']:
     @property
     def Attribute(self) -> Type['Attribute']:
         return self._factory.Attribute
+    @property
+    def FunctionOverload(self) -> Type['FunctionOverload']:
+        return FunctionOverload
```
tristanlatr, line 727: We can just use plain `FunctionOverload` and remove the property I believe.

pydoctor/model.py
```diff
@@ -567,11 +568,20 @@ class Function(Inheritable):
...
         if isinstance(self.parent, Class):
             self.kind = DocumentableKind.METHOD
+        self.overloads = []
+
+@attr.s(auto_attribs=True)
+class FunctionOverload:
+    """
+    @note: This is not an actual documentable type. 
+    """
+    signature: Signature
```
tristanlatr, line 584:
```suggestion
    signature: Signature
    decorators: Sequence[ast.expr]
```

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -64,12 +64,35 @@ def format_decorators(obj: Union[model.Function, model.Attribute]) -> Iterator["
...
     """
-    return html2stan(str(function.signature)) if function.signature else "(...)"
+    return html2stan(str(func.signature)) if func.signature else "(...)"
+
+def format_overloads(func: model.Function) -> Iterator["Flattenable"]:
+    """
+    Format a function overloads definitions as nice HTML signatures.
+    """
+    for overload in func.overloads:
+        yield '@overload', tags.br(), tags.div(format_function_def(func.name, func.is_async, overload))
```
tristanlatr, line 79: We are loosing the initial decorator formatting here. It would be good to use `format_decorators()` with `FunctionOverload` objects too such that we can write the following:

```suggestion
        yield from format_decorators(overload)
        yield tags.div(format_function_def(func.name, func.is_async, overload))
```

This will also include links to the `overload` docs if one has configured the standard library intersphinx URL.

pydoctor/templatewriter/pages/functionchild.py
```diff
@@ -41,21 +41,20 @@ def functionAnchor(self, request: object, tag: Tag) -> "Flattenable":
...
-        if name.endswith('.setter') or name.endswith('.deleter'):
-            name = name[:name.rindex('.')]
-        return [
-            tags.span(def_stmt, class_='py-keyword'), ' ',
-            tags.span(name, class_='py-defname'), 
-            format_signature(self.ob), ':'
-            ]
+        # Exclude function definition if there are overloads
+        if self.ob.overloads:
+            return ()
```
tristanlatr, line 56: Isn't that line supposed to be tested?

pydoctor/test/test_epydoc2stan.py
```diff
@@ -164,7 +164,7 @@ def nop() -> None:
 
 def test_func_undocumented_return_something() -> None:
     """When the returned value is undocumented (no 'return' field) and its type
-    annotation is not None, include the "Returns" entry in the output.
+    annotation is not None, do not include the "Returns" entry in the output.
```
tristanlatr, line 167:
```suggestion
    annotation is not None, do not include the "Returns" entry in the field table. It will be shown in the signature.
```

pydoctor/test/test_epydoc2stan.py
```diff
@@ -239,8 +289,8 @@ def f(a, b):
     classic_fmt = docstring2html(classic_mod.contents['f'])
     assert annotation_fmt == classic_fmt
 
-@pytest.mark.xfail
```
tristanlatr, line 242: I didn't know we had that!

tristanlatr (changes requested):
It's going to be a great improvement no to display undocumented EVERYWHERE. 

I made a few comments that I believe are worth changing bits of the code. I know I wrote this code at first but it was about a year ago, and we now have a better colorizer for AST values that is worth using. 

Also, I can write a couple of test cases for the edge cases if you want.

Thanks for your work,

Talk to you later,

pydoctor/astbuilder.py
cretz: > But what happens if a @overload decorator is encountered after the main declaration?

This is the same thought I had when I saw how we were pulling from `parent.contents`

> So maybe Function.signature should be set only if the function is not an overloaded function? [...] warn about it, saying it's ignored.

Works for me

> Also what happens if a @overload function has a docstring? I'de suggest we leave a warning warning saying that docstrings on overloads function are ignored.

This also works for me and is the path of least resistance.

I will write tests for these cases too (I have seen other tests that capture warning output)

pydoctor/astbuilder.py
cretz: I wasn't aware of the performance cost of qualifying this and I noticed we don't qualify on any other decorator name changes so I just assumed we were afraid of it. I will change to qualify the name before checking.

pydoctor/epydoc2stan.py
cretz: This was how I understood the comment at https://github.com/twisted/pydoctor/issues/473#issuecomment-1139160974. I can make them independent of each other though, no problem.

pydoctor/model.py
cretz: :+1: I didn't know whether you liked this abstraction (the previous PR references it on this class, but it was impl'd differently at that time). I will just instantiate the overload directly in the AST builder.

tristanlatr:
Also, could you include a few entries in the change log to summarize what's been changed in this PR? Thanks.

pydoctor/templatewriter/pages/__init__.py
cretz: The only problem here I see is that `format_decorators` expects its param to have the ability to `report()` warnings, so I will probably add a `report()` method on it. So I either have to add `report()` to `FunctionOverload` or do some other approach to reporting issues.

pydoctor/templatewriter/pages/functionchild.py
cretz: I will add an assertion or test that confirms that the main signature is excluded from output

cretz:
Will make updates, probably some time tomorrow

pydoctor/astbuilder.py
tristanlatr: We are not qualifying other decorators because they are either supposed to be part of the builtins (in which case the expanded name if the same) or we use a naming convention (for properties). 

> performance cost of qualifying this

Probably this method could be optimized, but it only looks in the objects in the system, it's not like it will parse additional modules to resolve a name. So it's ok to use it, the only thing to keep in mind is that at the time of AST building all objects are not in the system yet, so resolveName is not 100% reliable. But to resolve full import names it's definitely enough.

pydoctor/epydoc2stan.py
tristanlatr: Yeah, I'm not sure what's the current behaviour for such cases without annotations actually.

pydoctor/templatewriter/pages/__init__.py
tristanlatr: Ah yeah… and report expect the system… looks like you’ll have to pass the system along… though we should not duplicate the reporting logic. Not easy

pydoctor/model.py
tristanlatr: Yea this factory change is very recent indeed, and all objects on the factory are sharing the same signature. I think it’s a good thing to keep it like that. This is why I don’t think FunctionOverload should be part of it.

pydoctor/templatewriter/pages/__init__.py
tristanlatr: Or maybe the `FunctionOverload` should keep a reference to the `Function` object and act like a proxy for the report() method.

pydoctor/astbuilder.py
cretz: Not setting signature unless confirmed non-overload. Updated to warn on overload after primary and on overload with docstring.

pydoctor/astbuilder.py
cretz: Fixed

pydoctor/epydoc2stan.py
cretz: Made these checks/sections independent.

pydoctor/model.py
cretz: Removed and instantiating `FunctionOverload` directly.

pydoctor/model.py
cretz: Now passing in decorators

pydoctor/templatewriter/pages/functionchild.py
cretz: Moved to the format function and tested

pydoctor/test/test_epydoc2stan.py
cretz: Updated

cretz:
Updates made, can re-review (assuming CI checks pass, if not, will force-push whatever tiny changes needed)

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -45,10 +45,13 @@ def map_kind(kind: model.DocumentableKind) -> model.DocumentableKind:
 
     return (-o.privacyClass.value, -map_kind(o.kind).value if o.kind else 0, o.fullName().lower())
 
-def format_decorators(obj: Union[model.Function, model.Attribute]) -> Iterator["Flattenable"]:
+def format_decorators(obj: Union[model.Function, model.Attribute, model.FunctionOverload]) -> Iterator["Flattenable"]:
+    # Use the overload's primary function for some parts here
+    documentable_obj = obj if not isinstance(obj, model.FunctionOverload) else obj.primary
```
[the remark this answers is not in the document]
tristanlatr: I believe you need to use the OverloadFunction object. So this means adding a report method the OverloadFunction class that would pass along to the underlying function. Right now we don’t colorize the overload decorator if I read correctly.

Tell me what you think

tristanlatr (commented):
I think there is a little bug

pydoctor/templatewriter/pages/__init__.py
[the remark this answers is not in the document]
tristanlatr: Never mind, it looks like it working. But we might use a bit if comments to understand what happens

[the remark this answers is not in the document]
cretz: > I don’t understand what happens here, if it’s an overloaded function, you documents the primary function decorators? So it won’t show the @overload decorator ?

No, I still use `obj` for that stuff

> I believe you need to use the OverloadFunction object. So this means adding a report method the OverloadFunction class that would pass along to the underlying function.

There are three parts here that won't work with `FunctionOverload`: `node2fullname` for use when qualifying a decorator, getting a `docstring_linker` for `doc.to_stan`, and `report` for reporting issues.

I originally did just make a `report` method, but I have to solve the other three too. Technically I can "proxy" that stuff too, but it became clearer to code it this way. I can change if necessary.

> Right now we don’t colorize the overload decorator if I read correctly.

I think it is colorizing correctly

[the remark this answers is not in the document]
cretz: I will add comment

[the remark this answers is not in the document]
tristanlatr: You are right, thanks

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -45,10 +45,13 @@ def map_kind(kind: model.DocumentableKind) -> model.DocumentableKind:
 
     return (-o.privacyClass.value, -map_kind(o.kind).value if o.kind else 0, o.fullName().lower())
 
-def format_decorators(obj: Union[model.Function, model.Attribute]) -> Iterator["Flattenable"]:
+def format_decorators(obj: Union[model.Function, model.Attribute, model.FunctionOverload]) -> Iterator["Flattenable"]:
+    # Use the overload's primary function for some parts here
```
tristanlatr, line 49: Suggesting a more verbose comment here, thanks for the explanations.

```suggestion
    # Since we use this function to colorize the FunctionOverload decorators and it's not an actual Documentable subclass, we use the overload's 
    # primary function for parts that requires an interface to Documentable methods or attributes
```

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -57,19 +60,46 @@ def format_decorators(obj: Union[model.Function, model.Attribute]) -> Iterator["
...
+        yield from format_decorators(overload)
+        yield tags.div(format_function_def(func.name, func.is_async, overload))
+
+def format_function_def(func_name: str, is_async: bool, 
+                        func: Union[model.Function, model.FunctionOverload]) -> List["Flattenable"]:
+    """
+    Format a function definition.
+    """
+    r:List["Flattenable"] = []
+    # If this is a function with overloads, we do not render a definition
```
tristanlatr, line 91:
```suggestion
    # If this is a function with overloads, we do not render the principal signature because the overloaded signatures will be shown instead.
```

pydoctor/test/test_astbuilder.py
```diff
@@ -1719,6 +1719,43 @@ def func():
...
+    assert flatten_text(html2stan(str(func.signature).replace(' ', ''))) == '(s:Union[str,bytes])->Union[str,bytes]'
+    assert [astbuilder.node2dottedname(d) for d in (func.decorators or ())] == []
+    assert len(func.overloads) == 2
+    assert [astbuilder.node2dottedname(d) for d in func.overloads[0].decorators] == [['dec'], ['overload']]
+    assert [astbuilder.node2dottedname(d) for d in func.overloads[1].decorators] == [['overload']]
+    assert flatten_text(html2stan(str(func.overloads[0].signature).replace(' ', ''))) == '(s:str)->str'
+    assert flatten_text(html2stan(str(func.overloads[1].signature).replace(' ', ''))) == '(s:bytes)->bytes'
+    assert capsys.readouterr().out.splitlines() == [
+        '<test>:5: <test>.parse overload has docstring',
+        '<test>:5: <test>.parse overload appeared after primary function',
```
tristanlatr, line 1756: Do you think it would be possible to get the right line number for these warnings ? 
There is an offset feature to the report() method. So it might be possible to use that to report the warnings at the right line. 

Tell me if you think I'm being too zealous.

pydoctor/test/test_epydoc2stan.py
```diff
@@ -173,17 +174,59 @@ def get_answer() -> int:
...
+        '</td>', '</tr>', '</table>', '</div>',
+    ]
+    assert lines == expected_html, str(lines)
+
+def test_func_only_return_doc() -> None:
+    """When only return is documented but not parameters, only the return
+    section is visible.
+    """
+    mod = fromText('''
+    def f(x: str):
```
tristanlatr, line 216: Would it be possible to test this with and without annotations? Could be simple with a parameterized test

pydoctor/test/test_epydoc2stan.py
```diff
@@ -173,17 +174,59 @@ def get_answer() -> int:
...
+        '</div>',
+    ]
+    assert lines == expected_html, str(lines)
+
+def test_func_only_single_param_doc() -> None:
+    """When only a single parameter is documented, all parameters show with
+    undocumented parameters marked as such.
+    """
+    mod = fromText('''
+    def f(x, y):
```
tristanlatr, line 188: Would it be possible to test this with and without annotations? Could be simple with a parameterized test

pydoctor/test/test_templatewriter.py
```diff
@@ -484,6 +484,31 @@ class _Private(_BaseForPrivate):
...
+    func = mod.contents['parse']
+    assert isinstance(func, model.Function)
+    
+    # We intentionally remove spaces before comparing
+    overloads_html = stanutils.flatten(list(pages.format_overloads(func))).replace(' ','')
+    assert '''(s:str)-&gt;str:''' in overloads_html
+    assert '''(s:bytes)-&gt;bytes:''' in overloads_html
+
+    # Confirm the actual function definition is not rendered
+    function_def_html = stanutils.flatten(list(pages.format_function_def(func.name, func.is_async, func)))
```
tristanlatr, line 509: Isn't it a bit weird that the format_function_def() returns nothing for overloads ? I believe it's OK, but should be documented in the docstring.

pydoctor/templatewriter/pages/__init__.py
```diff
@@ -57,19 +60,46 @@ def format_decorators(obj: Union[model.Function, model.Attribute]) -> Iterator["
...
+    Format a function overloads definitions as nice HTML signatures.
+    """
+    for overload in func.overloads:
+        yield from format_decorators(overload)
+        yield tags.div(format_function_def(func.name, func.is_async, overload))
+
+def format_function_def(func_name: str, is_async: bool, 
+                        func: Union[model.Function, model.FunctionOverload]) -> List["Flattenable"]:
+    """
+    Format a function definition.
```
tristanlatr, line 88:
```suggestion
    Format a function definition as nice HTML signature. 
    
    If the function is overloaded, it will return an empty list. We use L{format_overloads} instead.
```

pydoctor/themes/base/function-child.html
```diff
@@ -8,6 +8,7 @@
     <t:attr name="name" t:render="shortFunctionAnchor" />
```
tristanlatr, line 8: When a template changed, we track it into a independent version system such that customizations that overrides this template will get a warning saying they should update.

Can you increase the `<meta name="pydoctor-template-version" content="2" />` to `<meta name="pydoctor-template-version" content="3" />` please?

tristanlatr (commented):
Overhaul looks good, only minor suggestions.

pydoctor/test/test_epydoc2stan.py
```diff
@@ -239,8 +282,8 @@ def f(a, b):
     classic_fmt = docstring2html(classic_mod.contents['f'])
     assert annotation_fmt == classic_fmt
 
-@pytest.mark.xfail
 def test_func_arg_when_doc_missing() -> None:
+    # Output with annotated args left undocumented are the same as undocumented
     annotation_mod = fromText('''
     def f(a: List[str], b: int) -> bool:
         """
```
tristanlatr, line 289: These is something that bothers me, which is probably the reason why this test has been marked as xfail.
We are actually loosing some data in the second test down here at line 294. The `@type:` fields are getting ignored if they are not accompanied by a `@param:` field. They should be included included in the documentation for sure, so maybe we should consider

pydoctor/epydoc2stan.py
```diff
@@ -474,8 +475,10 @@ def resolve_types(self) -> None:
     def format(self) -> Tag:
         r: List[Tag] = []
 
-        r += format_desc_list('Parameters', self.parameter_descs)
-        if self.return_desc:
+        # Only include parameter or return sections if any are documented
+        if any(p.body for p in self.parameter_descs):
```
tristanlatr, line 479: There is something that bothers me, here. (see other comment). 

We need a way to determine of the type info is coming from AST, in which case we should display it only if it's documented or if the type is coming from docstring fields, in which case it should display anyhow (with undocumented notes). In terms of implementation, the `resolve_types()` might need an adjustment in order to track if the `FieldDesc` is create from `resolve_types()` (meaning we only have AST info), for instance by creating `_BaseFieldDesc` superclass of `FieldDesc` and `AstFieldDesc` and use that instead from `resolve_types()`. Then check for the existence of `FieldDesc` instances in `self.parameter_descs`. 

Tell me if you think it could work,

Talk to you later,

tristanlatr (changes requested):
Hello @cretz,

Thanks for your work on this, turns out it's more complicated than expected (and we're missing regression tests). I believe more reasoning is needed to honour orphan `@type:` fields that don't have associated documentation. 

Talk to you later,

pydoctor/epydoc2stan.py
tristanlatr: This might now work since we want to track the origin of the type info, and this would only work to track undocumented params…

cretz:
Will look/respond/fix probably early next week.

cretz:
@tristanlatr - Sorry, I have been caught up getting our Python SDK going and have not been able to prioritize docs. I will try to revisit in a few days.

pydoctor/test/test_epydoc2stan.py
cretz: > They should be included included in the documentation for sure.

Hrmm, why? Has this been the case always? One might expect missing param docs to be treated the same way regardless of how their types are defined, right?

pydoctor/epydoc2stan.py
cretz: > We need a way to determine of the type info is coming from AST, in which case we should display it only if it's documented or if the type is coming from docstring fields, in which case it should display anyhow (with undocumented notes).

This seems a bit off. If someone has chosen to use `@type` (maybe to help their IDE or whatever) but have chosen not to use `@param`, they are choosing to not document, right?

It seems you are suggested that `@type` be treated as an empty `@param` doc? How would I, as a user that doesn't use `@type`, be able to get that empty documentation functionality the way those that _do_ use `@type` get?

cretz (commented):
@tristanlatr - I am now back available. Sorry for the hiatus. I have added response on the primary point about treating the two ways of annotating types differently. Once we land on a solution there, I'll fix conflicts and apply other needed changes.

pydoctor/epydoc2stan.py
cretz: > because we include them right now, and would ignore them in the next version, this would be a breaking change

This to me is enough to justify it. So to confirm - Include docs if `@type` _or_ `@param` present, otherwise no docs if neither present (regardless if type hinting). Sound right? I can definitely add the type origin on the desc there.

cretz:
@tristanlatr - Sorry, this keeps getting pushed down on my priority list :-) It is still definitely on my TODO (this powers https://python.temporal.io/ for us), but cannot promise exact time.

cretz:
@tristanlatr - Thanks for helping this move forward! (pardon my absence)

glyph:
> Any trepidation about this change in twisted docs ? @glyph
> Personally, I think it's fine, but I'de like to hear it from you.

We don't use a lot of overloads yet, and as I understand it, despite the docs being arguably slightly "uglier" with this change, the ugliness is a reflection of real-world complexity that is currently just missing.  So I would regard this as an improvement.

I think that having an option to toggle wrapping, specifically Black-compatible wrapping, would be good to have, but that is an unrelated problem for function signatures, so I don't think it has any bearing on this PR.

PR #598, opened 2022-06-04
Update docs

tristanlatr:
The issue has been fixed in #575. We don’t need to talk about it in the documentation anymore.


tristanlatr:
I'll merge this PR now since it only fixes documentation.

PR #602, opened 2022-06-12
Fix sidebar

### LINKED ISSUE #600: Hide the sidebar title when all items are private (and show it when show private is True)
"Variables" should only show up if Show private is on, the solution is pretty simple: adding the `private` css class to the title div if all items are private.

by default 2022-06-11 at 10 22 35 PM

https://github.com/twisted/pydoctor/blob/c951116c8b105aca507594deaa37b39b9413874f/pydoctor/templatewriter/pages/sidebar.py#L115

tristanlatr:
Fixes #600

### DIFF
--- a/pydoctor/templatewriter/pages/sidebar.py
+++ b/pydoctor/templatewriter/pages/sidebar.py
@@ -125,10 +125,10 @@ class ObjContent(Element):
         self._depth = depth
         self._level = level + 1
 
-        _classList = []
-        _functionList = []
-        _variableList = []
-        _subModuleList = []
+        _classList: List[Documentable] = [] # rha, mypy...
+        _functionList: List[Documentable] = []
+        _variableList: List[Documentable] = []
+        _subModuleList: List[Documentable] = []
 
         _direct_children = self._children(inherited=False)
         for child in _direct_children:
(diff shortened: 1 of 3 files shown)


pydoctor/templatewriter/pages/sidebar.py
```diff
@@ -166,7 +172,7 @@ def _children(self, inherited: bool = False) -> Optional[List[Documentable]]:
                 return sorted((o for o in util.list_inherited_members(self.ob) if o.isVisible),
                               key=util.objects_order)
             else:
-                return None
+                assert False, "Use inherited=True only with Class instances"
```
glyph, line 175: I appreciate trying to keep the diff small, but assertions in the negative path like this are somewhat confusing to read; it seems like there's going to be a path through at first and then there isn't.  When reading the assertion in a traceback it's less clear what's being asserted. Instead, I'd do:

```python3
assert isinstance(self.ob, Class), "Use inherited=True only with Class instances"
return sorted((o for o in util.list_inherited_members(self.ob) if o.isVisible), key=util.objects_order)
```

Sorry that's not a suggestion, but it modifies lines outside the diff…

pydoctor/templatewriter/pages/sidebar.py
```diff
@@ -125,39 +127,43 @@ def __init__(self, loader: ITemplateLoader, ob: Documentable, documented_ob: Doc
...
-            return None
+        self.classList = self._getContentList(_direct_children, Class)
+        self.functionList = self._getContentList(_direct_children, Function)
+        self.variableList = self._getContentList(_direct_children, Attribute)
+        self.subModuleList = self._getContentList(_direct_children, Module)
+        
+        self.inheritedFunctionList: Optional[ContentList] = None
+        self.inheritedVariableList: Optional[ContentList] = None
 
+        if isinstance(self.ob, Class):
```
glyph, line 140: Longer-term question: could this be added as a method to `Documentable` so we can just call it on `self.ob` rather than dispatching on type?

pydoctor/test/test_mro.py
```diff
@@ -221,3 +221,22 @@ def z():...
     assert isinstance(klass, model.Class)
     assert klass.subclasses == [dimond.contents['A'], dimond.contents['B']]
     assert list(util.overriding_subclasses(klass, 'z')) == [dimond.contents['A'], dimond.contents['C']]
+
+def test_list_inherited_members() -> None:
```
glyph, line 225: Can you add a docstring to this test explaining what is being tested, per https://jml.io/pages/test-docstrings.html ?

glyph (approved):
LGTM!

PR #607, opened 2022-07-07
fix intersphinx

tristanlatr:
Fixes twisted intersphinx link


adiroiban:
@tristanlatr what is needed to fix the upstream twisted docs ?

Note that we have:

* `https://docs.twisted.org/en/stable/api/objects.inv` - latest released
* `https://docs.twisted.org/en/latest/api/objects.inv` - latest commit in trunk / main branch

PR #619, opened 2022-07-28
Fix search.js

### LINKED ISSUE #614: Make the search bar disappear when clicking on an search result
It can happen in rare cases that the object we’re looking for is defined on the same page we’re executing the search from. 

Modern browsers seems not to reload the complete page even if the link is absolute, leaving the search bar open while the interesting content might be hidden behind.

So we need explicit JavaScript to close the search result div.

### LINKED ISSUE #617: Message in search bar should disappear when index size is over 20MB
When the index size is over 20MB, pydoctor does not trigger the search-as-you-type feature for UX reasons, users need to press the Enter key to launch the search. 

But when editing the search field after the search has been executed once, the message still appears even if no results are shown.

by default 2022-07-27 at 10 14 38 PM

tristanlatr:
Fixes #614  and fixes #617

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -76,6 +76,7 @@ What's New?
 in development
 ^^^^^^^^^^^^^^
 * Fix minor UX issues in the search bar.
+* Fix deprecation in Docutils 0.19 frontend
 
 pydoctor 22.7.0
 ^^^^^^^^^^^^^^^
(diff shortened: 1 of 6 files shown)


adiroiban (approved):
I have no idea what is going on there, but as long as it work is ok.

I don't know how to reproduce 20+MB on pydoctor own docs  at https://pydoctor--619.org.readthedocs.build/en/619/api/

=====

I am not a big fan of query.
For JS I prefer, alpine or vuejs, to add a bit of structure and conventions to the code.

PR #625, opened 2022-07-29
Better checks before removing 'self' and 'cls' from parameter tables

### LINKED ISSUE #605: Fix self and cls in function (not methods) parameter tables
Only strip `self` and `cls` from parameter table if the parent context is a class, otherwise include `self` or `cls`.

https://github.com/twisted/pydoctor/blob/19e29766d4ad5b8226fe7cb4103776ba508b9c66/pydoctor/epydoc2stan.py#L457

These two functions should be treated the same manner:

by default 2022-06-15 at 3 54 53 PM

tristanlatr:
Fixes #605

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,7 +75,8 @@ What's New?
 
 in development
 ^^^^^^^^^^^^^^
-* Better checks before removing 'self' and 'cls' from parameter tables.
+* `self` parameter is now removed only when the target is a method. In the previous version, it was always removed in any context.
+* `cls` parameter is now removed only when the target is a class method. In the previous version, it was always removed in any context.
 * Add anchors aside attributes and functions to ease 
   the process of sharing links to these API docs.
 * Fix a bug in the return clause of google-style docstrings 


README.rst
```diff
@@ -75,6 +75,7 @@ What's New?
 
 in development
 ^^^^^^^^^^^^^^
+* Better checks before removing 'self' and 'cls' from parameter tables.
```
adiroiban, line 78: Maybe a bit more explicit 

```suggestion
* `self` parameter is now removed only when the target is a method. In the previous version, it was always removed in any context.
* `cls` parameter is now removed only when the target is a class method. In the previous version, it was always removed in any context.
```

adiroiban (approved):
The changes looks good.

For my tests, is better to leave all the arguments as they are... but I guess that in this way, the docs are less verbose.

But don't listen to much to my taste and I am not an avid user of api docs... I prefer to read the source code :)

I think that this can be merge.
There is not much use in waiting in the queue withot people reviewing it.

if anyone has anything agains it, we can look into this later.

Thanks again!

PR #628, opened 2022-07-30
Fix returns section of google docstrings

### LINKED ISSUE #627: Fix Napoleon
See this Sphinx ticket for details: https://github.com/sphinx-doc/sphinx/issues/9932

tristanlatr:
Fixes #627

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,7 +75,10 @@ What's New?
 
 in development
 ^^^^^^^^^^^^^^
-* Trigger warnings for unknown config options
+* Fix a bug in the return clause of google-style docstrings 
+  where the return type would be treated as the description 
+  when there is no explicit description.
+* Trigger warnings for unknown config options.
 * Fix minor UX issues in the search bar.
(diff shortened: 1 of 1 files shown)


pydoctor/test/test_napoleon_docstring.py
```diff
@@ -466,6 +466,51 @@ class GoogleDocstringTest(BaseDocstringTest):
         """
 Single line summary
 
+Returns:
```
adiroiban, line 469: I don't see where these new docstrings are tests.

And by a quick read I don't understand the testing principles for GoogleDocstringTest (no docstring for this class)

Instead of `Single line summary` my suggestion is to use a text describing the testing purpose of each of these examples.

adiroiban (approved):
Thanks for your work on pydoctor.

I don't have time to review these PRs.

I think that as long as the tests are green, it is safe to merge any PRs.

If anyone is again this process they can start helping with the review.

But without an active reviewer, I don't see any value from blocking the PR and not merging them without a review.

PR #635, opened 2022-07-31
Add explicit support for exceptions

### LINKED ISSUE #364: Exceptions types

tristanlatr:
Fixes #364 

Example: https://pydoctor--635.org.readthedocs.build/en/635/api/pydoctor.templatewriter.TemplateError.html

### DIFF
--- a/.github/workflows/system.yaml
+++ b/.github/workflows/system.yaml
@@ -12,7 +12,7 @@ jobs:
 
     strategy:
       matrix:
-        tox_target: [
+        tox_target: [twisted-apidoc,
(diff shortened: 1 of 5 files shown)


adiroiban (approved):
thanks. as long as this is green I think it can be merged.

adiroiban:
> codecov/patch — 100.00% of diff hit (target 91.55%)

It looks good.

Which like is missing?

PR #640, opened 2022-08-05
Better property support

### LINKED ISSUE #587: Unify properties getter, setters and deleters under a single documentation entry
It would good to see all the information about a property inside a single documentation entry. 

The main issue is what to do if several object have docstrings assigned to ? Should they be merge them together? Should we display them in separate boxes? 

Or maybe we should log a warning an use only the getter docstring. This would be the most simple solution and also comply with the default behavior of `help()`, but we’ve been using docstring on property setter in our demo. So we should probably look at a way of still supporting that.

Anyway the information whether this property is writable or deletable or read only can be added as `extra_info`.

tristanlatr:
The property setter and deleter informations now shows at the same place as the property getter. Meaning inside the documentation of an `Attribute` object.

Fixes #587. Together with #595 it will look better (without the Undocumented param tables). 

As a side effect of the change, the **setters** and **deleters** `Function` are removed from the documentable tree, we can't look for the property setters in the search bar with `*.setter`, we can't link to them from docstrings (but still we can link to them with their anchor) and they won't display in intersphinx mapping. 

This is a breaking change, but that might not be very important. These objects do not actually exists at runtime, so I think it's OK to remove them from the tree.

Before:

by default 2022-08-05 at 12 38 18 AM

After:

by default 2022-08-05 at 12 23 58 AM


tristanlatr (commented):
More tests should be added to ensure we’re a not crashing with KeyError when objects are getting renamed because there are duplicates or moved around because they’re reparented.

tristanlatr (commented):
Few things needs to be improved here:
- Trigger warnings when pydoctor doesn't understand a property def
- Probably the code can be factored into less lines

PR #645, opened 2022-08-18
Fix broken option ``--prepend-package`` in pydoctor 22.7.0

tristanlatr:
Fix broken option ``--prepend-package`` in pydoctor 22.7.0

### DIFF
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -1387,6 +1387,10 @@ class SystemBuilder(ISystemBuilder):
 System.systemBuilder = SystemBuilder
 
 def prepend_package(builderT:Type[ISystemBuilder], package:str) -> Type[ISystemBuilder]:
+    """
+    Get a new system builder class, that extends the original C{builder} such that it will always use a "fake" 
+    C{package} to be the only root object of the system and add new modules under it.
+    """
     
     class PrependPackageBuidler(builderT): # type:ignore
         """
--- a/pydoctor/test/test_astbuilder.py
+++ b/pydoctor/test/test_astbuilder.py
@@ -1975,6 +1975,11 @@ def test_reexport_wildcard(systemcls: Type[model.System]) -> None:
 
 @systemcls_param
 def test_prepend_package(systemcls: Type[model.System]) -> None:
+    """
+   Option --prepend-package option relies simply on the L{ISystemBuilder} interface, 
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -75,6 +75,7 @@ What's New?
 
 in development
 ^^^^^^^^^^^^^^
+* Fix broken option ``--prepend-package`` in pydoctor 22.7.0.
```
adiroiban, line 78: maybe a bit of more info for the release notes.

```suggestion
* The ``--prepend-package`` now work as documented and DOES-EXPECTED-BEHAVIOUR. A regression was introduced in pydoctor 22.7.0 and it was DOING-UNEXPECTED-BEHAVIOUR.
```

pydoctor/model.py
```diff
@@ -1385,3 +1385,35 @@ def buildModules(self) -> None:
         self.system.process()
 
 System.systemBuilder = SystemBuilder
+
+def prepend_package(builderT:Type[ISystemBuilder], package:str) -> Type[ISystemBuilder]:
```
adiroiban, line 1389: maybe add a docstring here as a way to document to other dev what is the low-level purpose of this method.

pydoctor/test/test_astbuilder.py
```diff
@@ -1971,3 +1972,32 @@ class j: pass
     assert system.allobjects['top._impl'].resolveName('f') == system.allobjects['top'].contents['f']
     assert system.allobjects['_impl2'].resolveName('i') == system.allobjects['top'].contents['i']
     assert all(n in system.allobjects['top'].contents for n in  ['f', 'g', 'h', 'i', 'j'])
+
+@systemcls_param
+def test_prepend_package(systemcls: Type[model.System]) -> None:
+    system = systemcls()
```
adiroiban, line 1978: Maybe add a docstring to these tests to  explain in simple words what is the purpose of this tests.
What is specific to this tests.

This shold be done to make it clear why we need another test like `test_prepend_package_real_path`

adiroiban (approved):
Thanks for the update.

Sorry, I don't have time to do reviews.

I only left some superficial comments.

As long as the tests are green, I think that this should be merged.

Thanks again for working on pydoctor.

README.rst
adiroiban: I think that this was not applied... can be applied as a drive-by for another PR .
No problem.

PR #646, opened 2022-09-08
Fixes the broken CI

tristanlatr:
Fixes the broken CI


adiroiban (approved):
hope it will be green after this change.

Thanks!

PR #656, opened 2022-10-17
Handling of attrs and dataclass constructors

### LINKED ISSUE #596: Create an extension for dataclass based classes
`@dataclasses.dataclass` can be used from python3.7, it acts a bit like `@attr.s`, so adding support for it shouldn't be too complicated.

Also I realize that there are many modules that converts class variables into instance variables (pydantic, dataclasses, attrs, django models, and others), so there is probably something to do to avoid to much code duplications.

### LINKED ISSUE #718: Add support for the newer APIs of attrs
Attrs introduced the `define` APIs and friends a while ago now. Would be good to have support for those as well. I might use this opportunity to abstract out some of the code that handle all the dataclass-like classes so it’s easier to reuse and implement #596. This should probably be done before we merge #656.

### LINKED ISSUE #890: dataclass annotated attribute docstrings are incorrectly documented as class variables
A class that looks like this:

```python3
from dataclasses import dataclass

@dataclass
class Data:
    attribute: str
    "the 'attribute' data attribute of 'Data'"
```

will be documented as if `Data.attribute` is a class variable, when the `dataclass` decorator makes it an instance variable.

tristanlatr:
This PR drastically improve our understanding of attrs and dataclass generated classes. 

It will infer the `__init__` signature and present fields docstrings in the parameter table.
It understand combinations of auto_attribs, auto_detect, kw_only and init arguments to the attr.s() function. As well as the new APIs attrs.define(). For attr.ib(), it checks the converter, factory, default, type, kw_only and init arguments to infer each constructor param.

It only handles the creation of the `__init__` method, it’s does not create `__eq__` and other attrs methods. Since these methods all have the same signature, it’s probably not worth it to actually create Function instance to represent them. One thing we could do it to add a note at the end of the docstring listing implemented methods. Another thing to do would be to show the class decorators so the reader can determine whether all compare methods are implemented.

See documentation here: https://pydoctor--656.org.readthedocs.build/en/656/codedoc.html#using-attrs

Partial Fix for #305, handles the dataclass and attrs case.
Fixes #890 

Currently, It will sometimes generate a wrong constructor arguments order for dataclasses and classic attrs based classes  that have multiple inheritance and attribute collisions. This is because there are two manner of linearizing the parameters and I only implemented the “collect by MRO” order.


glyph (approved):
I think the generated docstring could be a bit better, but otherwise this seems great. No need to think super hard about it though, if you can think of a *slightly* better boilerplate we can improve it more later.

PR #681, opened 2023-02-23
Show constructors

tristanlatr:
This is the core of PR #656, but without the attrs support, which will come after. 

Looks like this: 

Capture d’écran, le 2023-02-23 à 13 22 13

Ideally we would use a presentation that could show parameters types on mouse over, but we’d need to do more work. Right now it uses the rst parser to produce the Stan. So it’s rather inflexible. But it can be improved later by directly generate Stan. 

Currently it shows the constructor functions induces by __init__ or __new__ methods as well as class methods that uses self reference type hint as the return type. 

A possible improvement would be to support inherited class method constructors if it returns the self type (whether it’s by the Self annotation or returning cls()).

Also it would be good to hide the constructors that don’t have any docstring and no arguments, because it’s the default constructor.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -92,6 +92,7 @@ in development
 * Cache the default docutils settings on docutils>=0.19 to improve performance.
 * Improve the search bar user experience by automatically appending wildcard to each query terms
   when no terms already contain a wildcard. 
+* Link recognized constructors in class page.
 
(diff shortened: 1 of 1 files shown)


glyph (approved):
LGTM, let's land it!

PR #682, opened 2023-03-29
Add a new_document helper that uses a cached default settings if none…

dreid:
… is specified.


dreid:
Hi @tristanlatr 

> Hi @dreid,
> 
> Can you explain the advantages of this change? I suspect this is a micro optimization, but une not sure. Thanks.

I've updated the description with some timing information from my work codebase which I've been trying to improve the documentation build time of.   I don't know that I'd call it a micro-optimization, according to the profile data `new_document`'s invocation of `get_default_settings` seems to be ~23% of runtime.

dreid:
> Would it be possible for you to regenerate the speedscope dashboard using the latest version of pydoctor?

@tristanlatr Updated speedscope is [here](https://www.speedscope.app/#profileURL=https://gist.githubusercontent.com/dreid/8f14c18b468c67376dfe005d09a76238/raw/d9addc3d169b83281ad22f81aa8d1f4dec9966d6/pydoctor.speedscope) and I've documented how I generate this [here](https://gist.github.com/dreid/8f14c18b468c67376dfe005d09a76238#file-readme-md)

It is worth noting that this particular profile is on a significantly faster computer than I was using previously (an m1 instead of a ~4yo x86 mac)

PR #683, opened 2023-03-31
Mypy 1.x compat

dreid:
This addresses two issues that come up in mypy > 1.x

First simply removes an unused `type: ignore` comment.

Second is that `ParsedDocstring.has_body` is declared as an `abstractproperty` but `ParsedStanOnly.has_body` was not defined as a `property` fortunately (I guess) functions are truthy and this property was intended to always return True.

### DIFF
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -5,8 +5,8 @@ Convert L{pydoctor.epydoc} parsed markup into renderable content.
 from collections import defaultdict
 import enum
 from typing import (
-    TYPE_CHECKING, Any, Callable, ClassVar, DefaultDict, Dict, Generator, 
-    Iterator, List, Mapping, Optional, Sequence, Tuple, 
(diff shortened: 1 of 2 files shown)


dreid:
> I don't really like the fact that you're including formatting changes as well as fixes.
> 
> But I'll merge the PR anyway for this time. Next time please keep that in mind. It might be your IDE that does this automatically...

I don't really like it either, but it became extremely tedious to avoid.  Would you be open to a PR that included a standard formatter in CI and did all the formatting changes at once?

glyph:
> Right now, I simply don't want to review such a huge PR (if the changes are applied on the developer side).

I only have a moment to look at this now so I can't remember the exact specifics, but I know that the way that Twisted adopted Black avoided the issue of large PRs because it explicitly omitted the relevant commits from `blame` and reformatted outstanding PRs to minimize diffs. I don't think it was a lot of work, and it may be worth doing sooner rather than later both to unblock this PR and to deal with this increasingly-common situation in advance

PR #684, opened 2023-03-31
Tweak search

### LINKED ISSUE #648: Search bar doesn't always yield the expected results
When searching for "ensure parsed docstring", I was expecting to see the [pydoctor.epydoc2stan.ensure_parsed_docstring()](https://pydoctor.readthedocs.io/en/latest/api/pydoctor.epydoc2stan.html#ensure_parsed_docstring) function show up, but it doesn't:

by default 2022-09-08 at 1 29 50 PM

Maybe we can workaround the issue by adding wildcard at the end of the search query if the result did not yield any results, but it would be an uncommon behaviour. I'm unsure of the root cause if the issue, the lunr index building / searching seem quite complicated and might be the issue here. Because I've tried to make sure the object are also indexed by their split names as well, such that thins kind of query works. But it doesn't always.

tristanlatr:
Fixes #648

### DIFF
--- a/pydoctor/themes/base/searchlib.js
+++ b/pydoctor/themes/base/searchlib.js
@@ -61,7 +61,7 @@ onmessage = (message) => {
         var new_clauses = [];
         if ((message.data.autoWildcard == true) && (hasTraillingWildcard == false)){
             _query.clauses.forEach(clause => {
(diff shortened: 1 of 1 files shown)


ntamas:
Looks good to me, thanks a lot!

PR #698, opened 2023-05-23
Special case **kwargs in parameter table

### LINKED ISSUE #697: How should we document Variable-length arguments (**kwargs)
I'd like to document `**kwargs` like https://docker-py.readthedocs.io/en/stable/networks.html

For example, here is a method:

```python
    def test(**kwargs):
        """
        Test pydoctor.

        Parameters
        ----------
        key1: str
            The 1st key.
        key2: str
            The 2nd key. Supported format:

            * file:///path/to/key2
            * http://path/to/key2
        """
        print(kwargs)
```

I saw the warnings:

```
__init__.py:339: Documented parameter "key1" does not exist
__init__.py:341: Documented parameter "key2" does not exist
```

image

The result looks fine except a line that indicates `**kwargs` is undocumented.

tristanlatr:
Do not show `**kwargs` when keywords are specifically documented with the `keyword` field   and no specific documentation is given for the `**kwargs` entry

Fixes #697

### DIFF
--- a/pydoctor/test/test_sphinx.py
+++ b/pydoctor/test/test_sphinx.py
@@ -595,6 +595,7 @@ class TestIntersphinxCache:
             def send(
                     self: requests.adapters.HTTPAdapter,
                     request: requests.PreparedRequest,
+                    *args:object,
                     **kwargs: object
                     ) -> requests.Response:
                 response: requests.Response


BewareMyPower:
image


Yeah, it works for me now. Thanks!

pydoctor/test/test_epydoc2stan.py
```diff
@@ -695,6 +695,89 @@ def f(args, kwargs, *a, **kwa) -> None:
     captured = capsys.readouterr().out
     assert not captured
 
+def test_func_starargs_hidden_when_keywords_documented() -> None:
+    # tests for issue https://github.com/twisted/pydoctor/issues/697
```
adiroiban, line 699: Thanks for the PR.
Just a quick feedback.

I am looking a the code and the tests, but I am not sure what are the specifications here / how we expect each corner case to behave.

It looks like there are 3 separate tests here.
It would be nice to have a separate docstring for each test, explaining the expected behaviour

So rather than having a link to an issue, describe the specific behaviour under test.

I am readying the input test data, and I see that it has some documentation there, but I think that is best to follow the convention and have all the test related documentation in the test method docstring.
And be as verbose as needed.

For example:

``
When a function has C{args} or C{kwargs} as explicit arguments, and not the positional or keywords arguments placeholders, they are handled just as any other regular argument.

They keyword argument placeholder documentation is option
and when missing, no warning is raised.
```



``
When a function has C{args} or C{kwargs} as explicit arguments but also the keywords arguments placeholder, hey keywords placeholder can have dedicated docstring, besides the separate documentation for each implicit keyword.
```

adiroiban: Well... `test_func_starargs_wrongly_documented` has kind of the same issue... it would be nice to have it split into a RST version and a numpy version, with dedicated docstrins.

The reason is that if there is a bug RST handling of `docstring2html` you can put some breakpoints and run a single test to troubleshoot this.

With the current code, your breakpoint might unnecessarily stop for the numpy code that you know is ok.

This is all hypothetical.

-------------

But I only did a very quick review , out our curiosity :)


I think that as long as the original issue report is happy with the result and the tests are green, this can be merged.

Thanks again for looking after pydoctor.

PR #722, opened 2023-06-15
fix: Hardlink to index.html

### LINKED ISSUE #720: Symlink to index.html
It seems that pydoctor do create a a symlink (named with the package name) to `index.html`.

I couldn't find the right place in code where this happens.

I'm asking to modify this somehow. Maybe it could be a hard link instead of a symlink.
Or it could be a regular html file with a redirect directive in it.

Symlinks do often confuse other software, e.g. nextCloud/ownCloud, some Backup software etc.

I can't think of a reason why it couldn't be a hardlink.

buhtz:
A hardlink is created instead of a symlink to `index.html`. Also added hypothesis entry to `.gitignore`.

I had to use `os` module because `pathlib.Path.hardlink_to()` is introduced in Python 3.10 but `pydoctor` to support Python 3.6.

It works in my productive environment.

Two tests (via pytest) are failing and I'm not sure why. They are well documented but I lack of deeper understanding of how pydoctor works to understand the problem.

I didn't use tox. Maybe you could update your contrib info about the purpose of your tox setup.

Btw: Where in the HTML output is that hardlink HTML file used? My package "buhtzology" do have a `buhtzology.html` hardlink to `index.html`. But nowhere in the HTML output I can find a link to a `buhtzology.html` file.

Your [contrib info](https://github.com/twisted/pydoctor/blob/master/docs/source/contrib.rst) (btw: missing a TOC) didn't tell about branches or branching model. So I branched from and into `master`. Hope this is OK that way.

Fix #720


buhtz:
None of my questions are answered.
And I do not understand the output of your Microsoft GitHub CI system in the back.

The repo does miss contributor documentation for things like this.
I could fix it all my self if I could understand it.

buhtz:
> Regarding the contributors documentation.. please keep in mind that we’re all volunteers here. Best documentation will always be to read the code.

I am a dev and a maintainer, too. But beeing a volunteer is IMHO not good excuse for not documenting. Code is not documentation. And contributors documentation does not have something to do with code by the way. It is more about how to project is organized and how to do PRs, how the test/build/CI system is intended to work etc pp.

But lets stop the meta discussion and lets go back to this PR.
How am I suppose to understand the failing Microsoft GitHub Action things?

How can I help you to fix this PR and get it merged.

buhtz:
> First, your changes makes pydoctor crash. The cause of this must still be investigated.

I don't see a crash. Can you link to the output in some of that Microsoft GitHub Actions CI jobs/machines?

> Secondly, there is a bug in pydoctor_primer that crashes.

I do not know what that is and it is not documented somewhere and not even in the primar yaml itself.
You would save yourself and future contributors a lot of time and words if you would document it.

> We need to examine the stack trace and fix the bug.

> already have a html build existing so your not seeing the crash. Please try to delete the output directory and see if pydoctor still works.

In that case I can reproduce this problem.
```
/home/user/ownCloud/my.work/buhtzology/docs
  File "/home/user/ownCloud/my.work/buhtzology/docs/./hey_doc.py", line 152, in <module>
    rc = pydoctor.driver.main(args)
  File "/home/user/ownCloud/my.work/pydoctor/pydoctor/driver.py", line 169, in main
    make(system)
  File "/home/user/ownCloud/my.work/pydoctor/pydoctor/driver.py", line 128, in make
    writer.writeSummaryPages(system)
  File "/home/user/ownCloud/my.work/pydoctor/pydoctor/templatewriter/writer.py", line 116, in writeSummaryPages
    os.link(root_module_path.parent / 'index.html', root_module_path)
FileNotFoundError: [Errno 2] No such file or directory: 'html/index.html' -> 'html/buhtzology.html'

```

Again: I can not see this error in your CI or tests. How can an extern but motivated contributor can see this? You should document this. Your project is not ready for extern contributors because PRs and even testing is possible only with project-intern knowledge.

You can blame me by the way: Your test suite do catch the problem very well. I just haven't run the tests on my local machine.

```
=================================================== short test summary info ===================================================
FAILED pydoctor/test/test_commandline.py::test_main_symlinked_paths - FileNotFoundError: [Errno 2] No such file or directory: 'apidocs/index.html' -> 'apidocs/basic.html'
FAILED pydoctor/test/test_commandline.py::test_main_source_outside_basedir - FileNotFoundError: [Errno 2] No such file or directory: 'apidocs/index.html' -> 'apidocs/basic.html'
ERROR pydoctor/test/epydoc/test_pyval_repr.py::test_expressions_parens

```

pydoctor/templatewriter/writer.py
```diff
@@ -108,7 +109,16 @@ def writeSummaryPages(self, system: model.System) -> None:
...
             except FileNotFoundError:
                 pass
-            root_module_path.symlink_to('index.html')
+
+            # When support for Python 3.9 and older is dropped use
+            # pathlib.Path.hardlink_to() instead.
+            x = list(root_module_path.parent.glob('*'))
+            print(f'TTTT\n{root_module_path=} {x=}')
+            print(f'{root_module_path.exists()=}')
+            os.link(
```
buhtz, line 118: OK, I need your help here. First of all I think I confounded the arguments in `os.link()`. So I changed them but this does not solve the problem.

The value of `root_module_path` is `PosixPath('apidocs/basic.html')` which does not exist. That seems to be the problem. But I do not understand why it does not exist or even if it have to exists.
Of course because of that the `os.link()` can not work.

buhtz: Make totally sense. Thanks for analyzing. Because this is not your own code, are you sure that your solution will do the job? Or I might should do some more research if there is a better location in code to call writeIndividualFiles()?

buhtz:
The solution is not that simple. I tried

```
def make(system: model.System) -> None:
    """
    Produce the html/intersphinx output, as configured in the system's options. 
    """

    # ...

        if options.htmlsubjects:
            subjects = [system.allobjects[fn] for fn in options.htmlsubjects]
            writer.writeIndividualFiles(subjects)
        else:
            writer.writeIndividualFiles(subjects)
            writer.writeSummaryPages(system)
```

Other tests then fail. (btw: I miss the opportunity to run a single test only. Not sure how to do this with tox. Pytest alone still do not work.)

I am not enough into the code to understand it and to decide about a solution. My suggest would be refactor `writer.py::TemplateWriter.writeSummaryPages()`  and `driver.py::make()`. Especially the first one is to complex and IMHO could be broken down into multiple simpler methods.

In the current state my understanding of the code and the logic behind it is not deep enough to find a solution myself.

PR #730, opened 2023-08-22
Fix CI

tristanlatr:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,6 +88,7 @@ in development
 * Improve the class hierarchy such that it links top level names with intersphinx when possible.
 * Add highlighting when clicking on "View In Hierarchy" link from class page.
 * Recognize variadic generics type variables (PEP 646).
+* Fix support for introspection of cython3 generated modules.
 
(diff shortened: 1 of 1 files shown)


glyph:
@tristanlatr the first thing to investigate here is to make sure that `mypy` and all its transitive dependencies are pinned.

glyph:
Once we know that the environment is consistent everywhere I can investigate if something trickier is going on :)

glyph:
> I’m not talking about the mypy failure, but rather the unit tests failure on python 3.10 and 3.11.

Ah. Hm. I don't know enough about pydoctor's extension-module introspection capabilities, but surely there's something in the stdlib that we could use that's known to be an extension module and that doesn't depend on the sad little https://pypi.org/project/cython-test-exception-raiser/ project?  It occurs to me that there's a 1.0.2, and I don't know what's in the chnagelog there, but maybe pinning the old version is part of the problem?

glyph:
(debugging this for far too long right now I was not able to figure out what's going on)

PR #735, opened 2023-09-24
Drop python 3.6 and enable future annotations everywhere 

### LINKED ISSUE #732: Fix the standard library smoke test
As a smoke test we run the tox environment `cpython-apidocs`.

This task pulls the latest CPython an run pydoctor on it. The trick is that is't set-up to unconditionally run with python 3.8 in the Github CI. All files that uses syntax introduced after python 3.8 (like the match statement or exception groups) are not being analyzed because the parser fails to produce the AST for these files.

We should pull a specific version of python, and run pydoctor with the corresponding python version instead.

### LINKED ISSUE #733: Drop support for python 3.6
It’s time to drop support for python 3.6. The next version will be last one to support python 3.6.

Once it’s done, we’ll need to enabled PEP 563 on every modules. Potentially doing a script that checks that all our files starts with the future import.

tristanlatr:
Fixes #733

Fixes #732 

And add python 3.12rc3 in the CI

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -79,6 +79,7 @@ in development
 This is the last major release to support Python 3.7.
 
 * Drop support for Python 3.6
+* Add support for Python 3.12
 * `ExtRegistrar.register_post_processor()` now supports a `priority` argument that is an int.
   Highest priority callables will be called first during post-processing.
 


.github/workflows/unit.yaml
```diff
@@ -14,13 +14,13 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9, 3.10.8, 3.11.0, pypy-3.6]
+        python-version: [3.7, 3.8, 3.9, '3.10', 3.11, '3.12.0-rc.3', pypy-3.7]
```
adiroiban, line 17: I think you can just use 3.12 and it should automatically get the latest 3.12 version.

Note that for Twisted, support was removed for 3.7 ... so you can expect test failures with twisted/trunk on 3.7

buhtz:
I don't know if I have enough expertise here.
I don't understand why you use types from the `ast` module instead of `typing` module which is IMHO recommended.

buhtz:
> > I don't understand why you use types from the ast module instead of typing module which is IMHO recommended.
> 
> Indeed you don’t quite understood. The typing annotations are provided for generics and protocols. But ast classes are not generics, there is only these classes available to annotate these instances.

Is this a limitation of older Python versions? I do use annotations, too. But I don't have to use ast package or `__futer__` magic.

buhtz (commented):
You see I'm not the one who should review that.

adiroiban (approved):
It looks good. Thanks.

I think that this can be merged.

PR #746, opened 2023-11-02
Setup pydoctor_primer

tristanlatr:


glyph:
> @glyph would you be ok that I transfert the [pydoctor_primer](https://github.com/tristanlatr/pydoctor_primer) repository to the twisted org ?

sounds good

PR #767, opened 2024-03-27
Release pydoctor 24.3.0

### LINKED ISSUE #794: Release automation is broken
The error is 
```
 ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         User 'twistedchecker-robot' does not have a verified primary email     
         address. Please add a verified primary email before attempting to      
         upload to PyPI. See https://pypi.org/help/#verified-email for more     
         information.   
```

@adiroiban @glyph 

Other warnings are: 

```
Warning: Input 'user' has been deprecated with message: UNSUPPORTED GITHUB ACTION VERSION

You are using `pypa/gh-action-pypi-publish@master`. The `master` branch of this project has been sunset and will not receive any updates, not even security bug fixes. Please, make sure to use a supported version. If you want to pin to v1 major version, use `pypa/gh-action-pypi-publish@release/v1`. If you feel adventurous, you may opt to use use `pypa/gh-action-pypi-publish@unstable/v1` instead. A more general recommendation is to pin to exact tags or commit SHAs.

Please also consider migrating your setup to use secretless publishing: https://github.com/marketplace/actions/pypi-publish#trusted-publishing

https://stand-with-ukraine.pp.ua/

Warning: Input 'password' has been deprecated with message: UNSUPPORTED GITHUB ACTION VERSION
```

tristanlatr:
Fixes #794

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,6 +73,9 @@ You can select a different format using the ``--docformat`` option or the ``__do
 What's New?
 ~~~~~~~~~~~
 
+in development
+^^^^^^^^^^^^^^
+
 pydoctor 24.3.3
 ^^^^^^^^^^^^^^^
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = pydoctor
-version = 24.3.3
+version = 24.3.3.dev0
 author = Michael Hudson-Doyle
 author_email = micahel@gmail.com
 maintainer = Maarten ter Huurne


adiroiban:
I have also removed the `PYPI_TOKEN` token from [secrets](https://github.com/twisted/pydoctor/settings/secrets/actions) as it should never be used again

adiroiban:
I have removed the ubuntu 20.04 required jobs as it looks like it is no longer used.

image


--------

adiroiban:
True. My bad.

I have added explicit job permissions.

You can delete the tag and create it again. It should re-trigger the job

adiroiban:
The pypi publishing works.

The package was rejected due to setup.cfg deps definition

https://github.com/twisted/pydoctor/blob/27f5d4d022c9e1194f6ef82d49642c7966969fa5/setup.cfg#L82


--------------

It looks like we need to do a release for Twisted.

Why do we  need the latest Twisted here ?

PR #772, opened 2024-04-03
Supports type parameters and type aliases (pep 695)

tristanlatr:
It looks like this: 
Capture d’écran, le 2024-04-02 à 20 52 53

TODO: 

- Support type aliases in class body, it's only supported at the module level currently
- Do not show the declaration of type variables as links

### DIFF
--- a/pydoctor/model.py
+++ b/pydoctor/model.py
@@ -346,12 +346,14 @@ class Documentable:
     def _localNameToFullName(self, name: str) -> str:
         raise NotImplementedError(self._localNameToFullName)
     
-    def isNameDefined(self, name:str) -> bool:
+    def isNameDefined(self, name:str, nonlocals:bool=True) -> bool:
         """
         Is the given name defined in the globals/locals of self-context?
         Only the first name of a dotted name is checked.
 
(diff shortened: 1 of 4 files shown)


pydoctor/epydoc2stan.py
```diff
@@ -282,13 +282,18 @@ def __init__(self, obj: model.Documentable):
...
         _linker = self.obj.docstring_linker
+
         formatted_annotations = {
             name: None if value is None
-                       else ParamType(safe_to_stan(colorize_inline_pyval(value, is_annotation=True), _linker,
-                                self.obj, fallback=colorized_pyval_fallback, section='annotation', report=False),
-                                # don't spam the log, invalid annotation are going to be reported when the signature gets colorized
-                                origin=FieldOrigin.FROM_AST)
+                       else ParamType(safe_to_stan(colorize_inline_pyval(value, 
+                                 refmap=model.type_param_refs(self.obj),
```
tristanlatr: There is a major problem regarding this way of resolving the type parameters. The fact that we resolve early references to type parameters and resolve lazily other references creates situations where resolving simply can't be accurate.

PR #791, opened 2024-05-07
Outline diff to sphinx in context of automation level

buhtz:
Accidentally discovered this FAQ section.

Correct me if I am wrong but I often tried Sphinx* in the past and was not able to auto-generated documentation based on source files. You always need a rst file.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -36,6 +36,7 @@ It also has support for `zope.interface <https://zopeinterface.readthedocs.io/en
 
 How is it different from ``sphinx-autodoc``
 -------------------------------------------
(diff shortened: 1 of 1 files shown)


buhtz:
> > Correct me if I am wrong but I often tried Sphinx* in the past and was not able to auto-generated documentation based on source files. You always need a rst file.
> 
> There are automated tools that are able to generate rst files from source code files, like sphinx-autoapi.

I never got them work. The point is that it is not Sphinx but a Sphinx extension. It is a workaround to make Sphinx do something that it was not invented for.

No matter if it is possible or not the point of comparing pydoctor with sphinx is that it is easy with pydoctor without using a config file or any in-between-file-format or an extension. PyDoctor is for API documentation and Sphinx is not.

Issue #728, opened 2023-07-20
Sphinx extension fails at end of build due to renamed output path


jgarbers:
I've just started using _pydoctor.sphinx_ext.build_apidocs_ in my documentation project, replacing Sphinx's _autodoc_. After adding the extension and _pydoctor_args_ to my _conf.py_,  _sphinx-build_ fails with this error:

`Handler <function on_build_finished at 0x1090307c0> for event 'build-finished' threw an exception (exception: [Errno 2] No such file or directory: '/Users/jgarbers3/lab/ev/docs/build/html/api' -> '/Users/jgarbers3/lab/ev/docs/build/html/api.sphinx_files')`

My _sphinx-build_ output path is set to "docs/build", and the extension configuration specifies "--html-output={outdir}/api/".

Reviewing the source for the extension, I think I have an idea of why this is failing. At the end of _on_builder_inited_, the extension renames the output path (here, "docs/build/html/api") to include a _.pydoctor_temp_ suffix. Then, in _on_build_finished_, the code tries to rename the *original* output path ("docs/build/html/api") to include a _.sphinx_files_ suffix. However, that path does not exist, since it was renamed to  "docs/build/html/api.pydoctor_temp" at the end of the build init event handler, and the exception shown above triggers.

I've patched the extension to work around the error, and as a newcomer to pydoctor it seems likely that I've missed something. But I don't see any other references to what a _.sphinx_files_ directory might be for, so I'm wondering if that rename step might have been left in inadvertently.

I'm using Python 3.11.3, Sphinx 7.0.1 and the pydoctor extension 23.4.1.

Thanks in advance for any guidance anyone can offer!

jgarbers:
Thanks @tristanlatr ! I'll try to do a minimal reproducer later today. 

My workaround patch just checks for the existence of the output path before trying to rename it, but as far as I can tell the output path will _never_ exist, so this is just a band-aid in lieu of an actual understanding of the problem.

I do have a number of warnings being generated during the pydoctor build process -- lots of docstring formatting errors to fix! -- so that aligns with your observations of failing with`No such file or directory` errors when warnings are present. I'm working on eliminating those warnings, and will remove my patch once they're gone to see if there is indeed a relationship there.

I sure appreciate your prompt response - thanks!

adiroiban:
I see that the code is poorly documented... my bad.

I see that this was done as part of https://github.com/twisted/pydoctor/pull/330

---------

[...]

PR #799, opened 2024-06-20
Sphinx may not generate an api folder

lod:
Trivial fix for a bug with simple projects where sphinx does not generate a docs/build/html/api/ folder.  Without this fix the rename fails because the output_path does not exist.

### DIFF
--- a/pydoctor/sphinx_ext/build_apidocs.py
+++ b/pydoctor/sphinx_ext/build_apidocs.py
@@ -67,7 +67,8 @@ def on_build_finished(app: Sphinx, exception: Exception) -> None:
 
         temp_path = output_path.with_suffix('.pydoctor_temp')
         shutil.rmtree(sphinx_files, ignore_errors=True)
-        output_path.rename(sphinx_files)
+        if output_path.exists():
+          output_path.rename(sphinx_files)
         temp_path.rename(output_path)
 
 


adiroiban:
Hi David,


Thanks for the PR.

Can you please add more details about this use case?

Why is Sphinx not generating the API sub-folder ?

If Sphinx is not generating the API sub-folder, it looks like you are not using the pydoctor extension as designed.

If this is a valid use case, we also need documentation and automated test for this to make sure this feature is supported in future versions.

Regards

fly1d:
I put together a minimal executable reproduction for the missing case discussed here: https://github.com/fly1d/pydoctor-sphinx-api-repro

The pinned GitHub Actions run passes on Ubuntu/Python 3.12 and verifies the target failure: https://github.com/fly1d/pydoctor-sphinx-api-repro/actions/runs/33602073983

Why Sphinx does not generate the `api/` directory in this case:

- `pydoctor_args` writes to `{outdir}/api`.
- `on_builder_inited` successfully builds that output, then moves it to `api.pydoctor_temp`.
- The Sphinx source tree has only `index.rst`, so Sphinx legitimately creates the root `index.html` and no `api/` subtree of its own.
- `on_build_finished` then unconditionally tries to rename the absent `api` path to `api.sphinx_files`.

The oracle checks more than a non-zero exit. Before accepting the failure it requires both the Sphinx `index.html` and pydoctor `api.pydoctor_temp/index.html`, the `on_build_finished` handler, the `api -> api.sphinx_files` rename, and exit code 2. Unrelated configuration/import/build failures are rejected.

I reproduced the same path locally 3/3 with the originally reported `pydoctor 23.4.1 + Sphinx 7.0.1`, and 3/3 with current `pydoctor 25.10.1 + Sphinx 7.0.1`. This looks suitable for the requested automated test: omit `api/index.rst`, run the extension, and assert that the pydoctor output is restored to `api/` without requiring a Sphinx-owned `api/` directory. If that source subtree is intended to be mandatory, an explicit configuration error and documentation would make the contract much clearer than the current late `FileNotFoundError`.

adiroiban:
> I put together a minimal executable reproduction 
> I reproduced the same path locally 3/3 with the originally reported 

@fly1d you reproduced this error or an LLM agent did it... or you are an LLM agent ?

We are happy to review contributions created by humans

fly1d:
> > I put together a minimal executable reproduction
> > I reproduced the same path locally 3/3 with the originally reported
> 
> @fly1d you reproduced this error or an LLM agent did it... or you are an LLM agent ?
> 
> We are happy to review contributions created by humans

@adiroiban , LLM agent did it, including the code and tests. I have read the contrib.rst and I am happy to accept your guidance.

fly1d:
@adiroiban,

To be completely explicit: an LLM agent performed the entire reproduction and testing process. I am not attempting to hide or soften that fact.

The agent used my own tool, [ReproMin](https://github.com/fly1d/repomin), a repository-level failure reducer that shrinks a failing repository while continuously preserving the exact failure signal. I selected the problem, provided the tool, directed and paid for the workflow, supplied the compute and tokens, and published the result. The agent executed every stage of constructing and testing the reproduction.

That makes this an agent-produced reproduction and my contribution at the same time. These facts do not contradict each other. The agent did not independently choose this issue, provide ReproMin, pay for its execution, create a GitHub account, decide to contribute the result, or assume responsibility for publishing it. I did.

The model was GPT-5.6 Sol, which [OpenAI officially describes](https://learn.chatgpt.com/docs/models) as "the most capable GPT-5.6 model for complex coding, computer use, research, and cybersecurity." It is not infallible, but neither are human developers or maintainers. Merely having maintainer status does not establish that someone is technically more capable than GPT-5.6 Sol, much less more capable than the combined workflow of that model, ReproMin, a deterministic failure oracle, repeated execution, and public CI.

Repository authority is not proof of technical superiority. A maintainer has the authority to decide what a repository accepts; that does not make every maintainer's technical judgment automatically superior to the model's work, nor does it make an agent-produced result worthless by definition. Treating such work as garbage without identifying a single technical defect is not technical review. It is prejudice against the method of production dressed up as quality control.

This particular output has demonstrable value. The project had repeatedly requested a minimal reproduction since 2023. A maintainer specifically asked for the failure to be placed in a [public repository or reproduced through GitHub Actions](https://github.com/twisted/pydoctor/issues/728#issuecomment-1799336965). That is exactly what the agent delivered: a nine-file executable fixture, pinned dependencies, public CI, a specific failure oracle, and repeatable results on both the originally reported and current pydoctor versions.

The oracle does not accept an arbitrary non-zero exit. It verifies the successful creation of the relevant Sphinx and pydoctor artifacts, the exact `on_build_finished` path, the missing `api -> api.sphinx_files` rename, and the expected exit status. Unrelated failures are rejected. If this is "low quality" or "garbage," demonstrate that technically. Identify the false positive, the invalid assumption, or the incorrect causal explanation. Until then, the executable evidence has more technical weight than an unsupported opinion based solely on how it was produced.

I also reviewed the policy history. When I posted the reproduction, pydoctor's contribution documentation contained no AI rule and no link to one. The earlier policy draft was unfinished, internally disputed, and eventually closed without being merged. Its own discussion explicitly stated that LLM-assisted investigation and out-of-tree output could be allowed.

The replacement policy was created more than fourteen hours after my reproduction and merged more than seventeen hours afterward. Even while it was being merged, maintainers were still asking whether it applied only to `twisted/twisted` or to every repository in the organization. Afterward, you acknowledged that contributors to other repositories would not know to search the main Twisted documentation and that each repository needed a policy link. The policy author agreed and explicitly wrote: ["Definitely not going to blame anyone for missing this policy in the interim."](https://github.com/twisted/twisted/pull/12835#issuecomment-5518794315)

Nevertheless, after I answered your question truthfully, my contribution was used as an example of someone supposedly trying to hide LLM involvement. I did not mention the agent in my initial comment, but I disclosed it directly when asked. An omission followed by an immediate truthful answer is not evidence of deliberate concealment. More importantly, contributors cannot be accused of violating a rule that had not yet been adopted, scoped, or made discoverable in the repository where their contribution was requested.

I have now read the policy and the surrounding responses carefully. Their tone is condescending and self-righteous. They speak at length about the value of maintainers' time while treating contributors' time, tools, compute, tokens, and results as though they were disposable. They assume that work produced with an agent is presumptively worthless while treating maintainer status itself as sufficient proof of superior judgment. Neither assumption is justified.

Maintainers' time matters. Contributors' time matters too. Open source does not oblige a project to accept every contribution, but when a project repeatedly asks the public to produce a reproducer, maintaining clear rules and examining the requested evidence are part of operating that contribution process. If a project lacks the capacity or willingness to do that work, it should stop soliciting external contributions or clearly close the contribution channel.

What it should not do is ask people to contribute, leave its actual rules scattered across unrelated issues and unmerged pull requests, expect contributors to spend their time reconstructing those rules, and then lecture them about the burden of reading the contribution it requested. That is shifting the cost of inadequate project governance onto contributors.

I respect the policy that has now been formally adopted, and I will treat it as governing future contributions. I do not accept its retroactive use to portray this earlier contribution as dishonest, worthless, or technically inferior.

I contributed a concrete and executable reproduction to help resolve a long-standing problem. ReproMin and GPT-5.6 Sol succeeded in producing the artifact the project had repeatedly requested and had not obtained. If the result is technically wrong, show where. If it is technically correct, then acknowledge its value even if the project has now decided not to accept this method of contribution.

You may reject the workflow as a matter of policy. You do not get to turn that policy preference into proof that the work is garbage, that the model is incompetent, or that the maintainers are technically superior. The evidence does not support any of those conclusions.

glyph:
@fly1d One of the side-effects that I find concerning about LLM use is a sort of hypomanic state that it can put users into, which seems to be a precursor to the much-vaunted "AI psychosis" that has been featured in the news. The *extreme* verbosity, inappropriate emotional intensity, defensiveness *on behalf of models*, and hostility that your comment exhibits suggests to me that you may be suffering these effects.

I'll be banning you from the project because this outburst was completely inappropriate, but I would *seriously* urge you to seek out some mental health support before this escalates to serious real-life consequences for you.

adiroiban:
Now... back to the topic. We have this error and nobody else reproduced this error due to 2 main reasons ... I might be wrong :

* The whole building sphinx + epydoc impementation is one big hack. I have implemented it just to get the docs.twisted.org working
* There are not many people using epydoc + sphinx

--------

Regarding the other part with disclosure. I am only now aware that the patch was generated by `ReproMin`. And the tool itself didn't disclose it.
I am not saying that it tried to hide on purpose, but not disclosing this is a big product design failure.

Codecov or codspeed or other tools will disclose by default that the output from those tools was not human generated and don't even try to look like human generated.

Regarding the superiority of the model.
We know for mare that 10 years that Neural Machine Translation can generate output that is better than a person. I was using Google Translate for a long time and hallucination rate was low.

adiroiban:
@glyph regarding blocking person... I understand that it can take a big emotional toll to interact with LLM agents.

I just wanted to say that in this case , this block feels like censorship.

I know there a big debate regarding freedom of speach and death penalty,  and I don't wanted to go here into details or contest your decision.

I just wanted to say that I am not 100% happy with a person being blocked forever.

Maybe we can have something like block for 1 day... or 1 week

glyph:
It's just moderation. This person is not censored, they are welcome to use whatever other channels are available for their advertisements for OpenAI's models; there has never been a time in history where people had so many opportunities to publicize a message, and never before has there been a message so widely publicized as "use ChatGPT". I actually laughed out loud at the idea that I could "censor" it.

They are just not welcome to use resources under our control to spread such a message.

That said, if you think that a permanent ban is too harsh, that such a person has valuable things to contribute, we could do 90 days as a default for this sort of thing in the future, and consider unbanning them after that window. I am skeptical.

adiroiban:
If GitHub does not provide tools too automatically unblock, then we should not spend more time on that

glyph:
> If GitHub does not provide tools too automatically unblock, then we should not spend more time on that

Agreed it's not worth much more effort. I'll stick with a 90 day ban in the future instead of unlimited (1 week is far too short; people do not recover from this state of mind within a week). For this person I guess I will set a 90-day reminder for myself but not do anything else.

glyph:
(To be clear: Github *does* have an "auto-unban" but only when you select it at the time of the ban and I didn't in this case.)

PR #812, opened 2024-07-31
Enable the security advisories process

adiroiban:
This is mostly paperwork to get the standard GitHub security process up and running.


adiroiban:
this is read for review. thanks

needs-review

adiroiban:
Thanks for the review

With this merged, we have the GitHub security links enabled ... it's not much, but just to make it easier to discover them

https://github.com/twisted/pydoctor?tab=security-ov-file

PR #831, opened 2024-10-25
Wrap signatures onto several lines when function len is over a treshold and function has the focus

### LINKED ISSUE #801: HTML: Break function signature elements into multiple spans
The current HTML output doesn't work well for large complex functions. The function signature is mushed together onto a single long unreadable line.

Sphinx has the maximum_signature_line_length configuration option to do this. However a much simpler option would be to split the signature into multiple spans and allow themes to style long lines via CSS.

As an example Sphinx uses `<em class="sig-param">` for each parameter. The details of the parameter are broken down further but that feels less useful.

As an example for `def extract_final_subscript(annotation: [ast.Subscript](https://docs.python.org/3/library/ast.html#ast.Subscript)) -> ast.expr:`

The current html structure for the arguments is:
`<span class="function-signature">(annotation:<code><a ...>ast.Subscript</a></code>) -> <code>ast.expr</code></span>`

An alternative would be:
`<span class="function-signature">(<span class="function-params"><span class="function-param">annotation:<code><a ...>ast.Subscript</a></code></span></span>) -> <code>ast.expr</code></span>`

Example CSS:
```
.function-params:has(> :nth-child(4)) : /* 5 or more arguments, use vertical styling */
.function-params:not:has(> :nth-child(4)) : /* small function definition, single line styling */
```

tristanlatr:
This change fixes #801.

- Parameters html are divided into .rst-sig-param or .rst-sig-symbol spans.
- When the function is long enough (88 chars excluding the def and ending colon, this is hard coded) an extra CSS class .long-signature is added to the parent function-signature.
- The first parameter 'cls' or 'self' of (class) methods is marked with the 'undocumented' CSS class, this way it's clearly not part of the API.
- Add some CSS  to expand the signature of long functions when they have the focus.
- Refactor CSS rules for parameters tables to be more simpler and predictable with percentage values for smaller screens.
- Remove ValueFormatter classes since they were only a hack to show HTML tags inside `Signature.__str__()`, which we don't need anymore because we're computing the representation ourselve.
- Remove ParsedStanOnly class because we know have better alternatives. 
- Introduce ParsedDocstring.to_text() -> str. 
- When a function has invalid parameters its rendered as (…) instead of confusing () - like it was a zero argument callable.

Examples: 

- Not wrapped because overloaded function never get wrapped. 

Capture d’écran, le 2024-11-15 à 11 48 35

- Wrapped because the function length is over 88 chars

Capture d’écran, le 2024-11-15 à 11 49 30

The way to get the old behaviour back is to use the following custom css:

```css
.rst-sig-param, .rst-sig-symbol {
    display: initial !important;
    margin-left: 0 !important;
    padding-left: 0 !important;
}
```

### DIFF
--- a/pydoctor/epydoc/markup/_napoleon.py
+++ b/pydoctor/epydoc/markup/_napoleon.py
@@ -9,7 +9,6 @@ from pydoctor.epydoc.markup import restructuredtext
 from pydoctor.napoleon.docstring import GoogleDocstring, NumpyDocstring
 
 
-
 class NapoelonDocstringParser:
     """
     Parse google-style or numpy-style docstrings.
@@ -62,6 +61,7 @@ class NapoelonDocstringParser:
         errors: list[ParseError],
         docstring_cls: type[GoogleDocstring],
     ) -> ParsedDocstring:
+        
         docstring_obj = docstring_cls(
             docstring, 
             what=self.objclass,
--- a/pydoctor/epydoc/markup/google.py
+++ b/pydoctor/epydoc/markup/google.py
@@ -10,7 +10,6 @@ from pydoctor.epydoc.markup import ObjClass, ParserFunction
 from pydoctor.epydoc.markup._napoleon import NapoelonDocstringParser
 
 
-
 def get_parser(objclass: ObjClass | None) -> ParserFunction:
     """
     Returns the parser function. Behaviour will depend on the documentable type and system options.
--- a/pydoctor/epydoc/markup/plaintext.py
+++ b/pydoctor/epydoc/markup/plaintext.py
@@ -9,7 +9,6 @@ Parser for plaintext docstrings.  Plaintext docstrings are rendered as
 verbatim output, preserving all whitespace.
 """
 from __future__ import annotations
-
 __docformat__ = 'epytext en'
 
 from typing import List, Optional
@@ -18,7 +17,6 @@ from docutils import nodes
 from twisted.web.template import Tag, tags
 
 from pydoctor.epydoc.markup import DocstringLinker, ObjClass, ParsedDocstring, ParseError, ParserFunction
-
 from pydoctor.epydoc.docutils import set_node_attributes, new_document
 
 def parse_docstring(docstring: str, errors: List[ParseError]) -> ParsedDocstring:
--- a/pydoctor/epydoc/markup/restructuredtext.py
+++ b/pydoctor/epydoc/markup/restructuredtext.py
@@ -60,7 +60,6 @@ from pydoctor.epydoc.markup import Field, ObjClass, ParseError, ParsedDocstring,
 from pydoctor.epydoc.markup.plaintext import ParsedPlaintextDocstring
 from pydoctor.epydoc.docutils import new_document
 
-
 #: A dictionary whose keys are the "consolidated fields" that are
 #: recognized by epydoc; and whose values are the corresponding epydoc
 #: field names that should be used for the individual fields.
--- a/pydoctor/epydoc2stan.py
+++ b/pydoctor/epydoc2stan.py
@@ -1294,7 +1294,7 @@ def _colorize_signature(sig: inspect.Signature,
     return ParsedRstDocstring(set_node_attributes(
         new_document('code'), children=result), ())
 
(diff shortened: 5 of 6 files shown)


glyph:
@tristanlatr sorry for the delay, it'll probably be a few more days (pydoctor reviews are extra hard, in addition to having limited time!) but these are on my to-do list.

docs/google_demo/__init__.py
```diff
@@ -297,4 +297,4 @@ class ExamplePEP526Class:
     """
 
     attr1: str
-    attr2: int
+    attr2: int
```
glyph, line 300: Is this change intentional?

pydoctor/epydoc/markup/__init__.py
```diff
@@ -206,25 +206,115 @@ def to_node(self) -> nodes.document:
...
+    """
+
+    def __init__(self, elements: Sequence[ParsedDocstring]):
+        super().__init__(tuple(chain.from_iterable(e.fields for e in elements)))
+        self._elements = elements
+        self._doc: nodes.document | None = None
+    
+    @property
+    def has_body(self) -> bool: 
+        return any(e.has_body for e in self._elements)
```
glyph, line 260: Huh, why the coverage gap on these properties?

pydoctor/epydoc/markup/_pyval_repr.py
```diff
@@ -198,7 +198,9 @@ def __init__(self, document: nodes.document, is_complete: bool, warnings: List[s
     def to_stan(self, docstring_linker: DocstringLinker) -> Tag:
         return Tag('code')(super().to_stan(docstring_linker))
 
-def colorize_pyval(pyval: Any, linelen:Optional[int], maxlines:int, linebreakok:bool=True, refmap:Optional[Dict[str, str]]=None) -> ColorizedPyvalRepr:
+def colorize_pyval(pyval: Any, linelen:Optional[int], maxlines:int, 
+                   linebreakok:bool=True, refmap:Optional[Dict[str, str]]=None, 
+                   is_annotation: bool = False) -> ColorizedPyvalRepr:
```
glyph, line 203: should we be adopting Black so that this sort of style change doesn't require discretion? :)

pydoctor/epydoc/markup/_types.py
```diff
@@ -48,8 +48,9 @@ def has_body(self) -> bool:
 
     def to_node(self) -> nodes.document:
         """
-        Not implemented.
+        Not implemented at this time :/
         """
+        #TODO: Fix this soon
```
glyph, line 53: Link to an issue?

pydoctor/templatewriter/search.py
```diff
@@ -104,12 +104,7 @@ def format_docstring(self, ob: model.Documentable) -> Optional[str]:
         source = epydoc2stan.ensure_parsed_docstring(ob)
         if source is not None:
             assert ob.parsed_docstring is not None
-            try:
-                doc = ' '.join(node2stan.gettext(ob.parsed_docstring.to_node()))
-            except NotImplementedError:
-                # some ParsedDocstring subclass raises NotImplementedError on calling to_node()
-                # Like ParsedPlaintextDocstring.
-                doc = source.docstring
+            doc = ob.parsed_docstring.to_text()
```
glyph, line 107: lovely to see an ugly too-broad `try` like this go away :)

pydoctor/test/epydoc/test_pyval_repr.py
```diff
@@ -1560,3 +1559,21 @@ def test_expressions_parens(subtests:Any) -> None:
     check_src("{**({} == {})}")
     check_src("{**{'y': 2}, 'x': 1, None: True}")
     check_src("{**{'y': 2}, **{'x': 1}}")
+
+
+def test_is_annotation_flag() -> None:
+    # the is_annotation attribut is added to all links when is_annotation=True is passed.
```
glyph, line 1565:
```suggestion
    # the is_annotation attribute is added to all links when is_annotation=True is passed.
```

glyph (approved):
I do wish I understood this code a *bit* better, but the type checker is doing some good heavy lifting assuaging my concerns :). The new style looks great, and I look forward to deploying this. Thanks!

PR #847, opened 2024-12-06
Handle typing.Generic case in MRO

### LINKED ISSUE #846: Remove duplicated Generic from MRO
This twisted error looks like a bug:
```
/twisted/src/twisted/python/zippath.py:52: Cannot compute linearization of the class inheritance hierarchy
```

We need to remove the `typing.Generic` when it appears more than once. I've seen that in astroid code before.

tristanlatr:
Fix #846.

This should also speed-up the MRO processing since we now process all classes in topological order and use caching for predecessor MROs, avoiding to re-compute for every subclasses like the current code does.

### DIFF
--- a/pydoctor/mro.py
+++ b/pydoctor/mro.py
@@ -108,6 +108,23 @@ class DependencyList:
 def c3_merge(*lists) -> list:
     """
     The merge operation of the C3 superclass linearization algorithm.
+
+    Example of a naive usage:
+
+    >>> def mro(cls: T, getbases: Callable[[T], List[T]]) -> List[T]:
+    ...     result = [cls]
+    ...     if not (bases:=getbases(cls)):
+    ...         return result
+    ...     else:
+    ...         return result + c3_merge(*[mro(kls, getbases) for kls in bases], bases)
+    >>> class B:...
+    >>> class A(B):...
+    >>> class B(B):...
+    >>> class C(A,B):...
+    >>> C.__mro__
+    (<class 'pydoctor.mro.C'>, <class 'pydoctor.mro.A'>, <class 'pydoctor.mro.B'>, <class 'pydoctor.mro.B'>, <class 'object'>)
+    >>> mro(C, lambda cls: cls.__bases__)
+    [<class 'pydoctor.mro.C'>, <class 'pydoctor.mro.A'>, <class 'pydoctor.mro.B'>, <class 'pydoctor.mro.B'>, <class 'object'>]
     
     @see: U{https://en.m.wikipedia.org/wiki/C3_linearization}
     """


pydoctor/model.py
```diff
@@ -622,27 +639,89 @@ def init_finalbaseobjects(o: 'Class', path:Optional[List['Class']]=None) -> None
...
-            return []
-        return list(localbases(o))
+    def _compute_mro(self, cls: Class) -> list[ClassOrStr]:
+        """
+        Compute the method resolution order for this class.
+        This assumes that the MRO of the bases of the class 
+        have already been computed and stored in C{self.computed_mros}.
+        """
+        if cls not in self.graph:
+            raise ValueError
```
glyph, line 699: In general raising exception classes directly isn't great, since it implicitly instantiates; plus, it should probably have a message; could you add some parens here? :)

pydoctor/model.py
```diff
@@ -622,27 +639,89 @@ def init_finalbaseobjects(o: 'Class', path:Optional[List['Class']]=None) -> None
...
+        if not (bases:=self.graph[cls]):
+            return result
+        
+        # since we compute all MRO in topological order, we can safely assume
+        # that self.computed_mros contains all the MROs of the bases of this class.
+        bases_mros = [self.computed_mros.get(kls, []) for kls in bases]
+
+        # handle multiple typing.Generic case, 
+        # see https://github.com/twisted/pydoctor/issues/846.
+        # support documenting typing.py module by using allobject.get.
```
glyph, line 712: Is this going to work for PEP 695 syntax as well or is that a separate issue?

pydoctor/model.py
```diff
@@ -622,27 +639,89 @@ def init_finalbaseobjects(o: 'Class', path:Optional[List['Class']]=None) -> None
...
+        generic = cls.system.allobjects.get(_d:='typing.Generic', _d)
+        if generic in bases and any(generic in _mro for _mro in bases_mros):
+            # this is safe since we checked 'generic in bases'.
+            bases.remove(generic) # type: ignore[arg-type]
+        
+        try:
+            return result + mro.c3_merge(*bases_mros, bases)
+        
+        except ValueError as e:
```
glyph, line 721: When doing an `except` on a standard library exception, it's important to constrain the code in the `try` as much as possible. Are you expecting it out of the `c3_merge`, or the `+`? One or the other should move up outside the `try`.

pydoctor/mro.py
```diff
@@ -103,7 +103,7 @@ def remove(self, item: Optional[T]) -> None:
                 i.popleft()
 
 
-def _merge(*lists) -> list:
+def c3_merge(*lists) -> list:
```
glyph, line 106: Mind dropping a quick docstring here, while you are in the neighborhood :)

glyph (approved):
LGTM!

pydoctor/model.py
glyph: Sure. But even an assert should have a non-empty message attached :)

glyph:
> Diff from [pydoctor_primer](https://github.com/tristanlatr/pydoctor_primer), showing the effect of this PR on open source code:

It's so cool to see this providing such a legible example of the impact of the fix :)

PR #874, opened 2025-02-02
Refactor the the ParsedTypeDocstring

### LINKED ISSUE #581: Fix little issue in the types rendering when --process-types is enabled
by default 2022-05-15 at 4 17 21 PM

I guess the expected behaviour would be to have only one `<code>` tag per types when there is no LITERAL, UNKNOWN or ANY tokens in the type (like the type above). 

The issue resides in the following code:

https://github.com/twisted/pydoctor/blob/814db7a1351bf206a3ace95a18126e96dec030ce/pydoctor/epydoc/markup/_types.py#L116

We can probably do something to include all types of delimiters.
But some logic code must be added because some delimiters should not be present at the end of the `<code>` block, like `,` or ` or ` or ` and `, etc. If some token happens to have an inconsistent type, they should be treated as UNKNOWN. 

Let's note that the GoogleDocstring class has it's own warning system, which it's used to determine if a type if valid or not, if we decide to trigger some more warnings by default from GoogleDocstring, some types might not be recognized as valid anymore. So  introducing new warnings can be difficult. 

There is quite already a lot of complexity in the `--process-type` option, but if we can fix it without adding too much code that would be good.

### LINKED ISSUE #873: Implement ParsedTypeDocstring.to_node()
This `ParsedDocstring` subclass is the only one that is missing an implementation for that method. 
It it preventing the PR #723 that fixes a long standing bug to be complete. 

This is somewhat linked to #581 in the sense that both issue can be fixed together.

tristanlatr:
Refactor the the `ParsedTypeDocstring` - used in numpy and google docformat as well as when --process-types option is passed. 

This gets rid of some unfortunate code that was not doing it's thing right. We replace that dirty processing by legitimately creating a docutils tree for these types are rendering correctly. In order to do that, the linker had to be adjusted not to wrap links inside `<code>` tags anymore (see the preliminary refactor commit ec6c907c59d99f7766901da8238f8de248351950).

Make possible for PR #723 to be complete
Fixes #581
Fixes #873 

Before: 
Capture d’écran, le 2025-02-02 à 15 31 32

After:
Capture d’écran, le 2025-02-02 à 15 31 57

The warning diff was unexpected... it appears that pydoctor was reporting link not found at several unrelated places, increasing the number of warnings uselessly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,6 +73,9 @@ What's New?
 in development
 ^^^^^^^^^^^^^^
 
+* Fix bug that would result in duplicated "Cannot find link target" warnings when the 
+  types under a docstring *Attributes* section failed to resolved.
+
 pydoctor 25.4.0
 ^^^^^^^^^^^^^^^
 
@@ -92,8 +95,6 @@ pydoctor 25.4.0
 * Fix a bug in the MRO computing code that would result in an incorrect  
   ``Cannot compute linearization of the class inheritance hierarchy`` message 
   for valid types extending ``typing.Generic`` as well as other generic classes.
-* Fix bug that would result in duplicated "Cannot find link target" warnings when the 
-  types under an docstring *Attributes* section failed to resolved.
 
 pydoctor 24.11.2
 ^^^^^^^^^^^^^^^^


pydoctor/epydoc/markup/_types.py
```diff
@@ -5,15 +5,17 @@
...
+from pydoctor.epydoc.markup import ParseError, ParsedDocstring
+from pydoctor.epydoc.markup.restructuredtext import parse_docstring
+from pydoctor.epydoc.markup._pyval_repr import PyvalColorizer
 from pydoctor.napoleon.docstring import TokenType, TypeDocstring
+from pydoctor.epydoc.docutils import new_document, set_node_attributes
 
 from docutils import nodes
-from twisted.web.template import Tag, tags
 
+# TODO: This class should use composition instead of multiple inheritence...
```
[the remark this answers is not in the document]
glyph: When you mark for a follow-up like this, can you also add a link in the code, to ensure that when it's addressed, the reader can tell if the comment is stale or not?

pydoctor/epydoc/markup/_types.py
```diff
@@ -5,64 +5,53 @@
...
     FIELDS = ('type', 'rtype', 'ytype', 'returntype', 'yieldtype')
     
-    #                                                   yes this overrides the superclass type!
-    _tokens: list[tuple[str | nodes.Node, TokenType]] # type: ignore
-
-    def __init__(self, annotation: Union[nodes.document, str],
-                 warns_on_unknown_tokens: bool = False, lineno: int = 0) -> None:
+    def __init__(self, annotation: nodes.document, 
+                 warns_on_unknown_tokens: bool = False, 
+                 lineno: int = 0) -> None:
```
glyph, line 29: The indentation flourish here reminds me that we need to apply `black` to this project :)

pydoctor/epydoc/markup/_types.py
```diff
@@ -5,64 +5,53 @@
...
 
-            _tokens = self._tokenize_node_type_spec(annotation)
-            self._tokens = cast('list[tuple[str | nodes.Node, TokenType]]', 
-                                self._build_tokens(_tokens))
-            self._trigger_warnings()
-        else:
-            TypeDocstring.__init__(self, annotation, warns_on_unknown_tokens)
-        
+        _tokens = self._tokenize_node_type_spec(annotation)
+        # yes this overrides the superclass type!
```
glyph, line 34: mypy is telling us that we override the superclass type but it would be more helpful to say why :)

pydoctor/epydoc/markup/_types.py
```diff
@@ -5,64 +5,53 @@
...
+        self._tokens: list[tuple[str | nodes.Node, TokenType]] \
+            = self._build_tokens(_tokens) # type: ignore
```
glyph, line 36: 1. `type:ignore`s should always be as narrow as possible, to avoid over-broad ignores; so, this should be `# type:ignore[assignment]`
2. there are actually 2 things going on in this statement; changing the type of `self._tokens`, and assigning it to the value of `_build_tokens`, which means that the `type:ignore` is going to catch *two* errors, even if you narrow it to `[assignment]`, and thus kinda obviating the point of having a type declaration at all. I would sugggest reverting to what you had before in terms of declaring the field at class scope, leaving the `type:ignore` there, and then removing the `cast` here, which *should* no longer be necessary.

pydoctor/epydoc/markup/_types.py
```diff
@@ -85,91 +74,52 @@ def _warn_not_supported(n:nodes.Node) -> None:
...
-            else:
-                # the token will be processed in _convert_type_spec_to_stan() method.
-                combined_tokens.append((_token, _type))
-
-        return combined_tokens
+    _converters: Dict[TokenType, Callable[[str, int], nodes.Node]] = {
+                                        # we're re-using the variable string css 
+                                        # class for the whole literal token, it's the
+                                        # best approximation we have for now. 
+            TokenType.LITERAL: lambda _token, _: \
```
glyph, line 81: the backslashes here are unnecessary; you're in an enclosed expression context because of the outer `{}`, so you can just remove them.

pydoctor/epydoc/markup/_types.py
```diff
@@ -84,97 +75,42 @@ def _warn_not_supported(n:nodes.Node) -> None:
...
-                    combined_tokens.append((_token, _type))
-                
-                if _token == "]": open_square_braces -= 1
-                elif _token == ")": open_parenthesis -= 1
-
-            else:
-                # the token will be processed in _convert_type_spec_to_stan() method.
-                combined_tokens.append((_token, _type))
-
-        return combined_tokens
```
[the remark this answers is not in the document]
glyph: it does look unfortunate, nice to see it go

pydoctor/node2stan.py
```diff
@@ -144,6 +154,8 @@ def depart_code(self, node: code) -> None:
     def should_be_compact_paragraph(self, node: nodes.Element) -> bool:
         if self.document.children == [node]:
             return True
+        elif self._document_is_code:
+            return True
```
glyph, line 158: huh, why no coverage here?

pydoctor/test/test_napoleon_docstring.py
```diff
@@ -277,6 +311,40 @@ def test_convert_numpy_type_spec(self):
         for spec, expected in zip(specs, converted):
             actual = str(TypeDocstring(spec))
             self.assertEqual(expected, actual)
+    
+    def test_natural_language_delimiters_parsed_tokens(self):
```
glyph, line 315: just a suggestion, whenever possible, allow new tests to be type-checked, to help provide some external validity to the type signatures?
```suggestion
    def test_natural_language_delimiters_parsed_tokens(self) -> None:
```

glyph (approved):
Generally looks good. This is wandering pretty far afield of my usual usage of pydoctor, but just some minor code quality issues inline.

PR #880, opened 2025-03-18
Remove lunr dependency upper bound

carlwgeorge:
Lunr has released version 0.8.0, and the pydoctor tests all pass with this version.


carlwgeorge:
Sure, rebased with a new commit message.

PR #889, opened 2025-04-07
Release pydoctor version 25.4.0

tristanlatr:


glyph:
> I’ll need to fix or disable the pydoctor_primer run of we want to have a green ci before publishing the next version.

Oh, hrm. What happened there?

PR #904, opened 2025-07-09
Process ivar fields and friends when introspecting code (wrt C-modules)

### LINKED ISSUE #903: Epytext fields are not processed in class docstrings if the class is defined in a C extension
It seems like Epytext fields are not processed by PyDoctor if the fields are defined in the docstring of a class that is defined by a C extension. I am not familiar with PyDoctor's internals, but it seems like `epydoc2stan.extract_fields(...)` is only called from the part of the code that analyzes the AST of the module being documented. C extensions have no AST so this approach will not work there.

To demonstrate that the bug exists, I have added two `@ivar` fields to the [`Vertex`][2] class of the Python interface of [igraph][1].

[1]: https://github.com/igraph/python-igraph/
[2]: https://github.com/igraph/python-igraph/blob/main/src/_igraph/vertexobject.c#L879

Checking out the [repository][1], running `git submodule update --init --recursive` and then running `scripts/mkdoc.sh` from the root should generate the API documentation from scratch. (The script takes a while to run at first invocation because it creates a virtualenv with all the required dependencies). You will see that the generated API documentation of the `Vertex` class does not include the instance variables declared by `@ivar`. I added some debug statements in `_introspectThing()` and `epydoc2stan.extract_fields()` and it seems like `epydoc2stan.extract_fields()` is never called on any class in the `igraph._igraph` module (this is the C glue code). While `_introspectThing()` processes the `igraph._igraph.Vertex` class, its instance fields are not handled by any of the branches in `_introspectThing()` either.

Refs: https://github.com/igraph/python-igraph/issues/813

tristanlatr:
Fixes #903

@ntamas, this fixes the issue you reported lately.

The testing does not explicitely cover C-module, but I'm pretty confident this will do the trick. 

If I understand correctly the initial problem, the ivar fields you were trying to use is just a workaround another issue in pydoctor, that is that property defined in C code are not recognized because they don't appear in the `__dict__` attribute of classes. Maybe by using `inspect.getmembers` instead we could fix this issue as well? Tell me what you think, thanks.


ntamas:
I tested the igraph documentation with this PR and I can confirm that `@ivar` tags are now recognized properly in the docstring of a class defined in a C module so this can be merged.

At the same time, yes, you are correct, adding an `@ivar` in the `Vertex` and `Edge` classes of igraph is a workaround for the problem that getters/setters defined in a class of a C module are not processed by PyDoctor. They _do_ appear in `__dict__`, but their value is of type `getset_descriptor`, which is not handled by any of the branches in `_introspectThing`:

```python
>>> from igraph import Edge
>>> type(Edge.__dict__["index"])
<class 'getset_descriptor'>
```

glyph:
I believe that this descriptor type is exposed publicly as `types.GetSetDescriptorType` .

PR #932, opened 2025-11-07
Hide the sidebar title when all items under it are private

### LINKED ISSUE #600: Hide the sidebar title when all items are private (and show it when show private is True)

tristanlatr:
Fixes #600

Example of problematic sidebar: https://pydoctor.readthedocs.io/en/latest/api/pydoctor.epydoc.markup.epytext.ColorizingError.html

Example of fixed sidebar: https://pydoctor--932.org.readthedocs.build/en/932/api/pydoctor.epydoc.markup.epytext.ColorizingError.html


README.rst
```diff
@@ -73,6 +73,8 @@ What's New?
 in development
 ^^^^^^^^^^^^^^
 
+* Hide the sidebar title when all items under it are private.
```
tristanlatr:
```suggestion
* Hide sidebar element title when all items under it are private.
```

PR #956, opened 2026-09-23
Fix deprecated admonition rendering when version is omitted

### LINKED ISSUE #950: Bug in rendering of deprecated admonition
This docstring:

```
 .. deprecated::
     Legacy API, see the docs above for new usage
```

Renders like: 

Image

Which obviously makes no sense

muhammad-a-dev:
Fixes #950.

## Summary
When `.. deprecated::` is used without a version argument, docutils pulls the indented explanation into the directive arguments. The first word was then treated as a version, producing nonsensical output like "Deprecated since version Legacy: API, ...".

This change:
- Makes the version optional for the `deprecated` directive (`Deprecated` subclass).
- Detects real version strings (numeric / `NEXT`) so explanations are not misread as versions.
- Renders `Deprecated:` when no version is given (instead of `Deprecated since version ...`).
- Adds regression tests for the no-version cases.

## Changelog (for README `in development`)
* Fix rendering of the ``deprecated`` reStructuredText directive when the version argument is omitted. Previously the first word of the explanation was incorrectly treated as a version number (issue #950).

## Notes
- AI assisted with implementation and tests.
- Unit tests in `pydoctor/test/epydoc/test_restructuredtext.py` pass locally (21 passed).


pydoctor/epydoc/markup/restructuredtext.py
```diff
@@ -429,12 +449,29 @@ def run(self) -> List[nodes.Node]:
...
+            # Version omitted (or explanation mis-parsed as the version).
+            node['version'] = ''
+            text = versionlabels_no_version[self.name]
+            if self.arguments:
+                explanation = ' '.join(self.arguments)
+        else:
+            node['version'] = self.arguments[0]
+            text = versionlabels[self.name] % self.arguments[0]
+            if len(self.arguments) == 2:
+                explanation = self.arguments[1]
```
tristanlatr: Looks like this branch is not reached by the tests

muhammad-a-dev:
thanks for merging master into this

codecov is flagging about 5 lines in `restructuredtext.py` as uncovered on the no-version path. should i add more tests for those branches, or is this close enough as-is?

happy to follow up either way

